PR #1, opened 2015-10-22
Recreate prison visits on bare Rails app

threedaymonk:
Still to do:

- [ ] Add logstasher
- [ ] Mask sensitive parameters
- [ ] Import JS for prison chooser
- [ ] Links back to previous steps on confirmation page
- [ ] Add emails (use premailer to avoid duplicating HTML and text)
- [ ] Add prison processing interface
- [ ] Add additional visitors


app/controllers/steps_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+  def create
+    select_step
+  end
+
+private
+
+  def available_slots
+    @prisoner_step.prison.available_slots
+  end
+  helper_method :available_slots
```
threedaymonk: Reaching all the way into `@prisoner_step.prison.available_slots` from the view also feels wrong, but I'm not really happy about this. It feels like the right place would be `@slots_step.available_slots`, but that turns out to be quite a bit more intrusive to set up.

app/views/steps/_hidden_prisoner_step.html.erb
```diff
@@ -0,0 +1,10 @@
+<%= hidden_field_tag('prisoner_step[first_name]',
+                     @prisoner_step.first_name) %>
+<%= hidden_field_tag('prisoner_step[last_name]',
+                     @prisoner_step.last_name) %>
+<%= hidden_field_tag('prisoner_step[date_of_birth]',
+                     @prisoner_step.date_of_birth) %>
+<%= hidden_field_tag('prisoner_step[number]',
+                     @prisoner_step.number) %>
+<%= hidden_field_tag('prisoner_step[prison_id]',
+                     @prisoner_step.prison_id) %>
```
threedaymonk: I'm torn between a desire to DRY this up and a general distaste of gratuitous metaprogramming

app/controllers/steps_controller.rb
```diff
@@ -0,0 +1,71 @@
...
+  STEPS = %i[
+    prisoner_step
+    visitors_step
+    slots_step
+    confirmation_step
+  ]
+
+  def select_step
+    STEPS.find { |s| send(s) } || complete
+  end
```
threedaymonk: I'm not really happy with the way this mixes up the functional and the side-effectful, but all my attempted refactors have been even worse.

PR #5, opened 2015-10-29
Port sendgrid api from the old app

tatyree:
This is one of the main components for sending mail from the old app.
I have broken down the very long spec into several smaller specs
organised by the instance methods.


SteveMarshall:
This is really good, and the splitting up of the specs is great :+1: 

How much of a pain would it be to break this up into a few commits, though? There are a couple of clear blocks of things I can see:

- Remove unnecessary documentation comments
- Use WebMock to block connections to the web during tests
- Port the Sendgrid code from the old app. (Although I might consider making that a note in the commit message, rather than the title. The important thing is that we're adding Sendgrid support, its source is secondary, or will be, to someone working on this in ~18 months…)

I do still kinda wish we could maintain some of the longer-term provenance of the code from the old app, but I suppose that's kinda the point… (And cherry-picking the commits will be an absolute nightmare.)

tatyree:
@SteveMarshall I thought about it and decided to err on the side of brevity.  

As soon as I finish with this machine rebuild, I'll pull it down, split it apart and see how neatly I can break it out.

app/services/sendgrid_api.rb
```diff
@@ -0,0 +1,58 @@
...
+  def user_name
+    smtp_settings[:user_name]
+  end
+
+  def password
+    smtp_settings[:password]
+  end
+
+  def smtp_settings
+    Rails.configuration.action_mailer.smtp_settings || {}
```
dan-palmer: I don't think we need the fallback to empty hash - sure it can be set in the environment config file.

dan-palmer: In fact it is already set in production/test/dev - so this isn't needed.

SteveMarshall:
This is looking muuuuch better, @tatyree. One suggestion though: `Extract` rather than `Deprecate` might be clearer?

SteveMarshall:
This looks a really solid refactor, @tatyree :clap: 

A couple of things I was wondering, though:

- How does this behave when running the app locally? Will we be able to step through the entire flow without having Sendgrid credentials configured?
- Is there a way to check Sendgrid is up, for when we add a health check? (Probably we'll want to add that later, when we actually put the health check together, rather than building it now, of course…)

app/services/sendgrid_api.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  def api(&_action)
+    yield
+  rescue => e
+    Rails.logger.error("SendgridApi error: #{e.class} #{e}")
+    false
+  end
+
+  def spam_reports
+    SendgridToolkit::SpamReports.new
```
dan-palmer: It seems a bit magical having the toolkit perform the environment variable lookup for the username & password - I was only aware of this feature after looking at the project.

I'd sooner it be set in the project - we could have a sendgrid initializer that grabs the ENV vars and sets this up for us.

dan-palmer: I do think the removal of checking the credentials is present for each api call is definitely warranted though.

spec/services/sendgrid_api/bounced_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+                "email": "test@example.com"
+              }
+            ])
+      }
+
+      it_should_behave_like 'there is something to report'
+    end
+  end
+
+  context 'sendgrid credentials are not set' do
```
dan-palmer: Couldn't this context be set in the included example? Have seen it a few times.

app/services/sendgrid_api.rb
tatyree: @dan-palmer relying on the toolkit's automagical behaviour was Ash's preferred approach the two times I worked on something like this when on IRAT.  I'm neutral about it, personally, but thought I'd err on the side of something I knew IRAT liked.

tatyree:
@SteveMarshall it will step through the workflow as well as the previous version did.  That one skipped the yield if the credentials weren't present.  This one rescues and logs all errors and returns false (an error is raised if api_user or api_key aren't set).

spec/services/sendgrid_api/bounced_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    include_examples 'error handling'
+
+    context 'when there is no bounce' do
+      it_should_behave_like 'there is nothing to report'
+    end
+
+    context 'when there is a bounce' do
+      let(:body) {
+        %([
+              {
```
threedaymonk: four spaces?!

spec/services/sendgrid_api/spam_reported_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    context 'when there is a spam report' do
+      let(:body) {
+        %([
+              {
+                "ip": "174.36.80.219",
+                "email": "test@example.com",
+                "created": "2009-12-06 15:45:08"
+              }
+            ])
```
threedaymonk: Could we format this with two spaces throughout?

SteveMarshall:
@tatyree :+1: So long as we can carry on through the flow locally, I'm happy.

PR #6, opened 2015-11-02
Prevent puma from breaking on El Capitan

tatyree:
Version 2.14.0 makes openssl optional, which prevents the gem build from
choking.


tatyree:
Description of issue is here:

https://github.com/puma/puma/issues/783

PR #7, opened 2015-11-02
Add logstasher

EGiataganas:


config/environments/production.rb
```diff
@@ -81,4 +81,21 @@
 
   # Do not dump schema after migrations.
   config.active_record.dump_schema_after_migration = false
+
+  # Enable the logstasher logs for the current environment
```
SteveMarshall: Can we remove these documentation comments? If someone needs to know what a line does, they can look it up…

EGiataganas: @SteveMarshall done :+1:

PR #9, opened 2015-11-03
Add shoulda-matchers Gem

EGiataganas:


SteveMarshall:
It might be nice, with some of these, to know *why* the gems are being added…

tatyree:
@SteveMarshall mea culpa--I got in a rush there.  I suggested @EGiataganas add this gem to simplify some of the basic specs needed for things like the additional_visitor model.  I also missed that it needed configuration, so we'll add the explanation there.

PR #11, opened 2015-11-03
Add additional visitor spec

EGiataganas:


spec/models/additional_visitor_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'rails_helper'
+
+RSpec.describe AdditionalVisitor do
+  it { should belong_to(:visit) }
```
dan-palmer: We should use RSpec's `expect` syntax here:

```ruby
it { is_expecetd.to belong_to :visit }
```

dan-palmer: Same applies to the other tests. This is worth a read: http://rspec.info/blog/2012/06/rspecs-new-expectation-syntax/

SteveMarshall: @dan-palmer `is_expecetd`, eh? Is that a new feature? :trollface:

dan-palmer: 😭

SteveMarshall:
@threedaymonk LGTM! Rebase?

PR #12, opened 2015-11-09
Email prisons when a booking request is received

threedaymonk:
To do:

- [x] Use `deliver_later` instead of `deliver_now`.
- [x] Extract email text to locale file.
- [x] Move header logo to attachment

We also need to wire up the links in the email, but we don't have the endpoint for these yet.


SteveMarshall:
While I agree with ditching the duplicated effort of maintaining the two copies of the emails, are there any features of the text emails that we'll lose with premailer?

In particular, I'm looking at how premailer converts links to be of the form `[Inner text] ([href])`. If you compare that to [how we explain URLs now](https://github.com/ministryofjustice/prison-visits/blob/master/app/views/prison_mailer/booking_request_email.text.erb#L32...L34), that strikes me as… less good. I wonder if there's some middle-ground?

threedaymonk:
I wonder if anyone at all actually sees the plain text emails! Apart from edge cases like Mutt (and those people need no help), are there any clients from the past decade that don't show HTML by default?

threedaymonk:
For the record, the text part comes out like this:

```
GOV.UK

Prisoner:
Oscar Wilde

Number:
a1234bc

Date of birth:
31 December 1980

Choice 1

Tuesday 17 November 14:00 - 16:10

Visitor 1

Ada Lovelace

Age:
44

Date of birth:
30 November 1970

Email:
ada@test.example.com

Phone:
01154960222

Process the booking ( /TODO )

If you cannot click on the link above, copy the entire link and
paste it into the address bar of your browser.

/TODO

Visit ID: 1
```

threedaymonk:
We could also force it to put things on the same line (it just depends on how we arrange the elements in the HTML:

```Run options: include {:focus=>true}
GOV.UK

Prisoner: Oscar Wilde
Number: a1234bc
Date of birth: 31 December 1980

Choice 1

Tuesday 17 November 14:00 - 16:10

Visitor 1

Ada Lovelace
Age: 44
Date of birth: 30 November 1970
Email: ada@test.example.com
Phone: 01154960222

Process the booking ( /TODO )

If you cannot click on the link above, copy the entire link and
paste it into the address bar of your browser.

/TODO

Visit ID: 1
```

SteveMarshall:
Great! As discussed, I was mis-remembering the order of events. The HTML part was added later, and *was* led by user-needs, and the text part was the original implementation. Given the structure of that text version, I'm pretty happy to go with it as-is.

@leighmoney: Do you mind casting an eye over those plain-text versions of the PVB emails? Which do you prefer? Is there anything you'd like to see fixed?

spec/services/booking_request_creator_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+  let(:visitors_step) {
+    VisitorsStep.new(
+      first_name: 'Ada',
+      last_name: 'Lovelace',
+      date_of_birth: Date.new(1970, 11, 30),
+      email_address: 'ada@test.example.com',
+      phone_no: '01154960222'
+    )
+  }
+
```
tatyree: Should we not abstract these into factories?

spec/services/booking_request_creator_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+  let(:slots_step) {
+    SlotsStep.new(
+      option_1: '2015-01-02T09:00/10:00',
+      option_2: '2015-01-03T09:00/10:00',
+      option_3: '2015-01-04T09:00/10:00'
+    )
+  }
+
```
tatyree: Same with SlotsStep re. factories.

tatyree: Also, SlotsStep sounds like something you do when the Zumba class is oversubscribed :smiling_imp:

spec/services/booking_request_creator_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+        visitor_first_name: 'Ada',
+        visitor_last_name: 'Lovelace',
+        visitor_date_of_birth: Date.new(1970, 11, 30),
+        visitor_email_address: 'ada@test.example.com',
+        visitor_phone_no: '01154960222',
+        slot_option_1: '2015-01-02T09:00/10:00',
+        slot_option_2: '2015-01-03T09:00/10:00',
+        slot_option_3: '2015-01-04T09:00/10:00'
+      )
+
```
tatyree: I preferred the way you did this in steps_controller_spec.

spec/services/booking_request_creator_spec.rb
threedaymonk: My feeling is that it wouldn't be useful, because it's the values that are important and that are being checked below. If we had factories, we could change `VisitorsStep.new(` to `build(:visitors_step, `, but they'd otherwise be much the same.

spec/services/booking_request_creator_spec.rb
threedaymonk: In what way?

PR #13, opened 2015-11-10
Add reference number to visit

tatyree:
We were going to use Base32 for this, but then decided to use v4 UUIDs
    instead.  A Base32 long enough to prevent collisions would only be
    slightly shorter than a UUID, would require further explanation and
    additional gems.

### DIFF
--- a/spec/factories/visits.rb
+++ b/spec/factories/visits.rb
@@ -10,7 +10,8 @@ FactoryGirl.define do
     end
 
     prisoner_date_of_birth '1970-01-01'
-    prisoner_number do |n|
+
+    sequence(:prisoner_number) do |n|
       'ABC%04d' % n
     end
 
@@ -35,6 +36,9 @@ FactoryGirl.define do
     slot_option_1 do |v|
       v.prison.available_slots.first.iso8601
     end
+
     processing_state 'start'
+
+    reference_number { SecureRandom.uuid }
   end
 end
--- a/app/models/visit.rb
+++ b/app/models/visit.rb
@@ -11,6 +11,8 @@ class Visit < ActiveRecord::Base
 
   delegate :email_address, to: :prison, prefix: true
 
+  before_create :set_reference_number
+
   def prisoner_full_name
     [prisoner_first_name, prisoner_last_name].join(' ')
   end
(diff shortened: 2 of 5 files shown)


SteveMarshall:
The reference here is really rather long. Are we going to expect visitors to remember it? I seem to recall (but can't find?) another service used ~4 short words as the reference; maybe that would be nicer (if the code exists somewhere that we can borrow from).

If we're not going to make the reference use words, we should at least ensure we avoid commonly-confused letters/numbers.

threedaymonk:
The motivation for this reference isn't to be memorable as such, but to be impossible to guess. We could equally well use a v4 (random) UUID; I suggested base 32 as it is shorter than hexadecimal.

[This base 32 encoding](http://crockford.com/wrmg/base32.html) *does* avoid commonly-confused characters, though:

> We chose a symbol set of 10 digits and 22 letters. We exclude 4 of the 26 letters: I L O U.
>
> When decoding, upper and lower case letters are accepted, and i and l will be treated as 1 and o will be treated as 0. When encoding, only upper case letters are used.

spec/factories/visits.rb
```diff
@@ -0,0 +1,38 @@
...
+
+    visitor_email_address do
+      FFaker::Internet.disposable_email
+    end
+
+    visitor_phone_no do
+      FFaker::PhoneNumber.short_phone_number
+    end
+
+    slot_option_1 '2015-01-02T09:00/11:00'
```
dan-palmer: ```slot_option_1 { |v| v.prison.available_slots.first.iso8601 }```

This would make it a genuine date as used by the prison.

threedaymonk: Great idea. I'm going to put that into #14.

threedaymonk:
We can remove the Visit factory commit now, as that's added by #14.

SteveMarshall:
@threedaymonk Ah! I didn't realise Crockford's base32 did that. In that case, I'm fine with this encoding, then, but I wonder if it might be worth reducing the number of digits from 27 to something maybe a little more human-friendly?

threedaymonk:
I wonder if `reference_number` is really a good name for this, and whether calling it `uuid` might be better.

I think it would be good to redefine `to_param` to use the UUID - and maybe the `find` method too.

PR #15, opened 2015-11-11
Use uuid for primary_key on all records

tatyree:
We were discussing the best way to provide a reference number for
visitors to allow them to follow the progress of their booking.  We
experimented with Base32, but discarded it in favour of a v4 UUID. In
order to provide security and avoid collisions, a Base32 reference would
have to be almost the same length as a UUID.

In the next step, I attempted to add a UUID column.  Whilst in the
process, I realised it would be more efficient to simply convert the
schema so that all models used uuid as their primary key.  This meant
that I didn't have to adapt any branching `find` methods, nor declare
`to_param` on any models wanting to expose their UUIDs.

I am leaving the Base32 stuff in the history for reference purposes.
This commit supercedes all of that, however.

NOTE: You will need to `bundle exec rake db:drop db:setup` to make this
work.


db/seeds/prisons/AYI-aylesbury.yml
```diff
@@ -1,35 +1,39 @@
...
+  - '2015-01-01'
+  - '2015-01-19'
+  - '2015-01-21'
+  - '2015-01-22'
+  - '2015-01-24'
+  - '2015-01-25'
+  - '2015-01-26'
+  - '2015-01-28'
+  - '2015-01-29'
+updated_at: 2015-11-11 16:20:53.960877000 Z
```
threedaymonk: We don't want `updated_at` or `created_at`.

db/seeds/prisons/AYI-aylesbury.yml
```diff
@@ -1,35 +1,39 @@
...
-- 2015-01-21
-- 2015-01-22
-- 2015-01-24
-- 2015-01-25
-- 2015-01-26
-- 2015-01-28
-- 2015-01-29
+id: 8cc929da-5954-44e2-8e3b-434215660965
+phone_no: 01296 444099
+slot_details:
```
threedaymonk: The importer needs to change to update this changed format.

threedaymonk:
After discussion, we've decided that it would be beneficial to keep the seed format the same as the files in the old application (so that updates can trivially be ported across). We'll maintain the file name to UUID mapping in a separate file.

db/migrate/20151021105400_enable_uuid_extension.rb
```diff
@@ -0,0 +1,5 @@
+class EnableUuidExtension < ActiveRecord::Migration
+  def change
+    enable_extension 'uuid-ossp'
```
dan-palmer: :+1: using the PG extension

tatyree: NONE SHALL PASS...

db/seeds/prisons/uuid_mappings.yml
```diff
@@ -0,0 +1,132 @@
+---
+"/Users/todd/projects/prison-visits-2/db/seeds/prisons/ACI-altcourse.yml": f48d069a-64e2-493a-b256-fa5a35517418
```
threedaymonk: Um. We're not all called Todd, you know!

SteveMarshall: @threedaymonk Are you sure about that, Todd?

db/seeds.rb
```diff
@@ -1,14 +1,22 @@
 Prison.transaction do
+  mapping_path =
+    Rails.root.join('db', 'seeds', 'prisons', 'uuid_mappings.yml')
+
+  mappings = YAML.load(File.read(mapping_path)) || {}
```
threedaymonk: I think I'd prefer a less generic name than "mappings".

spec/models/visit_spec.rb
```diff
@@ -1,6 +1,12 @@
 require 'rails_helper'
 
-RSpec.describe Visit do
+RSpec.describe Visit, type: :model do
+  subject { build(:visit) }
+
+  specify do
+    expect(subject.valid?).to be_truthy
```
threedaymonk: Is there any point to this, given that we already have `FactoryGirl.lint`? (It might be worth adding `factory_girl:lint` to the default Rake task, though.)

lib/tasks/maintenance.rake
```diff
@@ -0,0 +1,19 @@
+namespace :maintenance do
+  task prison_uuids: :environment do
+    mapping_path =
+      Rails.root.join('db', 'seeds', 'prisons', 'uuid_mappings.yml')
+
+    mappings = YAML.load(File.read(mapping_path)) || {}
+
+    Dir[Rails.root.join('db', 'seeds', 'prisons', '*.yml')].each do |path|
```
threedaymonk: This glob includes `uuid_mappings.yml`. Does that work? Should it?

db/migrate/20151021105451_create_prisons.rb
```diff
@@ -1,6 +1,6 @@
 class CreatePrisons < ActiveRecord::Migration
   def change
-    create_table :prisons do |t|
+    create_table :prisons, id: :uuid do |t|
```
threedaymonk: After the hassle that has been caused by changing migrations on the old app, I'm thinking that we should adopt an absolute rule of never changing a migration. Since there is no real data yet, this could be achieved by dropping the old tables and creating them afresh. I realise that it probably seems like unnecessary hassle in this case, but I think it's good hygiene.

db/seeds/prisons/uuid_mappings.yml
tatyree: Oopsie...

db/seeds.rb
tatyree: mapping_stuff? :trollface:

spec/models/visit_spec.rb
tatyree: Old habit from teams that didn't consistently lint.  I'll take it out.

lib/tasks/maintenance.rake
tatyree: It skips `uuid_mappings.yml`.  On reflection, this seems like a bit of a smell.  I'll move it up a level and rename it.

threedaymonk: :+1:

db/migrate/20151021105451_create_prisons.rb
tatyree: Sing with me now...sche-ma-less...Forward comrades!  To the brave new world of the schemaless key value store with relations!

db/seeds.rb
```diff
@@ -8,7 +16,7 @@
       email_address: hash.fetch('email', nil),
       enabled: hash.fetch('enabled', true),
       estate: hash.fetch('estate'),
-      id: hash.fetch('id'),
+      id: prison_uuid_mappings[path],
```
threedaymonk: if the mapping isn't found, does it attempt to create a null `id` and fail, or does the DB-generated UUID take precedence?

lib/tasks/maintenance.rake
```diff
@@ -0,0 +1,23 @@
...
+      if prison_uuid_mappings[filename]
+        puts "Mapping exists for #{filename}. Skipping."
+      else
+        puts "No mapping for #{filename}. Generating."
+        prison_uuid_mappings[filename] = SecureRandom.uuid
+      end
+
+      YAML.dump(
+        prison_uuid_mappings,
+        File.open(prison_uuid_mapping_path, 'w')
```
threedaymonk: TIL that you can write directly to a file from within `YAML.dump`

db/seeds.rb
tatyree: The DB generated one takes precedence.  However, I think it should simply skip that record and alert the user that they need to run the generator, commit the updated uuid mapping and re-run.

lib/tasks/maintenance.rake
tatyree: Your knowledge of anything yamlish unsettles me @threedaymonk...unsettles me deeply... :fearful:

threedaymonk: Well, it's complete lack of knowledge in this case. Mind you, even finding the documentation for the `YAML` module isn't trivial. ([It's actually under `Psych`](http://ruby-doc.org/stdlib-2.2.3/libdoc/psych/rdoc/Psych.html#method-c-dump) because … well … who knows?)

tatyree: Surely the clue is in the name...`Psych`osis?

db/seeds.rb
tatyree: Which is harder than you'd expect with rubocop, which dislikes put statements in `db/seeds.rb`.  As I want it to be very clear what is going on, I'm using a `fail` instead.

spec/spec_helper.rb
```diff
@@ -2,7 +2,10 @@
 require 'simplecov-rcov'
 SimpleCov.formatter = SimpleCov::Formatter::RcovFormatter
 SimpleCov.minimum_coverage 100
-
+SimpleCov.start do
+  add_filter "/.bundle/"
```
threedaymonk: This is specific to individual configuration. I think `/gem/` would be more general.

tatyree: Actually, on reflection, I'll use the example path from the bundler documentation: `vendor/bundle`.

spec/spec_helper.rb
```diff
@@ -2,7 +2,10 @@
 require 'simplecov-rcov'
 SimpleCov.formatter = SimpleCov::Formatter::RcovFormatter
 SimpleCov.minimum_coverage 100
-
+SimpleCov.start do
+  add_filter "/vendor/bundle/"
```
threedaymonk: I still think this is the wrong thing, because it forces a particular location - presumably, your previous configuration of `.bundle` now won't work.

tatyree: @threedaymonk Turns out this isn't necessary at all -- simplecov seems to be ignoring `vendor/bundle` by default.  I hadn't appreciated this before owing to my habit of install gems in `.bundle`.  I'm just removing this altogether.

app/models/prison.rb
```diff
@@ -1,4 +1,7 @@
 class Prison < ActiveRecord::Base
+  class MissingUuidMapping < StandardError
```
dan-palmer: Could be:
```ruby
MissingUuidMapping = Class.new(StandardError)
```

threedaymonk: @dan-palmer That's my preference, too.

db/seeds.rb
```diff
@@ -1,6 +1,21 @@
...
+  prison_uuid_mappings = YAML.load(File.read(prison_uuid_mapping_path)) || {}
+
   Dir[Rails.root.join('db', 'seeds', 'prisons', '*.yml')].each do |path|
     hash = YAML.load(File.read(path))
+    filename = Pathname.new(path).basename.to_s
+
+    next if Prison.find_by(id: prison_uuid_mappings[filename])
+
+    if prison_uuid_mappings[filename].blank?
+      fail Prison::MissingUuidMapping, <<-EOF
```
dan-palmer: `String`'s `:strip_heredoc` method can be used to remove indentation on heredocs.

```ruby
  fail Prison::MissingUuidMapping, 
    <<-INDENTATION_NAZI.strip_heredoc
      #{filename} is missing a UUID mapping. Rerun `rake maintenance:prison_uuids`,
      commit the result and rerun `rake db:seed`.
    INDENTATION_NAZI
```

PR #16, opened 2015-11-12
Add log metadata

EGiataganas:
We added log metadata method to ensure that we'll be able to see
the visit_id throughout the Booking Request Creator process.

### DIFF
--- a/app/services/booking_request_creator.rb
+++ b/app/services/booking_request_creator.rb
@@ -3,10 +3,15 @@ class BookingRequestCreator
     params = build_params(prisoner_step, visitors_step, slots_step)
     Visit.create!(params).tap { |visit|
       PrisonMailer.request_received(visit).deliver_later
+      add_log_metadata visit_id: visit.id
     }
   end
 
 private
+
+  def add_log_metadata(hash)
+    LogStasher.request_context.merge!(hash)
+  end
 
   def build_params(prisoner_step, visitors_step, slots_step)
     [
--- a/spec/services/booking_request_creator_spec.rb
+++ b/spec/services/booking_request_creator_spec.rb
@@ -54,27 +54,29 @@ RSpec.describe BookingRequestCreator do
         slot_option_1: '2015-01-02T09:00/10:00',
         slot_option_2: '2015-01-03T09:00/10:00',
         slot_option_3: '2015-01-04T09:00/10:00'
-      ).and_return double(Visit, id: 2)
+      ).and_return instance_double(Visit, id: 2)
 
-    described_class.new.create! prisoner_step, visitors_step, slots_step
+    subject.create! prisoner_step, visitors_step, slots_step
   end
 
-  it 'emails the prison' do
-    visit = double(Visit, id: 2)
-    allow(Visit).to receive(:create!).and_return(visit)
-    expect(PrisonMailer).to receive(:request_received).with(visit).
-      and_return(mailing)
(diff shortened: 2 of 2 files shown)


spec/services/booking_request_creator_spec.rb
```diff
@@ -54,13 +54,13 @@
         slot_option_1: '2015-01-02T09:00/10:00',
         slot_option_2: '2015-01-03T09:00/10:00',
         slot_option_3: '2015-01-04T09:00/10:00'
-      )
+      ).and_return double(Visit, id: 2)
```
tatyree: I'd like to see a spec that explains why `id` is necessary, please.

threedaymonk: Something like:

```ruby
expect(LogStasher.request_context).to have_attribute(id: 1234)
```

perhaps.

app/services/booking_request_creator.rb
```diff
@@ -3,11 +3,16 @@ def create!(prisoner_step, visitors_step, slots_step)
...
       PrisonMailer.request_received(visit).deliver_later
+      add_log_metadata visit_id: visit.id
     }
   end
 
 private
 
+  def add_log_metadata(hash)
+    LogStasher.request_context.merge!(hash)
+  end
```
tatyree: I'd like to see a couple of unit tests in the spec to give this and the `id` parameters passed by the doubles a bit of context, please.  If I didn't know the history, I would not find it immediately clear why we were logging something that would potentially be in the database.

spec/services/booking_request_creator_spec.rb
tatyree: @threedaymonk :+1:

EGiataganas: :+1:

spec/services/booking_request_creator_spec.rb
```diff
@@ -54,13 +54,13 @@
...
 
     described_class.new.create! prisoner_step, visitors_step, slots_step
   end
 
   it 'emails the prison' do
-    visit = double(Visit)
+    visit = double(Visit, id: 2)
     allow(Visit).to receive(:create!).and_return(visit)
```
tatyree: Could you DRY this and the visit declaration out, please? I'd suggest something like: 

```
context 'emailing and logging' do 
  let(:visit) { double(Visit, id: 2) }
  before do
    allow(Visit).to receive(:create!).and_return(visit)
  end

  it 'emails the prison' do 
    expect(PrisonMailer).to receive(:request_received).with(visit)...
   ...

 it 'logs the visit id' do 
   expect(LogStasher.request_context).to match(visit_id: 2) 
```

EGiataganas: @tatyree Sounds great. I will also do something like

```
    after do
      described_class.new.create! prisoner_step, visitors_step, slots_step
    end
```

spec/services/booking_request_creator_spec.rb
dan-palmer: Should be using `instance_double` here.

spec/services/booking_request_creator_spec.rb
```diff
@@ -33,6 +33,8 @@
     double(Mail::Message, deliver_later: nil)
   }
 
+  subject { described_class.new }
```
threedaymonk: I think this is implicit.

PR #17, opened 2015-11-13
Rationalise all stylesheets to SCSS

threedaymonk:
Some were in SCSS, others were in SASS. SCSS is more familiar, so let's stick with that.


SteveMarshall:
Looks good to me; I wonder if there's anything we can do to get git to recognise that these are all moves and then changes within the moved files, though?

threedaymonk:
@SteveMarshall there is: set `diff.renames = true`. Git doesn't record moves, but detects them. There's not much we can do to make GitHub detect them, however.

threedaymonk:
To clarify that a bit, if this commit is restructured so that the rename and the changes are separate commits, the default behaviour of Git will still list the rename as deletions and additions, unless `diff.renames` is set.

In this case, the files don't have any history beyond the fact that they were imported from the old PVB (where their history was already mostly obscured by the "open sourcing" dump) so I don't think it helps at all.

SteveMarshall:
Yeah, I figured that was the case, but thought I should mention it just in case you knew of some magic trick to make renames more obvious to GitHub (I certainly don't). Given that, I'll merge…

PR #18, opened 2015-11-13
Process visit requests

threedaymonk:
> As prison staff, I want to be able to process visitors' bookings so that they know if they can visit or not

This does not include sending emails or additional visitors (those are separate stories).


app/assets/stylesheets/confirmations.scss
```diff
@@ -8,14 +8,17 @@
...
 
   }
-  
+
   form .validation-summary h2 {
     border: none;
   }
 }
 
-.email {
```
dan-palmer: :+1:

app/services/booking_responder.rb
```diff
@@ -0,0 +1,26 @@
...
+      accept!
+    else
+      reject!
+    end
+  end
+
+private
+
+  attr_reader :booking_response
+  delegate :visit, to: :booking_response
```
dan-palmer: Although this delegation is declared beneath `private` it will actually expose the `visit` instance publicly.

spec/features/process_a_request_spec.rb
```diff
@@ -0,0 +1,31 @@
+require 'rails_helper'
+
+RSpec.feature 'Processing a request', js: true do
+  include ActiveJobHelper
+
+  let(:vst) { create(:visit) }
```
dan-palmer: I see what happened here :laughing:

app/models/booking_response.rb
```diff
@@ -0,0 +1,30 @@
+class BookingResponse
```
dan-palmer: Nice! Liking this approach a lot.

spec/features/process_a_request_spec.rb
threedaymonk: Yeah, it's unfortunate. I also considered `vizit` (à la `klass`) but thought it would be even worse!

app/services/booking_responder.rb
threedaymonk: To be honest, I don't think it really matters all that much – I see `private` in Ruby as mainly advisory – but I've added `private :visit` below to be sure.

PR #19, opened 2015-11-16
Deliver email acknowledgment of booking request

tatyree:
So that a visitor knows the prison has received their request. This
includes spam and bounce reset handling.

TODO:

 * wire up links in email
 * alert user when to expect reply (from PrisonSchedule in old app)


app/models/prison.rb
```diff
@@ -6,6 +6,8 @@ class Prison < ActiveRecord::Base
   validates :estate, :name, :nomis_id, :slot_details, presence: true
   validates :enabled, inclusion: { in: [true, false] }
 
+  alias_attribute :email, :email_address
```
threedaymonk: I've specifically been trying to use `email_address` everywhere to avoid the ambiguity of email-as-address and email-as-document, so please don't do this.

app/helpers/visit_helper.rb
```diff
@@ -0,0 +1,5 @@
+module VisitHelper
+  def when_to_expect_reply
+    # TODO
```
threedaymonk: If this helper has nothing in it, we can drop it entirely.

app/mailers/concerns/addresses.rb
```diff
@@ -0,0 +1,15 @@
+module Addresses
+  include ActiveSupport::Concern
+
+  def prison_mailbox_email
+    @visit.prison_email
```
threedaymonk: I'm confused by the naming here. What does 'mailbox' mean? Why is it not `prison_email_address`?

app/mailers/concerns/autoresponder.rb
```diff
@@ -0,0 +1,9 @@
+module Autoresponder
+  include ActiveSupport::Concern
+
+  def autorespond(parsed_email)
+    mail(from: noreply_address,
+         to: parsed_email.from.to_s,
+         subject: 'This mailbox is not monitored')
+  end
```
threedaymonk: Does this actually do anything? I saw it in the old app, but couldn't understand its mechanism of operation.

app/models/prison.rb
tatyree: Understood.

app/mailers/concerns/addresses.rb
tatyree: No idea.  That's ported from the old app.  Will change.

app/helpers/visit_helper.rb
tatyree: I intended to port PrisonSchedule at the same time, but then realised we don't need it yet.  We might not need it at all.  I'll drop this until such time as we decide how we're going to handle this.

app/mailers/concerns/autoresponder.rb
tatyree: I don't think so.  It's an artefact of me trying to pull all the existing logic from the old app.  I'll drop it.

tatyree:
Changing back to `in progress` because it seems sensible to get PrisonSchedule and PrisonDay ported over in the same commit.

app/mailers/concerns/ensure_quoted_printable.rb
```diff
@@ -0,0 +1,18 @@
...
+# see section 2.1.1 for more info.
+#
+# It will break lines with a new line when the limit is reached,
+# this can break links.
+#
+# This module when included in a Mailer will ensure the
+# Content-Transfer-Encoding header is set to quoted-printable
+# which results in correctly formatted html and text emails.
+
+module EnsureQuotedPrintable
```
threedaymonk: We no longer need this, because we aren't going to be generating URLs longer than 998 characters.

app/mailers/visitor_mailer.rb
```diff
@@ -0,0 +1,27 @@
...
+      reply_to: prison_email_address,
+      to: recipient,
+      subject: default_i18n_subject(
+        receipt_date: format_date_of_visit(first_date)
+      )
+    )
+  end
+
+  delegate :recipient, :prison_email_address,
+    :first_date, to: :visit
```
threedaymonk: Minor thing, but this could be on one line.

app/models/visit.rb
```diff
@@ -9,12 +9,26 @@ class Visit < ActiveRecord::Base
...
+
+  def total_number_of_visitors
+    additional_visitors.count + 1
+  end
 
   def prisoner_full_name
     [prisoner_first_name, prisoner_last_name].join(' ')
   end
 
+  def prisoner_last_name_initial
```
threedaymonk: do we ever use this in a context that isn't something like "Paul B"? If that's the case, it might be more intentional to have something like `prisoner_anonymized_name` instead of appending the initial in views.

tatyree: I'm not sure.  I had assumed we did and that this was the exception.  However, that is probably not a safe assumption...

app/views/visitor_mailer/request_acknowledged.html.erb
```diff
@@ -0,0 +1,67 @@
+<p>
+Dear <%= @visit.visitor_first_name %>,
```
threedaymonk: These bits of text all need to move to a locale file.

app/views/visitor_mailer/request_acknowledged.html.erb
```diff
@@ -0,0 +1,67 @@
...
+<p>
+You have requested a visit to <%= @visit.prison_name %> for <%= pluralize(@visit.total_number_of_visitors, 'person', 'people') %>
+</p>
+<p>
+<% @visit.slots.each_with_index do |slot, index| %>
+  <% unless index == 0 %>
+    <br/>
+  <% end %>
+  <% if index == 0 %>
+    First choice:
```
threedaymonk: For reference, I've punted on the First/Alternative distinction and made them all 'Choice 1/2/3' - I think we'll eventually want to improve that, but I wanted to avoid a whole load of ad-hoc implementations

config/application.rb
```diff
@@ -29,5 +29,7 @@ class Application < Rails::Application
     config.time_zone = 'London'
 
     config.active_record.raise_in_transactional_callbacks = true
+    config.unsubscribe_url =
+      '<https://www.prisonvisits.service.gov.uk/unsubscribe>'
```
threedaymonk: This needs an accompanying endpoint. In the old app, it's just a file in `public/`.

spec/features/request_a_visit_spec.rb
```diff
@@ -44,6 +44,8 @@
     expect(page).to have_text('Your request is being processed')
 
     expect(ActionMailer::Base.deliveries[0].subject).
+      to match(/we\'ve received your visit request for \w+ \d+ \w+\z/)
```
threedaymonk: 1. you don't need to escape `'`, but
2. let’s use real quotes!

config/application.rb
tatyree: This one needs a rethink.  @dan-palmer and I just looked at it on the old app and it doesn't give you the option of unsubscribing. It only tells you why you're receiving the email.  It also has a strange workflow in that it is being served by the existing app, but is also extracted as a static page into `/public` by a rake task.  Dan thinks this is so it can be served external to the app.  

I'll raise a separate ticket for this now.

config/environments/production.rb
```diff
@@ -1,10 +1,12 @@
...
-    domain: ENV['SMTP_DOMAIN'],
-    authentication: :login,
-    enable_starttls_auto: true
+    user_name:             ENV['SMTP_USERNAME'],
+    password:              ENV['SMTP_PASSWORD'],
+    address:               ENV['SMTP_HOSTNAME'],
+    port:                  ENV['SMTP_PORT'],
+    domain:                ENV['SMTP_DOMAIN'],
+    authentication:        :login,
+    enable_starttls_auto:  true
```
threedaymonk: I'd prefer not to line things up: it ends up making commits noisier and merges harder than they'd otherwise be.

tatyree: Ok, ok, I'll stop :(

app/models/visit.rb
```diff
@@ -10,7 +10,17 @@ class Visit < ActiveRecord::Base
     :processing_state,
     presence: true
 
-  delegate :email_address, to: :prison, prefix: true
+  validates :email_override,
+    inclusion: %w[ bounced spam_reported],
```
threedaymonk: Uneven `[]` spacing.

threedaymonk: And for this validation to work, I think it needs to be:

```ruby
validates :email_override,
  include: { in: %w[ bounced spam_reported ] }
```

app/models/visit.rb
```diff
@@ -28,6 +38,14 @@ def prisoner_full_name
     [prisoner_first_name, prisoner_last_name].join(' ')
   end
 
+  def prisoner_last_name_initial
+    prisoner_last_name[0]
+  end
```
threedaymonk: Can this now be private?

app/models/visit.rb
```diff
@@ -28,6 +38,14 @@ def prisoner_full_name
...
+
+  def anonymised_prisoner_name
```
threedaymonk: Style note: let's use US spelling, because everything else already does (`initialize` etc.).

app/views/visitor_mailer/request_acknowledged.html.erb
```diff
@@ -0,0 +1,70 @@
...
+</p>
+<h1>
+  <%= t('.details') %>
+</h1>
+<p>
+  <%= t('.request',
+        prison: @visit.prison_name,
+        number_of_visitors: pluralize(@visit.total_number_of_visitors,
+                                      'person',
+                                      'people')
```
threedaymonk: These English words don't belong in the template. If you use `count` as the variable, you can use `one`/`other` in the locale file to deal with plurals.

spec/services/sendgrid_api_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  context 'without sendgrid configured' do
+    around do |example|
+      smtp_settings = Rails.configuration.action_mailer.smtp_settings
+      Rails.configuration.action_mailer.smtp_settings = {}
+      example.run
+      Rails.configuration.action_mailer.smtp_settings = smtp_settings
+    end
+  end
+end
```
threedaymonk: This spec has contexts, but doesn't actually assert anything!

spec/services/spam_and_bounce_resets_spec.rb
```diff
@@ -0,0 +1,86 @@
+require 'rails_helper'
+
+RSpec.describe SpamAndBounceResets do
+  let(:visit) { build(:visit) }
+  subject { described_class.new(visit) }
+
+  describe '.perform_resets' do
+    context 'override_email_check == true' do
```
threedaymonk: Use something like 'when' to start contexts so that they form comprehensible names when glued together. (Personally, I'd prefer English like 'is true' or 'is false' to '== true', as well.)

spec/models/visit_spec.rb
```diff
@@ -1,7 +1,19 @@
...
+
+    specify do
+      is_expected.
+        to validate_inclusion_of(:email_override).
+        in_array(%w[ bounced spam_reported ])
+    end
+
+    specify do
+      is_expected.to allow_value(nil).for(:email_override)
+    end
```
threedaymonk: This could just be:

```ruby
it { is_expected.to allow_value(nil).for(:email_override) }
```

which is, IMO, a little more readable.

app/models/visit.rb
tatyree: I believe you'll find the OED accepts both :trollface:

threedaymonk: It does, but I'm appealing to Webster here :us:

spec/services/sendgrid_api_spec.rb
tatyree: Oopsie.  Artefact left over after breaking them out into separate methods.

spec/services/spam_and_bounce_resets_spec.rb
tatyree: Good suggestion, thanks.

PR #20, opened 2015-11-17
Add and use some fluent email matchers

threedaymonk:
This is more pleasant and intention-revealing than manipulating `ActionMailer::Base.deliveries` by hand.

The error messages could perhaps be better, but I don't think that detracts from the improvement this already provides.


spec/features/request_a_visit_spec.rb
```diff
@@ -3,9 +3,7 @@
 RSpec.feature 'Booking a visit', js: true do
   include ActiveJobHelper
 
-  before do
-    create(:prison, name: 'Reading Gaol')
-  end
+  let!(:prison) { create(:prison, name: 'Reading Gaol') }
```
dan-palmer: If i recall correctly the default name for the prison returned by factory girl will be 'Reading Gaol'.

spec/support/matchers/email_matchers.rb
```diff
@@ -0,0 +1,22 @@
...
+  match do |address|
+    emails =
+      ActionMailer::Base.deliveries.select { |e| e.to.include?(address) }
+
+    emails.select! { |e| e.subject.match(@subject) } if @subject
+    emails.select! { |e| e.text_part.body.match(@body) } if @body
+    emails.any?
+  end
+
+  chain :with_subject do |subject|
```
dan-palmer: With <a href="https://github.com/rspec/rspec-expectations/pull/644">this</a> change in RSpec 3.2+ we can shorten the `chain`'s to something like this:

```ruby
chain :with_subject, :subject
chain :with_body, :body
chain :and_body, :body
```

spec/features/request_a_visit_spec.rb
```diff
@@ -43,8 +41,10 @@
 
     expect(page).to have_text('Your request is being processed')
 
-    expect(ActionMailer::Base.deliveries[0].subject).
-      to match(/\AVisit request for Oscar Wilde on \w+ \d+ \w+\z/)
+    expect(prison.email_address).
+      to receive_email.
```
dan-palmer: Liking the matcher :+1:

spec/support/matchers/email_matchers.rb
dan-palmer: Don't think they don't get referenced as instance variables in the `match` block with the shorthand approach.

spec/features/request_a_visit_spec.rb
threedaymonk: It will, but I wanted it to be explicit in the feature. (I don't think it's a good default, and I think we'll have to change it eventually.)

spec/support/matchers/email_matchers.rb
threedaymonk: Very nice.

spec/features/request_a_visit_spec.rb
dan-palmer: Oki doki, thats cool with me.

dan-palmer:
:+1: :ship: :boom:

PR #21, opened 2015-11-17
Remove impossible and useless migration

threedaymonk:
Adding a non-null field with no default in one step does not work unless the database is empty. The normal solution to this would be to perform the operation in two steps. However, in this case, the next migration will delete the table entirely (because we do not yet have any useful data) so it's effectively useless.


dan-palmer:
Totes agree, this shouldn't be in the app.

:ship:

PR #25, opened 2015-11-18
Explain to the user why they received this email

tatyree:
After discussion, we decided `unsubscribe` was likely a bad name for
this page.  However, this is the from in which it exists on the current
app, so I am porting it as-is.  That said, there is an outstanding
ticket for the wording and intent to be reveiwed.

### DIFF
--- a/app/views/pages/unsubscribe.html.erb
+++ b/app/views/pages/unsubscribe.html.erb
@@ -1,0 +1,11 @@
+<% content_for :header, "Why did I receive this email?" %>
+<p>
+<%= t('.you_received_an_email',
+      email: 'no-reply@email.prisonvisits.service.gov.uk') %>
+</p>
+<p>
+  <%= t('.why') %>
+</p>
+<p>
+  <%= t('.confirmation') %>
+</p>
--- a/config/application.rb
+++ b/config/application.rb
@@ -29,7 +29,5 @@ module PrisonVisits
     config.time_zone = 'London'
 
     config.active_record.raise_in_transactional_callbacks = true
-    config.unsubscribe_url =
-      '<https://www.prisonvisits.service.gov.uk/unsubscribe>'
   end
 end
--- a/spec/features/unsubscribe_spec.rb
+++ b/spec/features/unsubscribe_spec.rb
@@ -1,0 +1,11 @@
+require 'rails_helper'
+
+RSpec.feature 'Booking a visit', js: true do
+  include ActiveJobHelper
+
+  scenario 'happy path' do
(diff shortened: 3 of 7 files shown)


app/views/pages/unsubscribe.html.erb
```diff
@@ -0,0 +1,11 @@
+<% content_for :header, "Why did I receive this email?" %>
+<p>
+  <%= t('.you_received_an_email') %>
+    no-reply@email.prisonvisits.service.gov.uk
```
threedaymonk: This interpolation is naïve, because it assumes that the email address always goes at the end of the sentence, which isn't always valid. I don't know about Welsh specifically, but it doesn't work in German or Japanese. Generally speaking, one should never form sentences by concatenation, but only via placeholders, _viz._:

```erb
<%= t('.you_received_an_email', from: no-reply@email.prisonvisits.service.gov.uk) %>
```

spec/features/unsubscribe_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+
+RSpec.feature 'Booking a visit', js: true do
+  include ActiveJobHelper
+
+  scenario 'happy path' do
+    visit unsubscribe_path
+    expect(page).to have_text('Why did I receive this email?')
+    expect(page).to have_text('You’ve received an email')
+    expect(page).to have_text('because you requested a prison visit.')
+    expect(page).to have_text('We’ll confirm your visit by email')
```
threedaymonk: This seems an excessive number of assertions for something so trivial!

tatyree: Just bein' thorough...

spec/features/unsubscribe_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'rails_helper'
+
+RSpec.feature 'Booking a visit', js: true do
+  include ActiveJobHelper
```
threedaymonk: You don't need this unless you're looking at email headers/bodies.

config/application.rb
```diff
@@ -29,7 +29,5 @@ class Application < Rails::Application
     config.time_zone = 'London'
 
     config.active_record.raise_in_transactional_callbacks = true
-    config.unsubscribe_url =
-      '<https://www.prisonvisits.service.gov.uk/unsubscribe>'
```
threedaymonk: We should also set this address in the `List-Unsubscribe` header when we send emails. That will probably have to be in a `before_action` in order for the URL method to work.

PR #27, opened 2015-11-20
Extract full/anonymised naming to concern

threedaymonk:
We're doing this in a number of places, so it's nice to avoid repetition, even if it does raise the spectre of gratuitous metaprogramming.

### DIFF
--- a/app/models/concerns/names.rb
+++ b/app/models/concerns/names.rb
@@ -1,0 +1,22 @@
+module Names
+  # rubocop:disable Metrics/MethodLength
+  def enhance_names(prefix: nil)
+    method_prefix = prefix ? "#{prefix}_" : ''
+
+    define_method "#{method_prefix}full_name" do
+      I18n.t(
+        'formats.name.full',
+        first: public_send("#{method_prefix}first_name"),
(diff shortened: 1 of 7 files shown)


spec/models/concerns/names_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+RSpec.describe Names do
+  context 'without prefix' do
+    subject {
+      described_module = described_class
+      Class.new {
+        attr_accessor :first_name, :last_name
+        extend described_module
+        enhance_names
+      }.new
+    }
```
tatyree: :+1:

app/models/concerns/names.rb
```diff
@@ -0,0 +1,22 @@
...
+
+    define_method "#{method_prefix}anonymized_name" do
+      I18n.t(
+        'formats.name.full',
+        first: public_send("#{method_prefix}first_name"),
+        last: public_send("#{method_prefix}last_name")[0]
+      )
+    end
+  end
+end
```
tatyree: It *might* be meta-overkill (urge-overkill, anyone?), but I like it.

PR #28, opened 2015-11-20
Email visitor when booking is accepted

threedaymonk:


app/mailers/visitor_mailer.rb
```diff
@@ -6,21 +6,29 @@ class VisitorMailer < ActionMailer::Base
...
+      reply_to: visit.prison_email_address,
+      to: visit.recipient,
       subject: default_i18n_subject(
-        receipt_date: format_date_of_visit(first_date)
+        receipt_date: format_date_of_visit(visit.first_date)
       )
     )
   end
 
-  delegate :recipient, :prison_email_address, :first_date, to: :visit
```
threedaymonk: These delegations didn't seem useful when they're not being used anywhere outside this class.

spec/spec_helper.rb
```diff
@@ -2,9 +2,6 @@
 require 'simplecov-rcov'
 SimpleCov.formatter = SimpleCov::Formatter::RcovFormatter
 SimpleCov.minimum_coverage 100
-SimpleCov.start do
-  add_filter '/.bundle/'
-end
```
threedaymonk: If you really need to add `.bundle`, see further down in the file. I suspect that `/gem/` might cover it - if not, change it.

tatyree: Right after we discussed it, I noticed that this one had gone through with an earlier PR.  I'm happy to continue to bundle to `vendor/bundle` which seems to fix it on my system without requiring any configuration.

app/mailers/visitor_mailer.rb
tatyree: Yep.  Lint.

PR #33, opened 2015-11-23
Send emails to visitors and prisons when a booking is rejected

threedaymonk:
This sends an email to a prospective visitor when their booking is rejected, and copies it to the prison for their records.

### DIFF
--- a/app/models/rejection.rb
+++ b/app/models/rejection.rb
@@ -11,4 +11,12 @@ class Rejection < ActiveRecord::Base
   belongs_to :visit
 
   validates :reason, inclusion: { in: REASONS }
+
+  def pvo_possible?
+    pvo_expires_on.present?
+  end
+
+  def vo_will_be_renewed?
+    vo_renewed_on.present?
+  end
 end
--- a/app/models/rejection.rb
+++ b/app/models/rejection.rb
@@ -12,11 +12,11 @@ class Rejection < ActiveRecord::Base
 
   validates :reason, inclusion: { in: REASONS }
 
-  def pvo_possible?
-    pvo_expires_on.present?
+  def privileged_allowance_available?
+    privileged_allowance_expires_on.present?
   end
 
-  def vo_will_be_renewed?
-    vo_renewed_on.present?
+  def allowance_will_renew?
+    allowance_renews_on.present?
   end
 end
--- a/app/controllers/prison/visits_controller.rb
+++ b/app/controllers/prison/visits_controller.rb
@@ -24,8 +24,9 @@ private
     params.
       require(:booking_response).
       permit(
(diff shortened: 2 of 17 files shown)


app/models/rejection.rb
```diff
@@ -11,4 +11,12 @@ class Rejection < ActiveRecord::Base
   belongs_to :visit
 
   validates :reason, inclusion: { in: REASONS }
+
+  def pvo_possible?
```
SteveMarshall: Can we use this as our chance to stop using `pvo` and `vo` acronyms?

threedaymonk: If you can capture it in English, I'm all for it!

threedaymonk: On the other hand, it's actually communicated in the interface as "VO" and "PVO", it's only relevant to prison staff, and it's just a flag/date that gets passed through and has no effect on anything else, so there's an argument that it's in the domain language.

SteveMarshall: A fair point… Although I might argue that the staff would have that domain knowledge but people working on the codebase might not initially?

Simplistically, `vo` could just be substituted with `visit_order`… But I have no idea what the `p` in `pvo` stands for. Maybe @kyliemulholland knows?

threedaymonk: The "P" in PVO is "privileged", but it's still an opaque identifier without context.

kyliemulholland: When a prisoner has an 'enhanced' status, meaning they have above the
standard allocation of visits per month, the extra visits they have are
often exclusively for specific slots. These extra visits are known as PVOs.

Most commonly, in prisons where standard visits happen during the week,
PVOs can be used on weekends. There is no consistent pattern to this
however.

For the purposes of PVB this shouldn't matter, as the staff manage all of
that stuff through NOMIS.

On 24 November 2015 at 10:43, Paul Battley <notifications@github.com> wrote:

> In app/models/rejection.rb
> <https://github.com/ministryofjustice/prison-visits-2/pull/33#discussion_r45719590>
> :
>
> > @@ -11,4 +11,12 @@ class Rejection < ActiveRecord::Base
> >    belongs_to :visit
> >
> >    validates :reason, inclusion: { in: REASONS }
> > +
> > +  def pvo_possible?
>
> The "P" in PVO is "privileged", but it's still an opaque identifier
> without context.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/ministryofjustice/prison-visits-2/pull/33/files#r45719590>
> .
>



-- 
*Kylie Mulholland*
Product Manager: Prison Visits Booking
Ministry of Justice | 102 Petty France | London | SW1H 9AJ

SteveMarshall: @kyliemulholland “manage all of that stuff through NOMIS”… for now! Dun-dun daaa

PR #34, opened 2015-11-24
Lead days

threedaymonk:
It takes each prison some number of days to process a booking. How long this is, and whether or not it includes weekends, depends on the prison.

When offering slots, we should only offer slots that have enough lead time to be responded to in time.


config/initializers/holidays.rb
```diff
@@ -0,0 +1,51 @@
...
+  Date.new(2016, 12, 27),
+  Date.new(2017, 1, 2),
+  Date.new(2017, 4, 14),
+  Date.new(2017, 4, 17),
+  Date.new(2017, 5, 1),
+  Date.new(2017, 5, 29),
+  Date.new(2017, 8, 28),
+  Date.new(2017, 12, 25),
+  Date.new(2017, 12, 26)
+]
```
tatyree: Could we not instantiate these at runtime from an external source or service?

SteveMarshall: In the old app, we maintained a copy of [GOV.UK's bank holidays JSON](https://www.gov.uk/bank-holidays.json), but we probably shouldn't directly couple ourselves to that because GOV.UK “APIs” like that aren't officially supported or guaranteed to not change.

That said, I wouldn't object to us having some sort of periodically-run import thing to our own cache of them, or something like that, so long as it logged an error if it failed to import the JSON…

SteveMarshall: … which I see @threedaymonk replicates below, but instead of maintaining a copy of the JSON, we have a direct mapping to this ruby file. I'm not entirely averse to this, then.

tatyree: I think that would be preferable to a rake task that writes ruby code into a config.

threedaymonk: My proposed workflow is that once a year or so, someone runs the rake task, verifies the changes, and commits them. I don't think it's desirable to automate this. I'll try to explain my reasoning.

1. This data very rarely changes. You'll see that we've got all the public holidays through to the end of 2017, so there's about 24 months before we need to update it.
2. The gov.uk API is not official. Writing something to automate fetching and detecting anomalies seems excessive when it needs updating less than annually: it might change more often than we run it!
3. The API JSON isn't particularly easy to read: it's all rammed together, has information for Scotland and Northern Ireland that is outside the remit of the MOJ, and contains irrelevant cruft like `"bunting": true`. It's much easier to visually check changes in a distilled list that contains only the information we need.
4. We always need this data instantiated, and having it already serialised as Ruby is the simplest way of doing so. It's easy to establish validity, which isn't true of JSON (doesn't even have dates as such) or YAML (too many parse rules). Why have a text-based file and a Ruby file to load them when it's as trivial as this? Equally, we could put it in the database, but that's more complex and less efficient than a linear scan of a couple of dozen dates.

So we could bake something more complicated than this, but I think it would be a waste of time and effort.

threedaymonk: It might be sensible to exclude past dates and add a comment about what each date is, however.

SteveMarshall: I was already aware of 1-3, hence “not entirely averse to this”. #4 seems eminently reasonable to me. @tatyree?

SteveMarshall:
As we discussed, the way we should ensure we never run out of bank holiday data might be to add a check to the health check (when we set that up) that we have >1y of bank holidays that causes a 500. I'll add that to the health check task in the backlog.

tatyree:
@threedaymonk thanks for the explanation.  I hadn't appreciated all the factors in play. 

@SteveMarshall :+1:

PR #35, opened 2015-11-24
Let visitors override spam/bounced email rejection

tatyree:
This improves the reporting of email errors when an address has been reported
by Sendgrid as having bounced or been marked as spam.

It adds a checkbox to the visitor form to allow the visitor to force acceptance
of an email address that failed the Sendgrid checks. The checkbox is only
displayed if validation fails for one of these two reasons. If selected, we
permit the email address, and tell Sendgrid to remove the spam or bounced
categorisation before sending the email.  # Please enter the commit message for
your changes.

### DIFF
--- a/app/models/visitors_step.rb
+++ b/app/models/visitors_step.rb
@@ -3,8 +3,24 @@ class VisitorsStep
   include Person
 
   attribute :email_address, String
+  attribute :override_spam_or_bounce, Boolean
+  attribute :spam_or_bounce, String
+  attribute :spam_or_bounce_has_occurred, Boolean
   attribute :phone_no, String
 
   validates :email_address, presence: true
   validates :phone_no, presence: true, length: { minimum: 9 }
+
+  validate :validate_email
+
+private
+
+  def validate_email
+    checker = EmailChecker.new(email_address, override_spam_or_bounce)
+    unless checker.valid?
+      errors.add :email_address, checker.message
+      @spam_or_bounce_has_occurred = checker.spam_or_bounce_occurred?
+      @spam_or_bounce = checker.error
+    end
+  end
 end
--- a/app/models/visitors_step.rb
+++ b/app/models/visitors_step.rb
@@ -3,9 +3,9 @@ class VisitorsStep
   include Person
 
   attribute :email_address, String
-  attribute :override_spam_or_bounce, Boolean
-  attribute :spam_or_bounce, String
-  attribute :spam_or_bounce_has_occurred, Boolean
+  attribute :override_delivery_error, Boolean
+  attribute :delivery_error_type, String
+  attribute :delivery_error_occurred, Boolean
   attribute :phone_no, String
 
   validates :email_address, presence: true
@@ -16,11 +16,11 @@ class VisitorsStep
 private
 
   def validate_email
-    checker = EmailChecker.new(email_address, override_spam_or_bounce)
+    checker = EmailChecker.new(email_address, override_delivery_error)
     unless checker.valid?
       errors.add :email_address, checker.message
-      @spam_or_bounce_has_occurred = checker.spam_or_bounce_occurred?
-      @spam_or_bounce = checker.error
+      @delivery_error_occurred = checker.delivery_error_occurred?
+      @delivery_error_type = checker.error.to_sym
     end
   end
 end
--- a/app/services/booking_request_creator.rb
+++ b/app/services/booking_request_creator.rb
@@ -38,7 +38,9 @@ private
       visitor_last_name: step.last_name,
       visitor_date_of_birth: step.date_of_birth,
       visitor_email_address: step.email_address,
-      visitor_phone_no: step.phone_no
+      visitor_phone_no: step.phone_no,
+      override_spam_or_bounce: step.override_spam_or_bounce,
+      spam_or_bounce: step.spam_or_bounce
     }
   end
 
--- a/app/services/booking_request_creator.rb
+++ b/app/services/booking_request_creator.rb
@@ -39,8 +39,8 @@ private
       visitor_date_of_birth: step.date_of_birth,
       visitor_email_address: step.email_address,
       visitor_phone_no: step.phone_no,
-      override_spam_or_bounce: step.override_spam_or_bounce,
-      spam_or_bounce: step.spam_or_bounce
+      override_delivery_error: step.override_delivery_error,
+      delivery_error_type: step.delivery_error_type
     }
   end
 
--- a/app/services/email_checker.rb
+++ b/app/services/email_checker.rb
@@ -1,6 +1,7 @@
 class EmailChecker
   extend Forwardable
   def_delegators :SendgridApi, :bounced?, :spam_reported?
+  def_delegator :error, :valid?
 
   def initialize(original_address, override_sendgrid = false)
     @original_address = original_address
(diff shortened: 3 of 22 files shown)


app/models/visitors_step.rb
```diff
@@ -3,8 +3,24 @@ class VisitorsStep
   include Person
 
   attribute :email_address, String
+  attribute :override_email_checks, Boolean
+  attribute :email_overridable, Boolean
+  attribute :email_override, String
```
threedaymonk: Would this work better as something imperative like `override_email`?

app/services/booking_request_creator.rb
```diff
@@ -38,7 +38,9 @@ def visitors_step_params(step)
       visitor_last_name: step.last_name,
       visitor_date_of_birth: step.date_of_birth,
       visitor_email_address: step.email_address,
-      visitor_phone_no: step.phone_no
+      visitor_phone_no: step.phone_no,
+      override_email_checks: step.override_email_checks,
+      email_override: step.email_override
```
threedaymonk: What's the difference between `override_email_checks` and `email_override`?

data/bad_domains.txt
```diff
@@ -0,0 +1,66 @@
...
+yahaoo.com
+yahoomail.com
+yaho.co.uk
+yaho.com
+yahoi.com
+yahooo.com
+yahooco.uk
+yhoo.com
+ucloud.com
+viginmedia.co.uk
```
threedaymonk: Are the pluses meant to be there?

Separately, I'm a bit sad about this whole file. Attempting to enumerate every typo just seems like an absurd thing to do. Do we really want or need this?

app/services/booking_request_creator.rb
tatyree: Yeah, on reflection they are named confusingly.  `email_override` is the category of the error and override_email_checks is the flag.  I'll rename `email_override`.

app/models/visitors_step.rb
tatyree: Probably.

data/bad_domains.txt
tatyree: I'd assumed that Jan had a good reason to put it in, but I did find it a bit absurd, myself.  The additions are down the fact I initially checked in an empty file to make the spec I was working on pass  and theb copied over the list later.

tatyree: Yep, looking at it more closely, the resolv lookup will produce the same outcome.

threedaymonk: Ignore my comment about the pluses - I was confused by GitHub.

SteveMarshall:
LGTM (though it might've been nice for the fixes to be in [a fixup commit](https://robots.thoughtbot.com/autosquashing-git-commits)! @threedaymonk?

app/services/spam_and_bounce_resets.rb
```diff
@@ -17,6 +17,6 @@ def perform_resets
 private
 
   def reset
-    email_override.to_s.inquiry
+    spam_or_bounce.to_s.inquiry
```
threedaymonk: I didn't know about `inquiry`! TIL.

app/models/visitors_step.rb
```diff
@@ -3,8 +3,24 @@ class VisitorsStep
...
+
+  validate :validate_email
+
+private
+
+  def validate_email
+    checker = EmailChecker.new(email_address, override_spam_or_bounce)
+    unless checker.valid?
+      errors.add :email_address, checker.message
+      @spam_or_bounce_has_occurred = checker.spam_or_bounce_occurred?
```
threedaymonk: Can we make these both `spam_or_bounce_occurred` for consistency?

SteveMarshall:
Oh, although… Do we need the empty/unchanged `spec/validators/shared_resolv_context.rb`?

app/models/visitors_step.rb
```diff
@@ -3,8 +3,24 @@ class VisitorsStep
   include Person
 
   attribute :email_address, String
+  attribute :override_spam_or_bounce, Boolean
+  attribute :spam_or_bounce, String
+  attribute :spam_or_bounce_has_occurred, Boolean
```
threedaymonk: A thought occurs to me somewhat late in the day: perhaps we should call it something like `*delivery_problem*` rather than `*spam_or_bounce*`. These would then be:

* `override_delivery_problem`
* `delivery_problem_type`
* `delivery_problem_occurred`

And I think that would explain itself better.

app/services/email_checker.rb
```diff
@@ -0,0 +1,89 @@
...
+  extend Forwardable
+  def_delegators :SendgridApi, :bounced?, :spam_reported?
+
+  def initialize(original_address, override_sendgrid = false)
+    @original_address = original_address
+    @parsed = parse_address(original_address)
+    @override_sendgrid = override_sendgrid
+  end
+
+  def error
```
dan-palmer: Suggestion to simplify the `error` method.

Firstly set `compute_error` to return a default `:valid` symbol when no delivery problems have been encountered:

```ruby
def compute_error
+    return :unparseable unless parsed
+    return :domain_dot if domain_dot_error?
+    return :malformed unless well_formed_address?
+    return :no_mx_record unless mx_records?
+    unless override_sendgrid?
+      return :spam_reported if spam_reported?(parsed.address)
+      return :bounced if bounced?(parsed.address)
+    end
+    :valid
+  end
```

We can then cache the `compute_error` return value and use active supports inquiry method so we can just delegate the `valid?` predicate method to the instances `:error` method - which now returns a string inquiry object.

```ruby
def error
  @error ||= compute_error.to_s.inquiry
end

delgate :valid?, to: :error
```

Copied the class and updated as a <a href="https://gist.github.com/dan-palmer/bd5ca41c89ba9c878455">gist</a> for copy and paste if you like the change.

threedaymonk: I think this is a really nice suggested change.

app/services/spam_and_bounce_resets.rb
tatyree: I didn't either.  I thought you added it...

app/models/visitors_step.rb
tatyree: Yep.  On it.

app/services/email_checker.rb
tatyree: I agree.  Working on it now.

app/models/visitors_step.rb
tatyree: TBH, I wasn't entirely happy with the names.  This seems clearer.

tatyree:
@SteveMarshall re `shared_resolv_context.rb`...something strange has happened there. I had several shared_context definitions in that.  Might have stashed and forgotten to pop.  I'll check.

app/services/email_checker.rb
tatyree: Oh how I wish there weren't multiple, clashing, and incompatible versions of `delegate`.

spec/rails_helper.rb
```diff
@@ -18,6 +18,8 @@
 
 ActiveRecord::Migration.maintain_test_schema!
 
+Dir[File.expand_path("../support/**/*.rb", __FILE__)].each { |f| require f }
```
threedaymonk: You don't need this: put the helper in `spec/support/helpers` instead and it will autoload.

PR #37, opened 2015-11-25
Add specs for emails

threedaymonk:
This ensures that we have no missing localisation strings or interpolated values, and checks expected parts of the email body. It's much easier to find and fix errors here than in the feature specs, where
the error is hidden by the delayed mailer.

The dates in some of the specs are coupled to the default slots in the factory, which isn't great, but I think it's probably no worse than laboriously building up a graph in the spec itself.

### DIFF
--- a/spec/mailers/prison_mailer/booked_spec.rb
+++ b/spec/mailers/prison_mailer/booked_spec.rb
@@ -1,0 +1,24 @@
+require 'rails_helper'
+require 'mailers/shared_mailer_examples'
+
+RSpec.describe PrisonMailer, '.booked' do
+  let(:visit) {
+    create(
+      :booked_visit,
+      prisoner_first_name: 'Arthur',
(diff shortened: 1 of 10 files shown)


spec/mailers/prison_mailer/booked_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+  subject { described_class.booked(visit) }
+
+  before do
+    ActionMailer::Base.deliveries.clear
+  end
+
+  include_examples 'template checks'
+
+  it 'sends an email confirming the booking' do
+    expect(subject.subject).
```
threedaymonk: feels-bad-man

SteveMarshall:
LGTM! @tatyree?

tatyree:
±1 @SteveMarshall.  Can you merge. I'm out at my appointment right now.

PR #39, opened 2015-11-25
Add smoke tests

tatyree:
This is an adaption of the DanP's smoke tests from the original version
of the application.  The main things that have changed are the end points and the way cancellations are handled.

### DIFF
--- a/README.md
+++ b/README.md
@@ -259,30 +259,14 @@ same effect as running it once.
 
 The files can be edited by anyone with access to this repository. The
 [YAML](http://en.wikipedia.org/wiki/YAML) specification is complex and fraught
-`SMOKE_TEST_EMAIL_LOCAL_PART`@`SMOKE_TEST_EMAIL_DOMAIN`, it will not
-send that email to the prison, but will direct all mail the
 with edge cases, so be careful.
 
 #### Prison visibility
-To start the smoke tests, set the environment variables and run
-`ruby ./smoke_test/smoke_test.rb`.
 
 All known prisons should exist in the data files. If a prison is not in scope
 of the service, it should be disabled.
 
-##### Smoke Test Steps:
-  1. Starts the form
-  2. Fills in prisoner details page
-  3. Fills in visitor details page for one visitor
-  4. Selects three available days on the slot page
-  5. Submits the booking on the check your request page
-  6. Checks that the user received a booking receipt email
-  7. Checks that the prison received a booking request email
-  8. Visits the process booking page and confirms the booking
 To disable visit requests to a prison, set `enabled` to `false`.
-  10. Checks that the visitor received a confirmation email
-  11. Checks the status of the visit and cancels it
-  12. Checks that the prison receives a booking cancellation email
 
 ```yaml
   nomis_id: SFI
@@ -291,7 +275,7 @@ To disable visit requests to a prison, set `enabled` to `false`.
   ...
 
 ```
-SMOKE_TEST_EMAIL_LOCAL_PART
+
 #### Weekly visiting slots
 
 Slots are defined per prison via a weekly schedule. Only days listed here with
@@ -306,12 +290,12 @@ slots:
   - 1350-1450 # creates a 1 hour slot every Wednesday from 1:50pm
   sat:
   - 0900-1100 # creates a 2 hour slot every Saturday from 9am
-SMOKE_TEST_EMAIL_DOMAIN
+  - 1330-1530 # creates a 2 hour slot every Saturday from 1:30pm
 ```
 
 #### Slot anomalies
 
-##### Required environment variables for the smoke test:
+Use this to make exceptions to the weekly schedule.
 
 When a day is found in `slot_anomalies` the whole day is replaced with this
 data. Therefore if the weekday usually contains multiple slots and only a
@@ -322,10 +306,10 @@ slot_anomalies:
   2015-01-10:
   - 0930-1130 # replaces Saturday 10 January 2015 with only one slot at 9:30am
 ```
-SMOKE_TEST_EMAIL_LOCAL_PART
-SMOKE_TEST_EMAIL_DOMAIN
-SMOKE_TEST_EMAIL_PASSWORD
-SMOKE_TEST_APP_HOST
+
+#### Non-bookable days
+
+Use this to remove specified dates, such as staff training days or Christmas
 Day from the schedule. Public holidays are already excluded by default:
 visiting can be enabled by adding them as anomalies, above.
 
@@ -333,21 +317,19 @@ This overrides `slots`.
 
 ```yaml
 unbookable:
-SMOKE_TEST_EMAIL_HOST
+- 2015-12-25 # removes any slots from 25 December 2015
 ```
-### `SMOKE_TEST_EMAIL_LOCAL_PART`, `SMOKE_TEST_EMAIL_DOMAIN`
 
-These determine the email address that will trigger the smoke tests on
-the application. See note above.
+**Note** If an enabled prison does not have any unbookable dates, please
+make sure you represent this in the yaml as:
 
 ```yaml
 unbookable: []
-### `SMOKE_TEST_EMAIL_HOST`, `SMOKE_TEST_EMAIL_PASSWORD`
+```
 
-These are the details for the IMAP host that allows the smoke test to
-check that the emails have been delivered.
+If you do not, the specs will fail.
 
-### `SMOKE_TEST_APP_HOST`
+#### Response times
 
 Set the amount of full working days which booking staff have to respond to each
 request. The default is 3 days.
@@ -373,7 +355,7 @@ works_weekends: true
 
 When the service links to [Prison
 Finder](https://www.justice.gov.uk/contacts/prison-finder), it turns the prison
-This is the url of the particular version of the application that the
+name into part of the URL. For example, 'Drake Hall' becomes
 [drake-hall](https://www.justice.gov.uk/contacts/prison-finder/drake-hall).
 
 When the Prison Finder link does not simply match the prison name in lower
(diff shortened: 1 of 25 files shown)


tatyree:
Please do go ahead an review, but bear in mind that I am adding in the cancellation mailer right now.

app/mailers/mail_utilities/smoke_test_email_check.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    private
+
+    def smoke_test_email_regex
+      %r{
+        \A             # match from the start of the string
+        #{local_part}
+        \+             # google address alias see: https://support.google.com/mail/answer/12096
+        [0-9a-z\-]{36} # RFC 4122 uuid extension
+        @              # its an email!
```
threedaymonk: "it's" :stuck_out_tongue_closed_eyes:

app/mailers/mail_utilities/smoke_test_email_check.rb
```diff
@@ -0,0 +1,29 @@
...
+        \z             # match until the end of the string
+      }x
+    end
+
+    def local_part
+      Regexp.escape(Rails.configuration.smoke_test_email_local_part)
+    end
+
+    def domain
+      Regexp.escape(Rails.configuration.smoke_test_email_domain)
```
threedaymonk: It might be nicer to have `Rails.configuration.smoke_test` (which could perhaps be delegated to) as an object (`OpenStruct`?) with these two attributes.

app/mailers/prison_mailer.rb
```diff
@@ -33,4 +37,17 @@ def rejected(visit)
            prisoner: visit.prisoner_full_name
          )
   end
+
+private
+
+  def do_not_send_to_prison
+    message.to = visit.contact_email_address
+    message.delivery_method.settings.
+      merge!(Rails.configuration.action_mailer.smtp_settings)
```
threedaymonk: Is it correct that `action_mailer.smtp_settings` has precedence over the message's settings?

app/mailers/mail_utilities/smoke_test_email_check.rb
```diff
@@ -0,0 +1,29 @@
+module MailUtilities
```
threedaymonk: Do we need this module? The additional hierarchy doesn't seem helpful.

config/environments/development.rb
```diff
@@ -15,4 +15,10 @@
   config.assets.raise_runtime_errors = true
 
   config.mx_checker = MxChecker::Dummy.new
+  config.smoke_test_email_local_part = ENV.fetch(
+    'SMOKE_TEST_EMAIL_LOCAL_PART',
+    'smoke-test')
+  config.smoke_test_email_domain = ENV.fetch(
+    'SMOKE_TEST_EMAIL_DOMAIN',
+    'example.com')
```
threedaymonk: The wrapping here feels a bit odd. Perhaps this would be better:

```ruby
config.smoke_test_email_domain =
  ENV.fetch('SMOKE_TEST_EMAIL_DOMAIN', 'example.com')
```

config/environments/production.rb
```diff
@@ -35,4 +35,6 @@
   config.logstasher.logger_path = "#{Rails.root}/log/logstash_#{Rails.env}.json"
 
   config.mx_checker = MxChecker.new
+  config.smoke_test_email_local_part = ENV['SMOKE_TEST_EMAIL_LOCAL_PART']
+  config.smoke_test_email_domain = ENV['SMOKE_TEST_EMAIL_DOMAIN']
```
threedaymonk: The app will blow up without these, won't it? Should they also use `fetch`?

threedaymonk:
Please add the new environment variables to the README.

app/mailers/mail_utilities/smoke_test_email_check.rb
tatyree: As @dan-palmer's editor, I take full responsibility.  Inconsistent application, however--you let @kyliemulholland get away with it yesterday :see_no_evil:

config/environments/production.rb
tatyree: It will.  My reasoning here was that both prod and preprod have smoke tests run against them regularly and so should probably not be run without the variables set.  On the other hand, we always use the same addresses, so I think it makes as much sense to got ahead and set them.

tatyree:
Re. ENV variables: partially done in the latest fixup, @threedaymonk

SteveMarshall:
It's probably a bit late now, but it might've been nice (as I think we discussed around #5?) to see what was ported from the old app and how you've adapted it?

tatyree:
@SteveMarshall it's all pretty atomic, so I could unpick it and do it again.  However, other than changing a few paths in the smoke tests and adding a couple of smoke-test-detectors to the main app, there was almost no adaption.

README.md
```diff
@@ -205,3 +205,60 @@ build information to be returned by `/ping.json`. e.g.:
...
+  - [x] Fills in visitor details page for one visitor
+  - [x] Selects three available days on the slot page
+  - [x] Submits the booking on the check your request page
+  - [x] Checks that the user received a booking receipt email
+  - [x] Checks that the prison received a booking request email
+  - [x] Visits the process booking page and confirms the booking
+  - [x] Checks that the prison received a copy of the confirmation email
+  - [x] Checks that the visitor received a confirmation email
+  - [x] Checks the status of the visit and cancels it
+  - [x] Checks that the prison receives a booking cancellation email
```
SteveMarshall: Why checkboxes? As this all works now, it might be nice if this were an ordered list…

smoke_test/steps/prisoner_page.rb
```diff
@@ -0,0 +1,36 @@
...
+    class PrisonerPage < BaseStep
+      PAGE_PATH = '/request'
+
+      def validate!
+        if page.current_path != PAGE_PATH
+          fail "expected #{PAGE_PATH}, got #{page.current_path}"
+        end
+      end
+
+      # rubocop:disable Metrics/AbcSize
```
SteveMarshall: Can we re-enable this after `complete_step`?

SteveMarshall:
@tatyree Fair enough: the way I read the commit message just sounded like it was a bit more of adaptation than it now sounds :+1:

app/mailers/prison_mailer.rb
tatyree: Oh...I just realised that this was intended to override the messagelabs settings in the old app.

smoke_test/steps/prisoner_page.rb
tatyree: Done, but it didn't pick it up. Check full changes.

SteveMarshall: :+1:

app/mailers/mail_utilities/smoke_test_email_check.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def smoke_test_email_regex
+    %r{
+    \A             # match from the start of the string
+    #{local_part}
+    \+             # google address alias see: https://support.google.com/mail/answer/12096
+    [0-9a-z\-]{36} # RFC 4122 uuid extension
+    @              # it's an email!
+    #{domain}
+    \z             # match until the end of the string
```
threedaymonk: This lot (inside `%r{` … `}`) should be indented one more step.

smoke_test/steps/process_visit_request_page.rb
```diff
@@ -0,0 +1,25 @@
+module SmokeTest
+  module Steps
+    class ProcessVisitRequestPage < BaseStep
+      PAGE_PATH = %r{/prison/visits/(\w|-){36}}
```
threedaymonk: I think `[0-9a-f-]` would be clearer and more accurate than `(\w|-)`

smoke_test/steps/base_step.rb
```diff
@@ -0,0 +1,41 @@
...
+        puts "Step:#{step_name}"
+        @state = state
+      end
+
+      def validate!
+        true
+      end
+
+      def complete_step
+        fail NotImplementedError
```
dan-palmer: Inheriting from this class requires this method to be overriden, as a result of this we end up implementing a couple of empty methods. A better approach is probably to just have the superclass have the empty method and don't require the override & then remove the empty methods in the relevant step implemtations.

threedaymonk: I have a feeling that inheriting from a base class is probably an anti-pattern, and depending on method definitions especially so. I think including shared behaviour via a module leads to better results. In this case, all the base step really gives us is the `with_retries` method (which could be a module on its own). Printing the current step might be more at home in the top-level smoke test, and initialising with `state` would come for free with a `Struct`.

dan-palmer: Yeah, that's spot on. Inheriting here is definitely a smell. Each step is being forced to have behaviour for the sake of a consistent interface when not all of the steps are the same. Most of the email related steps are really just validations - you can't progress, just make an assertion and do something else(another step).

:+1: The `with_retries` should definitely be extracted to a module and the printing should be hoisted up to the level that is controlling the steps.

tatyree: I'll go ahead and implement.  We're ahead of the game and if we don't do it now...

app/mailers/mail_utilities/smoke_test_email_check.rb
tatyree: So it should...

README.md
```diff
@@ -398,3 +398,60 @@ change the UUID.
...
+## Smoke Tests
+
+Runs the happy path to book, confirm and cancel a visit.  If the app
+receives a an email in the format of
+`SMOKE_TEST_EMAIL_LOCAL_PART`@`SMOKE_TEST_EMAIL_DOMAIN`, it will not
+send that email to the prison, but will direct all mail the
+`SMOKE_TEST_EMAIL_HOST`.
+
+To start the smoke tests, set the environment variables and run
+`ruby ./smoke_test/smoke_test.rb`.
```
SteveMarshall: I wonder if this would make sense as a rake task?

SteveMarshall: (To be clear, I don't object too much to this, it just feels odd to be peeking inside the folder structure, if you take my meaning?)

README.md
```diff
@@ -398,3 +398,60 @@ change the UUID.
...
+```
+SMOKE_TEST_EMAIL_LOCAL_PART
+SMOKE_TEST_EMAIL_DOMAIN
+SMOKE_TEST_EMAIL_PASSWORD
+SMOKE_TEST_APP_HOST
+SMOKE_TEST_EMAIL_HOST
+```
+### `SMOKE_TEST_EMAIL_LOCAL_PART`, `SMOKE_TEST_EMAIL_DOMAIN`
+
+These determine the email address that will trigger the smoke tests on
```
SteveMarshall: `the email address that will trigger the smoke tests` … and to which smoke test emails will be sent?

smoke_test/steps/visitors_page.rb
```diff
@@ -0,0 +1,29 @@
...
+    class VisitorsPage < BaseStep
+      PAGE_PATH = '/request'
+
+      def validate!
+        if page.current_path != PAGE_PATH
+          fail "expected #{PAGE_PATH}, got #{page.current_path}"
+        end
+      end
+
+      # rubocop:disable Metrics/AbcSize
```
SteveMarshall: As with my other comment, is it worth re-enabling this after the method it affects? (Though I realise there's not much after it just yet…)

README.md
tatyree: I don't think that would get us anything other than complexity.  It is almost a standalone app and is run from a four or five line shell script on Jenkins.

README.md
tatyree: Good point.  I'll list them.

README.md
SteveMarshall: :+1:

SteveMarshall:
Smoke tests are GO! :rocket:

PR #40, opened 2015-11-27
Additional visitors

threedaymonk:
Allow someone to request additional visitors to accompany them, and allow prisons to indicate who is banned or not on the list.

What works:

* Can select up to five additional visitors.
* Additional visitors are validated.
* There is a JavaScript-enhanced drop-down selector to choose how many additional visitors are required: this hides/shows the additional visitors.
* Prisons can say which visitors are banned/not listed, and this is communicated in the rejection email.

What doesn't work exactly the same as before:

* You can't delete an individual additional visitor, just shrink/grow the list.


app/helpers/date_helper.rb
```diff
@@ -3,8 +3,8 @@ def format_date_of_birth(date)
     I18n.l(date.to_date, format: :date_of_birth)
   end
 
-  def format_date_of_visit(date)
-    I18n.l(date.to_date, format: :date_of_visit)
+  def format_near_date(date)
```
tatyree: I find 'near date' somewhat confusing in contrast to 'date of visit'.  It refers to a date in the near future?

threedaymonk: Near future (or perhaps recent past). It's used for the renewal and expiry of visiting allowance as well as visits.

threedaymonk: Perhaps it could just be `format_date`, though, as it's used for everything that isn't a date of birth. I'm open to ideas.

app/models/booking_response.rb
```diff
@@ -40,7 +43,34 @@ def no_allowance?
...
+
+  def validate_checked_visitors
+    if selection == 'visitor_not_on_list' && unlisted_visitor_ids.empty?
+      errors.add :selection, :no_unlisted_visitors_selected
+    elsif selection == 'visitor_banned' && banned_visitor_ids.empty?
+      errors.add :selection, :no_banned_visitors_selected
+    end
+  end
+  validate :validate_checked_visitors
 end
```
tatyree: `selection` should be cast as a string.

threedaymonk: You mean in case somebody uses a symbol? Sounds reasonable.

app/models/booking_response.rb
```diff
@@ -40,7 +43,34 @@ def no_allowance?
...
+  end
+
+  def banned_visitors
+    visitors.select { |v| banned_visitor_ids.include?(v.id) }
+  end
+
+private
+
+  def validate_checked_visitors
+    if selection == 'visitor_not_on_list' && unlisted_visitor_ids.empty?
```
threedaymonk: This should use the `visitor_not_on_list?` method defined above.

app/models/booking_response.rb
threedaymonk: Good news: it turns out that once Virtus is told to use the correct attribute type, it will convert symbol inputs to strings. I've added a spec to prove this.

app/models/booking_response.rb
```diff
@@ -40,7 +43,34 @@ def no_allowance?
...
+      errors.add :selection, :no_unlisted_visitors_selected
```
tatyree: Could these errors be more declarative?  `:no_unlisted_visitors_have_been_selected` or something similar...

app/models/booking_response.rb
```diff
@@ -40,7 +43,34 @@ def no_allowance?
...
+    elsif selection == 'visitor_banned' && banned_visitor_ids.empty?
+      errors.add :selection, :no_banned_visitors_selected
+    end
+  end
+  validate :validate_checked_visitors
```
tatyree: Should this be a private method?

threedaymonk: Why would this need to be public? It's never used externally.

app/models/visitor.rb
```diff
@@ -0,0 +1,18 @@
+class Visitor < ActiveRecord::Base
+  include Person
+
+  belongs_to :visit
+  validates :visit, :rank, presence: true
```
tatyree: `rank` seems stilted in this context.  I don't have anything better to offer (`position` seems equally bad), so just flagging that I don't like it, really. :anguished:

threedaymonk: Me neither. I thought `position` would be misleading, but don't have any good ideas.

app/models/visitors_step.rb
```diff
@@ -1,18 +1,66 @@
...
+    num_needed = Prison::MAX_VISITORS - existing.count
+    backfill = num_needed.times.map { Visitor.new }
+    existing + backfill
+  end
+
+  def visitors_attributes=(params)
+    # params is of the form
+    # {"0" => {"foo" => "bar"}, "1" => {"foo" => "baz"}}
+    # so we sort by key and take the values. We prune empty parameters
+    # from not-filled-in additional visitors. Virtus will handle the rest.
```
tatyree: I'm afraid you lost me here.

app/services/booking_responder.rb
```diff
@@ -32,6 +32,7 @@ def reject!
     visit.reject!
     rejection = Rejection.new(visit: visit, reason: booking_response.selection)
     copy_no_allowance_parameters rejection if booking_response.no_allowance?
+    mark_visitors
```
tatyree: Maybe `mark_disallowed_visitors`?

db/migrate/20151127120844_add_prisoner_model.rb
```diff
@@ -0,0 +1,33 @@
+class AddPrisonerModel < ActiveRecord::Migration
+  def up
+    create_table :prisoners, id: :uuid, default: 'uuid_generate_v4()' do |t|
```
tatyree: As the PG uuid module is activated, so you don't need the default: 

`create_table :prisoners, id: :uuid do |t|`

app/services/booking_responder.rb
threedaymonk: Nice.

app/models/visitors_step.rb
```diff
@@ -1,18 +1,65 @@
...
+    )
+
+    # We always want at least one visitor. Leaving the rest blank is fine, but
+    # the first one must both exist and be valid.
+    self.visitors = pruned.empty? ? [{}] : pruned.take(Prison::MAX_VISITORS)
+  end
+
+  def valid?(*)
+    # This must be eager because we want to show errors on all objects.
+    results = [super]
```
dan-palmer: Could we use `reduce` here?

```ruby
 visitors.reduce([super]) { |acc, visitor| acc << visitor.valid? }.all?
```

threedaymonk: Yes, that's a bit terser, isn't it? (I'm going to use `inject` instead of `reduce` for consistency, but that's not significant.)

app/models/booking_response.rb
```diff
@@ -5,7 +5,7 @@ class BookingResponse
 
   attribute :visit
 
-  attribute :selection, Integer
+  attribute :selection, String
```
dan-palmer: My obsession with StringInquirer continues.. You <i>could</i> do something like this..

```ruby
class StringWithInquiry < Virtus::Attribute
  def coerce(value)
    value.to_s.inquiry
  end
end

class BookedResponse
  attribute :selection, StringWithInquiry
  delegate :no_allowance?, :visitor_not_on_list?, :visitor_banned?, to: :selection
end
```

This might be useful/reusable for other places in the app that use `inquiry` on strings values.

README.md
```diff
@@ -76,16 +76,25 @@ slots are determined by the recurring weekly pattern.
 
 #### `Visit`
 
-This is the main table in the application, and contains all the essential data
-for a visit: the prison, prisoner details, and primary visitor's details and
-contact information.
+This is the main table in the application, and contains the essential data for
```
dan-palmer: I like that you guys have notes like this, just hope that they are kept in sync.

dan-palmer:
Such nice codez :+1: :ship:

app/models/booking_response.rb
threedaymonk: Oh, that's really nice. I was looking at all my inquiry methods thinking that it ought to be possible to DRY them up.

threedaymonk: I did it even more simply:

```ruby
  delegate :inquiry, to: :selection, prefix: true
  delegate :no_allowance?, :visitor_not_on_list?, :visitor_banned?,
    to: :selection_inquiry
```

dan-palmer: Nice!

app/models/visitor.rb
SteveMarshall: Perhaps `index`? Though that seems somewhat… loaded. Another I was wondering about was `order`? I'm not *too* averse to `rank`, mind you, but I agree that it's somewhat stilted.

app/models/booking_response.rb
tatyree: Oh, spiffy!

app/models/visitor.rb
threedaymonk: I wanted to avoid `order` because it's already a keyword in SQL and a special key in Rails.

PR #41, opened 2015-12-03
Never attempt an MX check in test/dev

threedaymonk:
Checking DNS in tests is slow and failure-prone, and in development it's just plain annoying.

### DIFF
--- a/app/services/email_checker.rb
+++ b/app/services/email_checker.rb
@@ -73,10 +73,6 @@ private
   end
 
   def mx_records?
-    Resolv::DNS.new.getresource(domain, Resolv::DNS::Resource::IN::MX)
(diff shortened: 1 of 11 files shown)


app/services/email_checker.rb
```diff
@@ -73,10 +73,6 @@ def well_formed_address?
   end
 
   def mx_records?
-    Resolv::DNS.new.getresource(domain, Resolv::DNS::Resource::IN::MX)
-  rescue Resolv::ResolvError
-    false
-  rescue Resolv::ResolvTimeout
-    true
+    Rails.configuration.mx_checker.records?(domain)
```
tatyree: :+1: I really like this approach for isolating it to a particular environment.

PR #42, opened 2015-12-03
Standarized docker image creation

lukaszraczylo:


Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM ministryofjustice/ruby:2-webapp-onbuild
+
+RUN touch /etc/inittab
+
+ENV UNICORN_PORT 3000
+EXPOSE $UNICORN_PORT
+
+COPY run.sh /run.sh
+RUN chmod +x /run.sh
+CMD /run.sh
```
threedaymonk: Please finish your files with a new line.

Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM ministryofjustice/ruby:2-webapp-onbuild
+
+RUN touch /etc/inittab
+
+ENV UNICORN_PORT 3000
+EXPOSE $UNICORN_PORT
+
+COPY run.sh /run.sh
```
SteveMarshall: Is it standard behaviour to put `run.sh` in `/`? It feels more natural to me to have it in a location equivalent to its location on the host (e.g. `/usr/src/app/run.sh`)?

run.sh
```diff
@@ -0,0 +1,28 @@
...
+    echo "running migrate"
+    bundle exec rake db:migrate
+    ;;
+seed)
+    echo "running seed"
+    bundle exec rake db:migrate
+    bundle exec rake db:seed
+    ;;
+vagrant)
+    echo "running seed"
```
SteveMarshall: Should this be `running vagrant`?

run.sh
```diff
@@ -0,0 +1,28 @@
...
+    bundle exec rake db:create
+    bundle exec rake db:seed
+    bundle exec rake db:migrate
+    ;;
+create)
+    echo "running seed"
```
SteveMarshall: Should this be `running create`?

run.sh
```diff
@@ -0,0 +1,28 @@
...
+    bundle exec rake db:migrate
+    bundle exec rake db:seed
+    ;;
+vagrant)
+    echo "running seed"
+    bundle exec rake db:create
+    bundle exec rake db:seed
+    bundle exec rake db:migrate
+    ;;
```
SteveMarshall: Is this supposed to be in a different order (`create`, `seed`, `migrate` vs `create`, `migrate`, `seed`) to the `create` command?

SteveMarshall:
Can you remove the trailing `.` from your commit message, @lukaszraczylo? Commit messages should follow the [alphagov git styleguide](https://github.com/alphagov/styleguides/blob/master/git.md).

run.sh
```diff
@@ -0,0 +1,28 @@
...
+create)
+    echo "running seed"
+    bundle exec rake db:create
+    bundle exec rake db:migrate
+    bundle exec rake db:seed
+    ;;
+esac
+bundle exec rake import:all
+bundle exec rake assets:precompile
+bundle exec rails s -b 0.0.0.0
```
SteveMarshall: Another missing newline here!

run.sh
threedaymonk: I wonder whether it should have a more descriptive name than `vagrant`, which leads me to wonder why it's a separate step from `create`.

Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM ministryofjustice/ruby:2-webapp-onbuild
+
+RUN touch /etc/inittab
+
+ENV UNICORN_PORT 3000
+EXPOSE $UNICORN_PORT
+
+# COPY run.sh /run.sh
+# RUN chmod +x /run.sh
+# CMD /run.sh
```
SteveMarshall: Did you mean to comment this out? If we don't need it, delete it.

Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM ministryofjustice/ruby:2-webapp-onbuild
+
+RUN touch /etc/inittab
```
SteveMarshall: What's this line actually doing?

run.sh
threedaymonk: This might be nicer as the full, explicit form:

```sh
bundle exec rails server --binding 0.0.0.0
```

run.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+cd /usr/src/app
+case ${DOCKER_STATE} in
+create)
+    echo "running seed"
+    bundle exec rake db:create
+    bundle exec rake db:migrate
```
threedaymonk: It's faster and more reliable to do `rake db:schema:load` instead of running all the migrations from the bottom up.

run.sh
```diff
@@ -0,0 +1,22 @@
...
+    bundle exec rake db:seed
```
threedaymonk: Instead of repeatedly instantiating the whole of Rails, we should be able to speed this all up by saying

```sh
bundle exec rake db:create db:schema:load db:seed
```

The same goes for other sections.

run.sh
```diff
@@ -0,0 +1,22 @@
...
+    ;;
+migrate)
+    echo "running migrate"
+    bundle exec rake db:migrate
+    ;;
+esac
+bundle exec rake import:all
+bundle exec rake assets:precompile
```
threedaymonk: We can probably speed things up by combining these rake tasks:

```sh
bundle exec rake import:all assets:precompile
```

PR #43, opened 2015-12-04
Use Rails's built-in Time stubbing

threedaymonk:
We don't need Timecop, so let's not add unnecessary dependencies.

http://api.rubyonrails.org/classes/ActiveSupport/Testing/TimeHelpers.html

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -39,6 +39,5 @@ group :test do
   gem 'poltergeist'
   gem 'simplecov'
   gem 'simplecov-rcov'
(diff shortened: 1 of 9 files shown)


threedaymonk:
@dan-palmer as you suggested!

SteveMarshall:
Oooh, fancy! :tophat:

dan-palmer:
Sweet! :)

SteveMarshall:
Back… to the future!

PR #44, opened 2015-12-04
Enforce visitor age restrictions

threedaymonk:
A visit must have at least one, but not more than three, adult visitors (with adult being defined by an age specific to each prison).

### DIFF
--- a/app/models/prison.rb
+++ b/app/models/prison.rb
@@ -4,6 +4,8 @@ class Prison < ActiveRecord::Base
   MissingUuidMapping = Class.new(StandardError)
 
   MAX_VISITORS = 6
+  MAX_ADULTS = 3
+  MIN_ADULTS = 1
 
   has_many :visits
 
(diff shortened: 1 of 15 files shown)


spec/features/request_a_visit_spec.rb
```diff
@@ -72,5 +72,29 @@
...
+      fill_in 'Day', with: '30'
+      fill_in 'Month', with: '11'
+      fill_in 'Year', with: '2014'
+      fill_in 'Email address', with: visitor_email
+      fill_in 'Phone number', with: '01154960222'
+    end
+
+    click_button 'Continue'
+
+    expect(page).to have_text('There must be at least one adult visitor')
```
tatyree: We reuse these in several places.  They are partially DRY'd out in [`spec/support/helpers/features_helper.rb:12`](https://github.com/ministryofjustice/prison-visits-2/blob/master/spec/support/helpers/features_helper.rb#L12).  At some point, we should refactor so that all these calls use the same helper.

tatyree: Otherwise, LGTM.

threedaymonk: Good point. That did occur to me while I was writing it. I'll do so.

app/models/prison.rb
```diff
@@ -40,8 +42,21 @@ def confirm_by(today = Time.zone.today)
       last
   end
 
+  def validate_visitor_ages_on(target, field, ages)
+    adults, _children = ages.partition { |a| adult?(a) }.map(&:length)
```
dan-palmer: `partition` :+1:

spec/factories/prisons.rb
```diff
@@ -26,6 +26,10 @@
       '01154960%03d' % n
     end
 
+    adult_age do
+      10 + rand(9)
```
dan-palmer: The adult age can be any integer between 10 & 19? Can a 10 year old be classed as an adult?

threedaymonk: I want it to randomly generate children or adults to ensure that all assumptions are clearly stated in tests.

dan-palmer:
LGTM :+1:

app/models/prison.rb
tatyree: I dunno...CAP Theorem anyone? :trollface:

PR #46, opened 2015-12-07
Stop rake db:drop db:create db:migrate breaking

tatyree:
The columns get remaned in `8ac98e0`, which causes this to break if
drop, recreate and migrate.

### DIFF
--- a/db/migrate/20151127095415_rename_contact_details.rb
+++ b/db/migrate/20151127095415_rename_contact_details.rb
@@ -1,6 +1,11 @@
 class RenameContactDetails < ActiveRecord::Migration
   def change
-    rename_column :visits, :visitor_email_address, :contact_email_address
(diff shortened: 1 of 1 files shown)


SteveMarshall:
I'm not qualified enough in the vagaries of migrations to comment on this. @threedaymonk?

threedaymonk:
Ah. I understand. I screwed up: I was overzealous with the search and replace when changing the names, and ended up modifying the migration. No one else noticed either!

I think the solution is to change the migration back, so that it's all consistent.

However, it's really not a good idea to attempt to create the database from scratch by running all the migrations, as `schema.rb` says:

> Note that this schema.rb definition is the authoritative source for your database schema. If you need to create the application database on another system, you should be using db:schema:load, not running all the migrations from scratch. The latter is a flawed and unsustainable approach (the more migrations you'll amass, the slower it'll run and the greater likelihood for issues).

PR #52, opened 2015-12-09
Rationalise I18n YAML

threedaymonk:
There are at least nine separate ways to write a multi-line string in YAML: https://stackoverflow.com/a/21699210

The ones that are most relevant to us can be illustrated thus:

```yaml
a:
  one
  two
b: >
  one
  two
c: >-
  one
  two
d: |
  one
  two
e: |-
  one
  two
```

results in

```ruby
{
  "a" => "one two",
  "b" => "one two\n",
  "c" => "one two",
  "d" => "one\ntwo\n",
  "e" => "one\ntwo",
}
```

In other words, we always want to use c) in preference to b), in order to avoid the unnecessary additional newline. The reason for choosing form c) over a) is that c) is parsed as text, whereas a) is subject to the usual myriad edge cases of YAML - it will blow up if it starts with a percent sign or contains a colon, for example.

We want to use d) or e) whenever it's useful to preserve the layout of complex HTML in a localisation string, but the absence or presence of a final newline in this case is not important, and d) is shorter.

This updates the files in `config/locales` accordingly.

### DIFF
--- a/config/locales/en/email_checker.yml
+++ b/config/locales/en/email_checker.yml
@@ -5,8 +5,8 @@ en:
       domain_dot: "is not a valid address because it ends with a dot or starts with a dot"
       malformed: "is not a valid address"
       no_mx_record: "does not appear to be valid"
-      spam_reported: >
+      spam_reported: >-
         needs to be checked as past messages were marked as spam.
         Check your spam folder too
-      bounced: >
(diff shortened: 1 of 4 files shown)


SteveMarshall:
:boom:

PR #54, opened 2015-12-10
Allow visitor to cancel

threedaymonk:
The visitor can now follow a link in their email (when a request is received or a booking accepted) through to a visit page which shows the current state and, if it's in an appropriate state, allows them to cancel (if the booking is accepted) or withdraw (if the visit is requested but still pending).


README.md
```diff
@@ -80,6 +80,21 @@ This is the main table in the application, and contains the essential data for
...
+     .---------.  withdraw  .---------.  reject  .--------.
+    ( withdrawn ) <------- ( requested ) -----> ( rejected )
+     '---------'            '---------'          '--------'
+                                 |
+                                 | accept
+                                 v
+                              .------.          .--------.
+                             ( booked ) -----> ( canceled )
+                              '------'  cancel  '--------'
+
```
SteveMarshall: Top diagramming! :tophat:

app/controllers/application_controller.rb
```diff
@@ -2,4 +2,13 @@ class ApplicationController < ActionController::Base
...
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
+
+private
+
+  def error_unprocessable_entity
+    render \
+      file: Rails.root.join('public', '422.html'),
+      layout: false,
+      status: 422
```
SteveMarshall: This whole method feels a bit… odd… to me, and I can't quite place why. Perhaps it should be handled by raising an exception? I suspect my sense is a bit off here, though. :confused:

Separately, it could be good to replace the `422` with `:unprocessable_entity`? Rails will handle that equivalently to the status code, and it seems more descriptive that way?

threedaymonk: My first thought was the same, that this might be better as an exception, but it wasn't obvious how to do it. Your comment has prompted me to look in more detail, and I think I've found a way to do it.

config/application.rb
```diff
@@ -29,5 +29,9 @@ class Application < Rails::Application
     config.time_zone = 'London'
 
     config.active_record.raise_in_transactional_callbacks = true
+
+    config.action_dispatch.rescue_responses.merge!(
+      'StateMachines::InvalidTransition' => :unprocessable_entity
```
threedaymonk: This is the magic line: it causes `InvalidTransition` exceptions to be turned into 422 errors. (Except in development mode, *unless* you disable the `config.consider_all_requests_local = true` option in `development.rb`). I think this will always be the right thing to do: it's the same as `ActiveRecord::RecordInvalid`, for example.

SteveMarshall: :+1:

SteveMarshall:
LGTM! @tatyree?

app/controllers/booking_requests_controller.rb
```diff
@@ -1,4 +1,4 @@
-class StepsController < ApplicationController
+class BookingRequestsController < ApplicationController
```
tatyree: :+1:

tatyree:
LGTM, too, @SteveMarshall

PR #56, opened 2015-12-10
Send tickets to Zendesk from in-app feedback form

threedaymonk:
Unlike the old app, we no longer bother with the duplication of the email and Zendesk routes. (It would send all feedback via email, but only feedback with a supplied email address went to Zendesk.) Instead, we now send all feedback to Zendesk, using a placeholder address (currently `feedback@` etc.) when someone has chosen not to supply one.

### DIFF
--- a/app/jobs/zendesk_tickets_job.rb
+++ b/app/jobs/zendesk_tickets_job.rb
@@ -1,0 +1,28 @@
+class ZendeskTicketsJob < ActiveJob::Base
+  queue_as :zendesk
+
+  # Custom ticket fields configured in the MOJ Digital Zendesk account
+  URL_FIELD = '23730083'
+  SERVICE_FIELD = '23757677'
+  BROWSER_FIELD = '23791776'
+
+  def perform(feedback)
+    ZendeskAPI::Ticket.create!(
+      Rails.configuration.zendesk_client,
+      description: feedback.body,
+      requester: {
+        email: feedback.email_address,
+        name: 'Unknown'
+      },
+      custom_fields: custom_fields(feedback)
+    )
+  end
+
+  def custom_fields(feedback)
+    [
+      { id: URL_FIELD, value: feedback.referrer },
+      { id: SERVICE_FIELD, value: 'prison_visits' },
+      { id: BROWSER_FIELD, value: feedback.user_agent }
+    ]
+  end
+end
--- a/app/models/feedback_submission.rb
+++ b/app/models/feedback_submission.rb
@@ -3,7 +3,7 @@ class FeedbackSubmission < ActiveRecord::Base
   validate :validate_email
 
   def email_address
-    super.present? ? super : 'feedback@email.prisonvisits.service.gov.uk'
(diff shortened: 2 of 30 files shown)


app/jobs/zendesk_tickets_job.rb
```diff
@@ -0,0 +1,23 @@
...
+      },
+      custom_fields: custom_fields(feedback)
+    )
+  end
+
+  def custom_fields(feedback)
+    [
+      { id: '23730083', value: feedback.referrer },
+      { id: '23757677', value: 'prison_visits' },
+      { id: '23791776', value: feedback.user_agent }
```
SteveMarshall: What do these magic numbers mean?

app/models/feedback_submission.rb
```diff
@@ -0,0 +1,16 @@
...
+  def email_address
+    super.present? ? super : Rails.configuration.address_book.feedback
+  end
+
+private
+
+  def validate_email
+    return unless email_address.present?
+    checker = EmailChecker.new(email_address)
+    errors.add :email_address, checker.message unless checker.valid?
```
SteveMarshall: This might be clearer if `checker` was `email` or `email_checker` or something?

app/jobs/zendesk_tickets_job.rb
threedaymonk: I'm not sure! They're like this in the old app, but there's no explanation (except our old friend 'Initial commit (open sourcing)').

SteveMarshall:
781de07 is :star::star::star::star::star:!

app/jobs/zendesk_tickets_job.rb
SteveMarshall: According to the [Zendesk API docs on creating tickets](https://developer.zendesk.com/rest_api/docs/core/tickets#setting-custom-field-values), they relate to our custom ticket fields.

After poking around in the admin UI, it would appear that they are:

- `23730083` = `URL` text field
- `23757677` = `Service` dropdown (with our service name being `prison_visits`)
- `23791776` = `Browser` text field

app/services/address_book.rb
```diff
@@ -0,0 +1,13 @@
+class AddressBook
+  def initialize(domain)
+    @domain = domain
+  end
+
+  def feedback
+    "feedback@#{@domain}"
```
tatyree: This and 'no-reply' need translations, don't they?

config/locales/en/mailers.yml
```diff
@@ -58,8 +58,7 @@ en:
       add_to_address_book_title: Add this email to your address book
       add_to_address_book_body: >-
```
tatyree: I was momentarily confused by the naming of these two owing to the existence of `services/address_book.rb`.

app/services/address_book.rb
threedaymonk: `feedback` doesn't, because it's just a stand-in for sending things to Zendesk, and the public won't ever see it. `no-reply` is an interesting one, though!

PR #57, opened 2015-12-11
Destroy dependent visitors and rejection objects

tatyree:
The foreign key constraint prevents deletion of a Visit object in the console without
first deleting all its associated Visitors and the Rejection.


threedaymonk:
Typo in commit: "Destory"

threedaymonk:
Should we do the same for the other `has_many`s?

tatyree:
No there ins't :)

And yes, we should...and done.

PR #58, opened 2015-12-14
Correct misnamed attribute from hostname to host

tatyree:
I discovered this while running the app in production mode to check that
the smoke tests worked.


threedaymonk:
Oops, you're quite right.

SteveMarshall:
I wonder if there's a way we can catch this general class of error that isn't 'deploy and hope smoke tests catch it'?

SteveMarshall:
(Good spot though, @tatyree!)

tatyree:
Well, I caught it by running the smoke tests locally, @SteveMarshall...

PR #59, opened 2015-12-14
Add links out to Prison Finder

threedaymonk:
At various points, the app directs visitors to Prison Finder to find more information about prisons. Up till now, these links have been placeholders.

This adds in the correct links, either to the top of Prison Finder, or to the prison-specific page if we know which prison we're talking about.

As in the previous app, the Prison Finder slug is generated by using Rails's `parameterize` method on the name of the prison; where this does not produce the desired result, it can be overridden via a `finder_slug` key in the seed file. The slug is generated or updated at seed time.


app/models/prisoner_step.rb
```diff
@@ -16,6 +16,6 @@ class PrisonerStep
   delegate :name, to: :prison, prefix: true
 
   def prison
-    Prison.find(prison_id)
+    prison_id.present? ? Prison.find(prison_id) : nil
```
tatyree: How about just `Prison.find_by(id: prison_id)`?

db/seeds.rb
```diff
@@ -1,40 +1,3 @@
...
-      phone_no: hash.fetch('phone', nil),
-      adult_age: hash.fetch('adult_age', 18),
-      slot_details: {
-        'recurring' => hash.fetch('slots', {}),
-        'anomalous' => hash.fetch('slot_anomalies', {}),
-        'unbookable' => hash.fetch('unbookable', [])
-      }
-    )
-  end
+  PrisonSeeder.seed! Rails.root.join('db', 'seeds')
```
tatyree: :+1:

app/models/prisoner_step.rb
threedaymonk: Well spotted!

app/services/link_directory.rb
```diff
@@ -0,0 +1,10 @@
+class LinkDirectory
+  def initialize(prison_finder:)
+    @prison_finder_base = prison_finder
+  end
+
+  def prison_finder(prison = nil)
+    slug = prison ? prison.finder_slug : nil
+    [@prison_finder_base, slug].compact.join('/')
```
SteveMarshall: Is this generating a (stringy) URI? If so, might be better to use `URI.join`?

SteveMarshall: (Although I realise that will return a `URI`, not a string, so perhaps we need to do a `to_s` or similar.)

threedaymonk: I suspect it'll turn into an unreadable mess:

```ruby
URI.join(*[@prison_finder_base, slug].compact).to_s
```

PR #60, opened 2015-12-15
Add timestamps to every DB table

threedaymonk:
It's always useful to know when a row was created or updated, but we forgot to add these columns.

Existing records are backfilled with the current date and time. In practice, as we aren't live yet, this doesn't matter, but it allows for painless migration in development.


SteveMarshall:
:+1: I wonder if it's worth adding some sort of audit trail for updates, too? Probably something for post-launch, though…

PR #61, opened 2015-12-15
Implement asynchronous queues

threedaymonk:
We don't want the unavailability of the SMTP server or Zendesk to inconvenience the user or cause a request to be lost, so we use Redis/Sidekiq to process these asynchronously.

This is a straight port from the old application.


tatyree:
LGTM :+1:

tatyree:
@SteveMarshall, any thoughts? If not, I'll merge.

SteveMarshall:
@tatyree My only thought was “wait, I didn't do anything with Sidekiq today…”, then realised @threedaymonk had cherry-picked these commits from the old app!

PR #62, opened 2015-12-15
Add /healthcheck.json endpoint

threedaymonk:
This exposes a JSON document at `/healthcheck.json` that reports the state of queues, database connectivity, SMTP server, and overall health of the application.

### DIFF
--- a/app/services/healthcheck.rb
+++ b/app/services/healthcheck.rb
@@ -5,7 +5,8 @@ class Healthcheck
     @components = {
       database: DatabaseCheck.new('Postgres database'),
       mailers: QueueCheck.new('Email queue', queue_name: 'mailers'),
-      zendesk: QueueCheck.new('Zendesk queue', queue_name: 'zendesk')
+      zendesk: QueueCheck.new('Zendesk queue', queue_name: 'zendesk'),
+      smtp: SmtpCheck.new('SMTP server', smtp_settings: smtp_settings)
     }
   end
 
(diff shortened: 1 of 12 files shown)


SteveMarshall:
Might be nice to add a check for connecting to SendGrid and reporting that in the health check, too, to give us a more direct indicator of if the SMTP connection is down.

SteveMarshall:
I wonder if it's worth adding more explicit error messages for some of the compound checks here? For example, making it explicit that we couldn't connect to the queue in `queue_info`.

app/services/healthcheck.rb
```diff
@@ -0,0 +1,66 @@
...
+  def queue_info(name, description)
+    q = Sidekiq::Queue.new(name)
+    {
+      description: description,
+      ok: fresh?(q),
+      oldest: oldest(q),
+      count: q.count
+    }
+  rescue StandardError
+    { description: description, ok: false, oldest: nil, count: 0 }
```
SteveMarshall: Can we offer a more explicit “do not trust these values” value for  `count`? (Perhaps `nil` as in `oldest`, or maybe that should have a more explicit one, too…)

threedaymonk:
@SteveMarshall It now reports any errors it catches, and I've added an SMTP connection check.

app/services/healthcheck/check_component.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  private
+
+    def build_report(description)
+      @report =
+        begin
+          yield
+        rescue StandardError => err
+          { error: err.to_s, ok: false }
+        end.merge(description: description)
```
SteveMarshall: This is really neat! :+1:

SteveMarshall:
LGTM; @tatyree?

app/services/healthcheck/check_component.rb
tatyree: I agree .

PR #63, opened 2015-12-15
Notify prison of cancellation

tatyree:
This sends the email notifying a prison of a cancellation or a withdrawal.  It also stops prisons from being able to process unprocessable requests and normalises the spelling of 'cancelled'.


app/views/prison/visits/show.html.erb
```diff
@@ -0,0 +1,34 @@
+<% content_for :header do %>
+  <%= t('.title') %>
+<% end %>
+
+<%= render(partial: 'people_box') %>
+
+<% if @booking_response.unavailable? %>
+  <h2>
+    <%= t('.unavailable', status: @booking_response.processing_state) %>
+  </h2>
```
threedaymonk: I think it would be nice to break these two branches out into separate partials.

config/locales/en/mailers.yml
```diff
@@ -27,6 +27,18 @@ en:
...
         This is a copy of the booking rejection email sent to the visitor
+    cancelled:
+      subject: "%{status}: %{prisoner} on %{date}"
+      salutation: Dear staff,
+      details: >-
+        %{visitor} has cancelled the visit to %{prisoner} (%{prisoner_number) on
+        %{date}
+      reference: "Reference number: %{reference_no}"
+      please_delete: Please delete this visit from NOMIS.
+      closing: |
```
threedaymonk: The `|` is misleading, as the line breaks won't be preserved in the view.

app/views/prison/visits/show.html.erb
```diff
@@ -0,0 +1,34 @@
+<% content_for :header do %>
+  <%= t('.title') %>
+<% end %>
+
+<%= render(partial: 'people_box') %>
+
+<% if @booking_response.unavailable? %>
+  <h2>
+    <%= t('.unavailable', status: @booking_response.processing_state) %>
```
threedaymonk: This should be localised - we can't rely on `processing_state` being English (or even human-friendly). I think the right thing here is to have each condition as a separate translation:

```yaml
withdrawn: "The visitor has withdrawn this request"
rejected: "This visit request has been rejected"
```

etc.

db/migrate/20151215154514_alter_visit_change_spelling_of_canceled_to_cancelled.rb
```diff
@@ -0,0 +1,15 @@
+class AlterVisitChangeSpellingOfCanceledToCancelled < ActiveRecord::Migration
+  def up
+    execute(
+      "UPDATE visits SET processing_state = 'cancelled' " +
+      "WHERE processing_state = 'canceled'"
+    )
```
threedaymonk: You can use `<<-` to avoid unnecessary concatenation (and GitHub will even syntax-highlight it as SQL!):

```ruby
execute <<-SQL
  UPDATE visits SET processing_state = 'cancelled'
  WHERE processing_state = 'canceled'
SQL
```

db/schema.rb
```diff
@@ -90,10 +90,10 @@
     t.string   "processing_state",        default: "requested", null: false
     t.datetime "created_at",                                    null: false
     t.datetime "updated_at",                                    null: false
-    t.string   "reference_no"
-    t.boolean  "closed"
     t.boolean  "override_delivery_error", default: false
     t.string   "delivery_error_type"
+    t.string   "reference_no"
+    t.boolean  "closed"
```
threedaymonk: This change is unnecessary, and just reflects your migration order.

spec/mailers/prison_mailer/cancelled_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    travel_to(Date.new(2015, 6, 1)) do
+      example.run
+    end
+  end
+
+  context 'cancelled visit' do
+    let(:visit) { create(:cancelled_visit, prisoner: prisoner) }
+    include_examples 'template checks'
+
+    it 'sends an email notfiyting the prison of the cancellation' do
```
threedaymonk: Typo in "notifying".

app/models/visit.rb
```diff
@@ -76,4 +76,8 @@ def banned_visitors
   def unlisted_visitors
     visitors.unlisted
   end
+
+  def unavailable?
```
threedaymonk: I don't think `unavailable?` is the right word here: it implies something different about the visit. Perhaps `awaiting_processing?` might be closer.

app/mailers/prison_mailer.rb
```diff
@@ -34,4 +34,18 @@ def rejected(visit)
            prisoner: visit.prisoner_full_name
          )
   end
+
+  def cancelled(visit)
+    @visit = visit
+
+    headers('X-Priority' => '1 (Highest)', 'X-MSMail-Priority' => 'High')
```
threedaymonk: This is a bit fiddly. Could we extract it to a method that explains the intention?

config/locales/en/mailers.yml
tatyree: Ah, yeah, it was a semi-text message in the old app and I was considering just making it `text.erb`.

db/schema.rb
tatyree: I know.  It's something that pesters me all the time and I'm afraid I let it slip past this time.

app/models/visit.rb
tatyree: `awaiting_processing?` won't work either as it includes `booked` and `rejected`...I'll think about it and see if I can come up with something more suitable.

app/mailers/prison_mailer.rb
tatyree: @threedaymonk you mean the line setting the headers or the whole method?

app/views/prison/visits/show.html.erb
tatyree: Do you mean into `withdrawn` and `cancelled` partials or into a partial for requests that cannot be processed and those that can?

app/models/visit.rb
tatyree: giphy

app/views/prison/visits/show.html.erb
threedaymonk: I mean the two branches of the `if`/`else`, so unavailable vs. available, I guess.

app/models/visit.rb
threedaymonk: In that case, I'm more confused about what this method is for! As I understand it, the scenarios are:

1. The booking is in the initial requested state, and a prison can process it
2. The booking has been processed (accepted, rejected) or withdrawn or cancelled, and they can't do anything except look at it.

app/mailers/prison_mailer.rb
threedaymonk: The line setting the headers.

db/schema.rb
SteveMarshall: That seems like a… not-good position to be in. Is there any way to resolve those sorts of issues so they don't keep pestering you?

spec/models/visit_spec.rb
```diff
@@ -28,6 +28,38 @@
       subject.reject!
       expect(subject).to be_rejected
     end
+
+    it 'is withdrawn after cancellation if unaccpeted' do
```
SteveMarshall: `unaccpeted

spec/models/visit_spec.rb
```diff
@@ -28,6 +28,38 @@
...
+      subject.cancel!
+      expect(subject).to be_withdrawn
+    end
+
+    it 'is cancelled after cancellation if accpeted' do
```
SteveMarshall: `accpeted`

db/schema.rb
threedaymonk: Someone needs to write a better schema generator for ActiveRecord that doesn't muddle the order. If it didn't bother trying to line up the columns, the diffs would be less volatile, too. Perhaps I'll do that …

spec/models/visit_spec.rb
threedaymonk: Hey, at least he's consistent :stuck_out_tongue:

spec/models/visit_spec.rb
```diff
@@ -28,6 +28,38 @@
...
+      expect(subject).to be_withdrawn
+    end
+
+    it 'is cancelled after cancellation if accepted' do
+      subject.accept!
+      subject.cancel!
+      expect(subject).to be_cancelled
+    end
+
+    it 'exposes booked as not processable' do
```
SteveMarshall: These names feel a bit passive-voice-y to me: `it 'exposes booked as not processable'` is a bit… odd? Maybe `it 'is not processable after booking'` and so on?

app/helpers/prison_admin_helper.rb
```diff
@@ -0,0 +1,5 @@
+module PrisonAdminHelper
+  def reason_reqeuest_cannot_be_processed(booking_response)
```
threedaymonk: typo; should be 'request'

app/views/prison/visits/show.html.erb
```diff
@@ -0,0 +1,33 @@
...
+    form_for(
+      @booking_response,
+      url: prison_visit_path(@booking_response),
+      method: :put
+    ) do |f|
+  %>
+<%= render(partial: 'visit_date_section', locals: { f: f }) %>
+<%= render(partial: 'visiting_allowance_section', locals: { f: f }) %>
+<%= render(partial: 'issue_with_prisoner_section', locals: { f: f }) %>
+<%= render(partial: 'issue_with_visitors_section', locals: { f: f }) %>
```
threedaymonk: The indentation is off. Luckily, we're not using Haml, so we don't get a totally different HTML document, but it would be nice to indent them properly nonetheless!

config/locales/en/views.yml
```diff
@@ -169,6 +169,10 @@ en:
             </li>
           </ul>
         submit: Send email
+        booked: This request has already been accepted
+        cancelled: The visitor has cancelled this request
```
threedaymonk: It's an accepted booking that has been cancelled.

app/models/booking_response.rb
```diff
@@ -31,6 +31,7 @@ class BookingResponse
     :prisoner_full_name, :prisoner_number, :prisoner_date_of_birth,
     :contact_email_address, :contact_phone_no,
     :visitors,
+    :processable?, :human_processing_state_name,
```
threedaymonk: where does `human_processing_state_name` come from in `Visit`? Anyway, I think this is misnamed: it's not human, but internal, and we only convert to human at the i18n layer.

app/helpers/prison_admin_helper.rb
```diff
@@ -0,0 +1,5 @@
+module PrisonAdminHelper
+  def reason_request_cannot_be_processed(booking_response)
+    content_tag(:h2, t(".#{booking_response.processing_state_name}"))
```
threedaymonk: This method seems out of place to me - especially moving the H2 out of the template.

app/views/prison/visits/show.html.erb
```diff
@@ -0,0 +1,31 @@
+<% content_for :header do %>
+  <%= t('.title') %>
+<% end %>
+
+<%= render(partial: 'people_box') %>
+
+<% unless @booking_response.processable? %>
+  <h2><%= t(".#{@booking_response.processing_state_name}") %></h2>
+<% else %>
```
threedaymonk: You're going to hate me for this, but one more thing: use `if`/`else`, not `unless`/`else`.

PR #64, opened 2015-12-16
Dockerize

lukaszraczylo:


run.sh
```diff
@@ -14,5 +14,5 @@ migrate)
     bundle exec rake db:schema:load
     ;;
 esac
-bundle exec import:all assets:precompile
+RAILS_ENV=production bundle exec import:all assets:precompile
```
threedaymonk: This should be set for every command.

threedaymonk:
I know GitHub has a button to merge master into your branch, but please don't use it: rebase *onto* master instead.

lukaszraczylo:
Thanks, I didn't actually knew there's such feature.

run.sh
lukaszraczylo:
```
[Dec-14 17:18] PB: @Lukasz I ran rake assets:precompile with the environment set to production, ran the server (in production env with it set to serve static files), and all the assets worked as expected
```

There was no mention of "every command" so please be more specific next time.

threedaymonk: I wasn't telling you what to do! I was recounting what I did, to reproduce a production environment as closely as I could.

If this line does fix asset generation, then it might be that the environment is not set properly. `RAILS_ENV` **must** be set to `production` on production (and pre-production) servers.

SteveMarshall:
@lukaszraczylo That's not quite what @threedaymonk meant. His point was that you've got a merge commit in this PR (0c9e486) which shouldn't exist. Can you run a `git rebase -i master` (after pulling the  latest master from GitHub) and force-push the result to this branch?

That way, we end up with no unnecessary merge commits in the repo's history. It would also be nice if you could fix the title of this PR to that of 5602256, and add some more description to that commit (probably when rebasing) to describe why this change fixes the missing assets.

PR #65, opened 2015-12-16
Inhibit caching in browsers and proxies

threedaymonk:
We don't want intermediates to cache booking requests, and we don't want other users of the same computer to be able to see what was previously booked.

See https://stackoverflow.com/questions/49547 for more detail about the headers used.

Tested in the browser: the headers are set, and attempting to go back re-requests the page as expected.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -3,6 +3,8 @@ class ApplicationController < ActionController::Base
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
 
+  before_action :do_not_cache
+
   helper LinksHelper
 
 private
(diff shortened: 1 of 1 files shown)


app/controllers/application_controller.rb
```diff
@@ -3,6 +3,8 @@ class ApplicationController < ActionController::Base
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
 
+  before_action :do_not_cache
```
SteveMarshall: Do we actually want to apply this application-wide? We probably do as it stands, but if we have to host the staff info pages, we should probably remember to remove this.

Also, does this apply to assets? If not, then we're fine, but we should probably allow those to be cached as heavily as possible…

threedaymonk: It doesn't apply to assets: those aren't served by the controllers, so even if we do serve them from the app (which I'm not sure is the case in production) they get the correct headers.

PR #66, opened 2015-12-16
Restrict booking admin pages to prisons

threedaymonk:
This accepts the same comma-separated environment variable as the old application, but adds the option to use CIDR ranges. This code is adapted from a similar facility in People Finder.

Denying access is implemented in the router: agents outside the permitted ranges are rebuffed with a 404 response.


tatyree:
LGTM.  @SteveMarshall?

PR #67, opened 2015-12-16
Let visitors edit their request before confirming

threedaymonk:
This adds a `review_step` parameter that will override the default behaviour of the `StepsProcessor` and be shown instead of the first incomplete step.

In order to distinguish incomplete from not-yet-filled steps (necessary to avoid showing validation errors on the first try), the hidden fields for later steps are only shown when reviewing.

The option selectors for visit slots are returned to using Rails's `FormBuilder` methods, reverting the change made in a2ca599, as this ignored existing selections. As the correct CSS class is applied, the
JavaScript enhanced selector continues to work.

To implement the edit, there is a form with hidden fields for each step on the confirmation page. As there are three editable steps, this means that the fields are replicated three times. This is slightly
unfortunate, but works well.

### DIFF
--- a/app/views/booking_requests/confirmation_step.html.erb
+++ b/app/views/booking_requests/confirmation_step.html.erb
@@ -33,9 +33,16 @@
     <strong><%= prisoner_step.prison_name %></strong>
   </p>
 
-  <p class="edit-link">
-    <%= link_to(t('.change_prisoner_details'), '/TODO') %>
-  </p>
+  <%= form_for(confirmation_step, url: booking_requests_path) do |f| %>
+    <%= render(partial: 'hidden_prisoner_step') %>
+    <%= render(partial: 'hidden_visitors_step') %>
+    <%= render(partial: 'hidden_slots_step') %>
+    <%= hidden_field_tag('review_step', 'prisoner_step') %>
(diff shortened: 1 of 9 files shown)


app/views/booking_requests/confirmation_step.html.erb
```diff
@@ -33,9 +33,16 @@
...
-  </p>
+  <%= form_for(confirmation_step, url: booking_requests_path) do |f| %>
+    <%= render(partial: 'hidden_prisoner_step') %>
+    <%= render(partial: 'hidden_visitors_step') %>
+    <%= render(partial: 'hidden_slots_step') %>
+    <%= hidden_field_tag('review_step', 'prisoner_step') %>
+
+    <p class="edit-link">
+      <%= f.submit(t('.change_prisoner_details'), class: 'button') %>
+    </p>
```
SteveMarshall: This is so much better, passing the state around this way rather than implicitly using cookies and so on! :boom:

SteveMarshall:
LGTM! @tatyree?

(Also, needs rebasing @threedaymonk.)

threedaymonk:
Rebased!

PR #69, opened 2015-12-17
Legal pages

threedaymonk:
This adds the terms and conditions and cookie pages, ported from the previous app.

Dumping a load of HTML into the locale file doesn't seem ideal, but breaking down each paragraph of legalese would be pointlessly onerous.


SteveMarshall:
LGTM! Rebase?

PR #75, opened 2015-12-18
Compile rails assets in production mode

lukaszraczylo:
Change made as few assets were missing while being compiled without setting appropriate RAILS_ENV.

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+export RAILS_ENV=production
 cd /usr/src/app
 case ${DOCKER_STATE} in
 create)


run.sh
```diff
@@ -14,5 +14,5 @@ migrate)
     bundle exec rake db:schema:load
     ;;
 esac
-bundle exec import:all assets:precompile
+RAILS_ENV=production bundle exec import:all assets:precompile
```
threedaymonk: As I said previously, I don't think it makes sense to change just this line. `RAILS_ENV` should be set to `production` for every `bundle exec …` command, so it should be set once somewhere that covers all these command lines and the servers.

lukaszraczylo: https://github.com/ministryofjustice/prison-visits-2/blob/master/config/database.yml#L26
https://github.com/ministryofjustice/prison-visits-2/blob/master/config/database.yml#L81

Rails environments were created for development, testing and production. If we will enforce RAILS_ENV production during every container run ( even during jenkins build / tests ) there might be a problem with at least connectivity. It counts for email servers details as well.

threedaymonk: We don't need `test` or `development` environments: they're only used locally and by our continuous integration service. *Everything* on production and pre-production – scripts, queue consumers, and web servers – should be run with `RAILS_ENV` set to `production`.

threedaymonk: With that said, if you do want to run tests in Jenkins, then yes, you'll need a `test` environment. However, in that case, you don't need to build assets at all. So what I don't understand is why you'd ever need to run the asset compilation in the `production` `RAILS_ENV` – if you're compiling assets, you should already be in `production`.

lukaszraczylo: That changes everything then :) Added fixup commit.

threedaymonk:
If you rebase this onto master to get rid of the fixups, I'll merge it.

PR #77, opened 2015-12-21
Account for IPv6 when restricting admin access

tatyree:
This is unlikely to be a problem for a while, but I ran into trying to
smoke test the app on my development machine. As it was, the constraint
would raise an exception when it encountered an IPv6 address.  Now, it
treats them exactly the same as IPv4 addresses.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -33,7 +33,6 @@ module PrisonVisits
     config.action_dispatch.rescue_responses.merge!(
       'StateMachines::InvalidTransition' => :unprocessable_entity
     )
-
-    config.prison_ip_ranges = ENV.fetch('PRISON_ESTATE_IPS', '127.0.0.1')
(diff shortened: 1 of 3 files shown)


config/application.rb
```diff
@@ -33,7 +33,18 @@ class Application < Rails::Application
...
+      OpenStruct.new(
+        local_part:
+          Regexp.escape(
+            ENV.fetch('SMOKE_TEST_EMAIL_LOCAL_PART', 'prison-visits-smoke-test')
+          ),
+        domain:
+          Regexp.escape(
+            ENV.fetch('SMOKE_TEST_EMAIL_DOMAIN', 'digital.justice.gov.uk')
+          )
+      )
```
SteveMarshall: I think `config.smoke_test` is for some other PR? :wink:

PR #79, opened 2015-12-22
Require 'rails_helper' consistently

tatyree:
Noticed that some specs here were requiring 'spec_helper' and some
weren't requiring anyting at all.  I changed them all to use
'rails_helper' as per standard practice.


threedaymonk:
Incidentally, there is a reason for not requiring `rails_helper`: if you don't need to load the whole Rails, it's *much* faster, which makes TDD more pleasant. `require 'spec_helper'` is always redundant because that's already specified in `.rspec`.

PR #80, opened 2015-12-22
Timestamp visit state transitions

tatyree:
For ease of calculation of date-based metrics.

I decided to make this its own PR so that any issues with this approach can be flagged early and so that we can get the collection of the metrics in play sooner rather than later.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20151215154514) do
+ActiveRecord::Schema.define(version: 20151222125110) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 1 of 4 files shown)


db/schema.rb
```diff
@@ -90,11 +90,15 @@
     t.string   "processing_state",        default: "requested", null: false
     t.datetime "created_at",                                    null: false
     t.datetime "updated_at",                                    null: false
-    t.string   "reference_no"
-    t.boolean  "closed"
     t.boolean  "override_delivery_error", default: false
     t.string   "delivery_error_type"
+    t.string   "reference_no"
+    t.boolean  "closed"
```
tatyree: I give up.  Thus it was decreed by Caesar that all numbers of reference numbers booleans that spaketh 'closed' shall return to place of their birth there to be counted by the order of the arbitrary men of Judea....

SteveMarshall: Is this another of those “your schema application order is different” thing?

tatyree: Yes.  This time, however, it also has legitimate schema changes.

PR #83, opened 2015-12-23
Metrics timestamps on visits

tatyree:
First steps for review/comment.  At this point I know I don't understand the existing metrics well at all, so I'm trying to work through and translate the basics in very small steps.

### DIFF
--- a/app/models/visit.rb
+++ b/app/models/visit.rb
@@ -29,19 +29,19 @@ class Visit < ActiveRecord::Base
 
   state_machine :processing_state, initial: :requested do
     after_transition on: :accept do |visit|
-      visit.send(:set_details_for_metrics, :accepted_at)
+      visit.send(:details_for_metrics, :accepted_at)
     end
 
     after_transition on: :reject do |visit|
-      visit.send(:set_details_for_metrics, :rejected_at)
+      visit.send(:details_for_metrics, :rejected_at)
     end
 
     after_transition booked: :cancelled do |visit|
-      visit.send(:set_details_for_metrics, :cancelled_at)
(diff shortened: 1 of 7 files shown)


SteveMarshall:
I'm not entirely sold on this: the storing of the per-visit date calculations, while more processing-efficient, risks storing data that will end up needing to be updated/removed if we change how we measure prisons' performance later to something more granular (some prisons already process in <0.5d, for example) or something else entirely. Also, if we make a mistake in our calculations, we'll end up having to recalculate everything instead of just rolling out a new version of the app.

The overall average calculated in `Metrics.end_to_end_processing_time` is one of the calculations we should offer, yeah, but I'm not sure it's the main one.

In the old app, we *tried* to measure processing time for visits in the last two weeks, although the exact algorithm used is kinda meaningless because it broke the year down into 2-ISO-week chunks and anything that crossed a boundary between one chunk and the next was simply ignored. As well as that, it looked at a rolling last-two-weeks window.

I *think* what we're after is mean, median, and percentiles for each prison (and maybe for the entire estate) for visits processed in:

- Evert 2-ISO-week chunks of the year so far
- The current 2-ISO-week chunk (although this could swing wildly)
- The past 2 weeks, as a rolling window
- For this year, last year, and every other year
- For all time

We also have, in the current app, a view that shows when during the week visits are processed/requested to each prison (and over the entire estate, I think), but that's buried and I'm not sure many people even know it's there!

tatyree:
After discussing with @threedaymonk and @kyliemulholland, we decided to remove the denormalised calculations and report in ISO week blocks.

PR #84, opened 2016-01-04
Validate YAML

threedaymonk:
YAML is brittle and easy to get wrong, and even syntactically valid YAML can contain other problems. As in the old application, we want to validate the seed data as part of the test process, so that bad data doesn't get deployed.

This application is already somewhat more robust due to the different way in which it handles the YAML files: they are parsed and imported into the database once at deploy time. However, there are still some parts of the data (around slots) that could pass through this process and cause a problem.

Thus, we now do a few things to protect ourselves:

1.Raise exceptions upon encountering invalid slots or day names.
2. Validate that unbookable dates are valid and do not conflict with anomalous dates.
3. Attempt to import the seed data as part of the test process.


app/models/recurring_slot.rb
```diff
@@ -1,15 +1,48 @@
-RecurringSlot = Struct.new(
-  :begin_hour, :begin_minute, :end_hour, :end_minute
-) do
+TimeRange = Struct.new(:begin_hour, :begin_minute, :end_hour, :end_minute)
+
+class RecurringSlot < TimeRange
```
tatyree: much neater way to do this :+1:

db/seeds/prisons/LCI-leicester.yml
```diff
@@ -33,5 +33,4 @@ slots:
 unbookable:
 - 2014-12-25
 - 2014-12-26
-- 2014-12-31
```
tatyree: Hah! See it caught things already...

threedaymonk: Indeed! Though I think those things had been caught in the old app between the time when we initially imported the data and the time when you added YAML checking over there.

PR #85, opened 2016-01-04
Configure test metadata for CircleCI

threedaymonk:
CircleCI is showing a large pink banner warning that it cannot collect metadata. This attempts to fix that.

See https://circleci.com/docs/test-metadata


threedaymonk:
This is not enough on its own: we also need to configure the RSpec step to run with the custom formatter.

PR #87, opened 2016-01-05
Track state changes with a has many association

tatyree:
Previously, I added named timestamps for each state.  After discussion
with Paul, it seemed more sensible to change this to use a has many
association where each change is tracked indiviudally. We can use the
created_at timestamp from VisitStateChange to calculate processing
times and this approach allows for new states to be added.


threedaymonk:
Looks fine to me.

PR #88, opened 2016-01-05
Update prisons from old app

threedaymonk:
This imports and cleans up prison data from old app with effect from 11 Jan 2016, i.e. including Whitemoor and excluding old data from before this date.


tatyree:
LGTM, @kyliemulholland should probably take a look as well...

PR #89, opened 2016-01-06
Extract estate to its own table

threedaymonk:
It makes reporting queries more straightforward if estate is normalised to a separate table rather than being a (possibly unique, possibly duplicated) string on each prison.

### DIFF
--- a/db/migrate/20160106102516_extract_estate_to_table.rb
+++ b/db/migrate/20160106102516_extract_estate_to_table.rb
@@ -1,0 +1,42 @@
+class ExtractEstateToTable < ActiveRecord::Migration
+  def up
+    create_table :estates, id: :uuid do |t|
+      t.string :name, null: false
+      t.timestamps
+    end
+
+    add_index :estates, :name, unique: true
+    add_column :prisons, :estate_id, :uuid
+
+    execute <<-SQL
+      INSERT INTO estates (name)
+      SELECT DISTINCT estate FROM prisons
+    SQL
+
+    execute <<-SQL
+      UPDATE prisons
+      SET estate_id = estates.id
+      FROM estates
+      WHERE estate = estates.name
+    SQL
+
+    change_column_null :prisons, :estate_id, false
+    remove_column :prisons, :estate
+    add_index :prisons, :estate_id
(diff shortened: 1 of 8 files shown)


db/migrate/20160106102516_extract_estate_to_table.rb
```diff
@@ -0,0 +1,41 @@
+class ExtractEstateToTable < ActiveRecord::Migration
+  def up
+    create_table :estates, id: :uuid do |t|
+      t.string :name, null: false
+      t.timestamps
+    end
+
+    add_index :estates, :name, unique: true
+    add_column :prisons, :estate_id, :uuid
```
tatyree: Don't we want an index and a foreign key constraint on this?

db/migrate/20160106102516_extract_estate_to_table.rb
```diff
@@ -0,0 +1,41 @@
...
+    execute <<-SQL
+      UPDATE prisons
+      SET estate_id = estates.id
+      FROM estates
+      WHERE estate = estates.name
+    SQL
+
+    change_column_null :prisons, :estate_id, false
+    remove_column :prisons, :estate
+    add_foreign_key :prisons, :estates
```
tatyree: Oh, oops about the FK.

tatyree: However, do we want an `on_delete: :cascade` here?

threedaymonk: No. Whilst that would preserve relational integrity, I don't think we would ever want to make it easy to delete everything!

tatyree: How about `on_delete:  :nullify` instead?

threedaymonk: I don't think an estate should ever be deleted, and I don't think a prison should be allowed to have no estate.

tatyree: The scenario I envisaged was an entire estate being removed.  I suppose we could deal with that by disabling all the prisons, though.  Ok, I'm satisfied...

PR #90, opened 2016-01-06
Document the process for merging/splitting prisons

threedaymonk:


README.md
```diff
@@ -411,6 +411,25 @@ change the UUID.
 You can't. This is because historical bookings still refer to that prison.
 Disable it instead (see above).
 
+#### Merging two prisons
+
+If two formerly-separate parts of an estate now have the same booking team and
```
tatyree: We all know what "formerly-separate parts of an estate" means, but I'm not sure that would be the case with someone new coming to the team.

threedaymonk: I'm open to any suggestions. I think it's explained in the next section, though.

tatyree: Yes, but it requires a bit of interpretation.  How about leading with an explanation of estate?  Maybe: 

"An `estate` is the way the application groups prisons that share common characteristics but otherwise handle booking separately."

README.md
```diff
@@ -74,6 +74,15 @@ No slots will be listed for `unbookable` days. For `anomalous` days, the slots
...
 
+Each prison belongs to an estate.
+
+#### `Estate`
+
+An estate groups prisons that share common characteristics but otherwise handle
+booking separately. This might be separate wings with different visiting hours,
+or the main and high-security parts of a prison that are handled by different
+booking teams.
+
```
tatyree: :+1:

PR #91, opened 2016-01-06
Validate the return status for requested pages

tatyree:
The smoke tests previously assumed that requests never failed. This
resulted in test failures where Capayabara could not find the trigger
content for the next step in the returned error page.

This fixes that by checking the return status as well as the current
path.

### DIFF
--- a/smoke_test/helpers/http_status_validation.rb
+++ b/smoke_test/helpers/http_status_validation.rb
@@ -1,0 +1,12 @@
+module HttpStatusValidation
+module_function
+
+  def validate_response_status!
+    if (400..599).include?(page.status_code)
+      puts "Failed: bad status #{page.status_code} for #{page.current_path}"
+      fail
+    end
+  end
+
+  private_class_method :validate_response_status!
+end
--- a/smoke_test/smoke_test.rb
+++ b/smoke_test/smoke_test.rb
@@ -4,6 +4,7 @@ require 'phantomjs/poltergeist'
 require 'capybara/poltergeist'
 
 require_relative 'helpers/imap_processor'
+require_relative 'helpers/http_status_validation'
 require_relative 'state'
 require_relative 'mail_box'
 require_relative 'steps/base_step'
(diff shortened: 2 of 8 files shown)


smoke_test/helpers/http_status_validation.rb
```diff
@@ -0,0 +1,16 @@
+module HttpStatusValidation
+module_function
+
+  def validate_response_status!
+    unless page.status_code == 200
+      puts <<-EOF
+Step Failed: Expected status 200, got status #{page.status_code}
+Step Failed: Request path was: #{page.current_path}
+      EOF
```
threedaymonk: You could use `strip_heredoc` to avoid the obtrusive outdenting.

smoke_test/helpers/http_status_validation.rb
```diff
@@ -0,0 +1,16 @@
+module HttpStatusValidation
+module_function
+
+  def validate_response_status!
+    unless page.status_code == 200
```
threedaymonk: I think it would be better to check for 4xx or 5xx, as 2xx and 3xx responses are fine and redirects should be handled by Capybara.

smoke_test/helpers/http_status_validation.rb
tatyree: It's not rails :(

I didn't want in add a dependency for active_support just for that.

threedaymonk: Ah. In that case, perhaps just two lines of `puts "..."`?

tatyree: Yeah, ok.

smoke_test/helpers/http_status_validation.rb
```diff
@@ -0,0 +1,12 @@
+module HttpStatusValidation
+module_function
+
+  def validate_response_status!
+    if page.status_code.to_s.match(/^(4|5)\d\d/)
```
threedaymonk: Turning an integer into a string to check a digit offends me.

tatyree: Ok, ok, I'm tired :/

threedaymonk: `(400..599).include?(page.status_code)` would probably be most expressive, though `>= 400` would also work.

PR #92, opened 2016-01-06
Custom error pages

mloughran:
The approach (generating static pages) is copied from the old app (https://github.com/ministryofjustice/prison-visits/blob/master/lib/tasks/static_pages.rake), with a couple of small tweaks:

* The rake task is automatically invoked on assets:precompile, but is not a dependency of it
* Only the 404 & error pages are precompiled rather than all the other static pages – that seems unnecessary

There was no custom 422 page in the old app, so I removed it – any objection?


config/locales/en/pages.yml
```diff
@@ -1,5 +1,31 @@
...
+    "503":
+      title: We are experiencing technical difficulties
+      body_html: |
+        <p>The prison visits booking system is not available as we're carrying
+        out some planned maintenance. You can try again later or book your
+        visit by phone instead.</p>
+
+        <p>You can try again later or book your visit by phone instead. See the
+        <a href="http://www.justice.gov.uk/contacts/prison-finder">prison
+        finder</a> for contact details.</p>
```
threedaymonk: I think it would be worth splitting these paragraphs out to two fields, and feeding in the Prison Finder link: see [other template](https://github.com/ministryofjustice/prison-visits-2/blob/4c64d9ac6299e036944e07cb7430aa29b9608d95/app/views/feedback_submissions/new.html.erb#L6-L7) for an example.

lib/tasks/assets.rake
```diff
@@ -0,0 +1,30 @@
...
+      '/pages/404' => '404.html',
+      '/pages/500' => '500.html',
+      '/pages/503' => '503.html'
+    }
+
+    app = ActionDispatch::Integration::Session.new(Rails.application)
+
+    pages.each do |route, output|
+      puts "Generating #{output}..."
+      outpath = File.join([Rails.root, 'public', output])
```
threedaymonk: This could be `Rails.root.join('public', output)`, I think.

config/locales/en/pages.yml
mloughran: Yes it does, I hadn't spotted that

lib/tasks/assets.rake
```diff
@@ -0,0 +1,30 @@
...
+      resp = app.get(route)
+      if resp == 200
+        File.delete outpath if File.exist?(outpath)
```
threedaymonk: This seems redundant: `File.open(…, 'w')` will overwrite any existing file.

lib/tasks/assets.rake
```diff
@@ -0,0 +1,30 @@
...
+        File.open(outpath, 'w') do |f|
+          f.write(app.response.body)
+        end
+      else
+        puts "Error generating #{output}!"
```
threedaymonk: Could we `fail` instead, so that we catch the problem early?

config/locales/en/pages.yml
```diff
@@ -1,5 +1,31 @@
 en:
   pages:
+    "404":
+      title: This page cannot be found
+      body_html: |
+        <p>You can <a href="/">start a new prison visit request</a> or book
```
threedaymonk: This one could be `new_booking_request_path`.

threedaymonk:
I think it might be nice if these were exercised as part of the test suite.

tatyree:
LGTM pending @threedaymonk's comments and a rebase with master.

config/locales/en/pages.yml
mloughran: @threedaymonk I'm not convinced that separating the paragraphs is useful thought – surely having one or two paragraphs is a copy decision and it's best not to couple to the template. Personally I'd rather content editors were able to write markdown rather than HTML in this file.

mloughran:
@threedaymonk what makes sense to test for you? I can test the page generation, but it seems a bit pointless, and the other static pages are not tested. The rake task could be tested, although personally I'm not convinced it's worth the effort.

config/locales/en/pages.yml
threedaymonk: I'm thinking about translation, mainly, and I think there are several reasons to prefer to break it up:

1. It allows us to change structural or presentational (it exists, alas) HTML separately from translations.
2. Translators have less chance to break the layout by accident.
3. If we change part of the copy, it's clearer that there's only a small section whose translation should be changed.

lib/tasks/assets.rake
mloughran: Yes, that's a better idea.

threedaymonk:
I'm just thinking that, at the moment, if something goes wrong with the generation process, we won't realise until we deploy, which isn't a good time to find out. Even just running the rake task as part of the test run would be fine, I think, though I'm not sure if the asset pipeline will alow that.

config/locales/en/pages.yml
mloughran: I'm not really sold... For 1, I can't see a scenario where these paragraphs were split / joined without reference to the copy, but I can foresee plenty of scenarios where copy changes dictate adding or removing paragraphs. In terms of keeping changes. 2 and 3 are valid, but I don't think in either case that coupling paragraphs between the copy file and the templates is a good solution to either problem.

And compare with the terms and conditions page: https://github.com/ministryofjustice/prison-visits-2/blob/master/config/locales/en/pages.yml#L10. There's no way it makes sense to split that page up, while allowing markdown to be written would be more user friendly and safer.

threedaymonk: You might be right. If you can figure out a way to allow Markdown in translations in a similar way to HTML, I'd be all for it.

mloughran:
This is ready for another review. @threedaymonk I have not split those body paragraphs for now, but will investigate markdown separately.

lib/tasks/assets.rake
```diff
@@ -0,0 +1,12 @@
+Rake::Task['assets:precompile'].enhance do
+  Rake::Task['assets:generate_static_pages'].invoke
+end
+
+namespace :assets do
+  desc 'Generates static pages'
+  task generate_static_pages: :environment do
+    StaticPageGenerator.generate!('/pages/404' => '404.html',
+                                  '/pages/500' => '500.html',
+                                  '/pages/503' => '503.html')
```
threedaymonk: This ought to be:

```ruby
StaticPageGenerator.generate!(
  '/pages/404' => '404.html',
  '/pages/500' => '500.html',
  '/pages/503' => '503.html'
)
```

for consistency with the rest of the codebase.

spec/services/static_page_generator_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'rails_helper'
+
+RSpec.describe StaticPageGenerator do
+  it "renders pages and save them to the filesystem" do
+    expected_file_path = Rails.root.join('public', '404.html')
+    File.delete(expected_file_path) if File.exist?(expected_file_path)
+
+    described_class.generate!('/pages/404' => '404.html',
+                              '/pages/500' => '500.html',
+                              '/pages/503' => '503.html')
```
threedaymonk: The same applies here.

PR #93, opened 2016-01-07
Move NOMIS ID and finder slug to estate and seed estates separately

threedaymonk:
As the Prison Finder slug and the NOMIS ID are per-estate rather than per-prison, it makes the data model more consistent if this information is stored on the estate. This also simplifies the seed data by reducing duplication.

### DIFF
--- a/app/services/estate_seeder.rb
+++ b/app/services/estate_seeder.rb
@@ -1,0 +1,17 @@
+class EstateSeeder
+  ImportFailure = Class.new(StandardError)
+
+  def self.seed!(base_path)
+    estates = YAML.load(File.read(base_path.join('estates.yml')))
+    seeder = new
+    estates.each do |nomis_id, attributes|
+      seeder.import nomis_id, attributes
+    end
+  end
+
+  def import(nomis_id, attributes)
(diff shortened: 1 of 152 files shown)


mloughran:
Looks good to me

app/services/estate_seeder.rb
```diff
@@ -0,0 +1,17 @@
...
+    estates = YAML.load(File.read(base_path.join('estates.yml')))
+    seeder = new
+    estates.each do |nomis_id, attributes|
+      seeder.import nomis_id, attributes
+    end
+  end
+
+  def import(nomis_id, attributes)
+    estate = Estate.find_or_initialize_by(nomis_id: nomis_id)
+    entry = EstateSeeder::SeedEntry.new(nomis_id, attributes)
```
tatyree: I understand what this is doing, but it took me a bit.  I don't think the intent is as clear as it could be.

threedaymonk: Do you have any ideas for how it might be clearer?

tatyree: I'm not sure.   Rename it to something like EstateSeeder::BuildSlugAndName?

PR #96, opened 2016-01-07
Prototype use of markdown for static pages

mloughran:
Thoughts on this before I bother converting any other pages?

I have renamed the key to `body_md` which makes the type explicit to content writers, but no magic happens based on the suffix (`t` has enough magic already imo).


tatyree:
I'm dubious, @mloughran.  It increases our dependencies but I'm not sure it offers that much over the html.  Is markdown Leigh's preferred format?

mloughran:
@tatyree I agree the value for the 404 is low, but for other pages I think it's a lot friendlier for content writers (and Leigh has expressed a slight preference for markdown). Besides that, it's harder to generate invalid HTML, and invalid markdown will raise errors and therefore be easily caught in tests.

tatyree:
@mloughran Ok, I'm convinced.  @threedaymonk, what's your take?

spec/controllers/pages_controller_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'rails_helper'
+
+RSpec.describe HighVoltage::PagesController do
```
tatyree: I'm OK with it if you want to disable rubocop on this one.  I can't find the exact rule name, though.

PR #97, opened 2016-01-11
Enable double-submit protection on all forms

threedaymonk:
The script to do this is already in the app: it only needs an appropriate class on each form.


SteveMarshall:
Is there anything we can do Rails-side to gracefully handle double-submission, too?

tatyree:
@SteveMarshall I'm going to go ahead and merge this one.  @threedaymonk and I discussed, and we only need to check uniqueness on the final submit.  I'll add a trello card for that.

PR #98, opened 2016-01-11
Switch interface language, part 1

threedaymonk:
This is preparatory work for adding Welsh: it moves all public-facing paths under `/en/` and makes the code locale-aware. This enables us to go live with a URL structure that supports Welsh.


tatyree:
LGTM.  @mloughran?

mloughran:
Yes, looks good.

tatyree:
I'm going to hold off merging this until Lukasz gets staging working.  If I don't, then the smoke tests will fail as the container is still on an old version.

PR #99, opened 2016-01-11
Count visits by prison, state and prison and state

tatyree:
Report visit counts using various facets

After discussing with Paul and Martyn, I decided to refactor this using
straight SQL instead of trying to cobble it together with a combination
of ActiveRecord and SQL. This is the first step. Once we agree on this
approach, I will work on the centiles.


app/models/visit.rb
```diff
@@ -28,6 +28,16 @@ def total_number_of_visitors
     :allowance_will_renew?, :allowance_renews_on,
     to: :rejection
 
+  scope :by_day, -> {
+    group('id, extract(day from created_at::timestamp::date)::integer')
+  }
+  scope :by_week, -> {
+    group('id, extract(week from created_at::timestamp::date)::integer')
```
threedaymonk: This needs to include the `isoyear` to be accurate for more than the first year.

app/services/metrics.rb
```diff
@@ -0,0 +1,105 @@
...
+
+  def visit_count_by_prison_and_state
+    result = execute_query_and_type_result <<-SQL
+      SELECT prisons.name AS prison_name,
+             processing_state,
+             COUNT(*)
+        FROM visits
+      INNER JOIN prisons ON prisons.id = visits.prison_id
+      GROUP BY processing_state,
+               prison_name
```
threedaymonk: I'm wondering whether we should make these database views. We could map them to Active Record objects and it might make the code more natural on the Ruby side.

tatyree:
I'm not sure `services/metrics.rb` shouldn't be moved into its own category: `app/metrics/counters.rb`, for example.  Thoughts?

threedaymonk:
@tatyree Moving it to `app/metrics` sounds very sensible to me.

config/application.rb
```diff
@@ -13,10 +13,11 @@
 
 module PrisonVisits
   class Application < Rails::Application
+    config.active_record.schema_format = :sql
```
threedaymonk: I always end up here. Perhaps this should be the default schema format in Rails.

threedaymonk:
Looks good. I think database views do work well here: both the Ruby and the SQL end up being straightforward in their own domains.

app/metrics/counters.rb
```diff
@@ -0,0 +1,29 @@
...
+    def self.run
+      first.count
+    end
+  end
+
+  class CountVisitsByState < ActiveRecord::Base
+    def self.run
+      all.each_with_object({}){ |counter, result|
+        result[counter.processing_state] = counter.count
+      }
```
threedaymonk: This might be clearer as:

```ruby
all.map { |row| [row.processing_state, row.count] }.to_h
```

app/metrics/support/counter_support.rb
```diff
@@ -0,0 +1,21 @@
...
+      else
+        result[prison] = hash_visit_data(visit)
+      end
+    end
+  end
+
+protected
+
+  def hash_visit_data(visit)
+    visit.values.reverse.inject { |result, value|
```
threedaymonk: Neither the name nor the code make the intention clear here.

app/metrics/support/counter_support.rb
```diff
@@ -0,0 +1,21 @@
+module CounterSupport
+  def run
```
threedaymonk: I wonder whether `run` is the right name: it seems to imply some kind of process, whereas this is fetching and formatting the current data.

app/metrics/support/counter_support.rb
```diff
@@ -0,0 +1,21 @@
...
+    all.each_with_object({}) do |visit, result|
+      visit = visit.attributes
+      prison = visit.delete('prison_name') || 'all'
+      if result.key?(prison)
+        result[prison] = result[prison].deep_merge(hash_visit_data(visit))
+      else
+        result[prison] = hash_visit_data(visit)
+      end
```
threedaymonk: Could just be:

```ruby
result[prison] = result.fetch(prison, {}).deep_merge(...)
```

app/metrics/counters.rb
```diff
@@ -0,0 +1,27 @@
...
+module Counters
+  class CountVisits < ActiveRecord::Base
+    def self.fetch_and_format
+      first.count
+    end
+  end
+
+  class CountVisitsByState < ActiveRecord::Base
+    def self.fetch_and_format
+      all.map { |row| [row.processing_state, row.count] }.to_h
```
dan-palmer: could be:

```ruby
pluck(:processing_state, :count).to_h
```

dan-palmer:
Worth knowing about: https://github.com/thoughtbot/scenic

tatyree:
@dan-palmer I looked at it, but it didn't seem to buy very much for the price of a dependency.

app/metrics/counters.rb
threedaymonk: Even better! :+1:

app/metrics/support/counter_support.rb
```diff
@@ -0,0 +1,26 @@
...
+        order_and_hash_visit_values(visit)
+      )
+    end
+  end
+
+protected
+
+  # Takes the attributes of a visitor counter object, extracts the values,
+  # orders them from least specific to most specific and returns a hash of
+  # the resutls.
```
threedaymonk: 'resutls'

app/metrics/support/counter_support.rb
```diff
@@ -0,0 +1,26 @@
...
+  #
+  # For example:
+  #
+  # isoyear => { month => { day => { state => count}}}
```
threedaymonk: Might be worth saying that the input in this example is:

```
# [isoyear, month, day, state, count]
```

config/application.rb
```diff
@@ -13,10 +13,11 @@
...
     config.phase = 'live'
     config.product_type = 'service'
 
-    config.autoload_paths += %w[ app/mailers/concerns ]
+    config.autoload_paths += %w[ app/mailers/concerns app/metrics ]
```
threedaymonk: Is this definitely necessary? I thought all paths under `app` that weren't already special were automatically included in autoload paths.

threedaymonk: … and experimentation supports that.

app/metrics/support/counter_support.rb
```diff
@@ -0,0 +1,26 @@
+module CounterSupport
+  def fetch_and_format
+    all.each_with_object({}) do |visit, result|
+      visit = visit.attributes
+      prison = visit.delete('prison_name') || 'all'
```
dan-palmer: I can't see a test for the `'all'` prison case 😭

tatyree: Good catch.  I meant to put that in but forgot.  Doing it...

PR #100, opened 2016-01-12
Switch interface language, part 2

threedaymonk:
In order to be able to display Welsh to users, they must be able to select it.

This continues #98.

The exact design pattern for the language switching link is still to be confirmed.


mloughran:
LGTM

PR #101, opened 2016-01-12
Disable query caching for healthchecks

SteveMarshall:
By default, `ActiveRecord::QueryCache` will try to connect to the database before anything else happens to prepare cached queries, which breaks for the healthcheck. We don't need the database in `ping.json`, and `healthcheck.json` should be able to control when we connect to the database (so it can report on the outcome), so connecting before either isn't worth it.

In the old app, we got around this problem by completely disabling query-caching, but that seems like a bad idea for a database-backed app.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -22,6 +22,7 @@ gem 'uglifier', '>= 1.3.0'
 gem 'uri_template'
 gem 'virtus'
 gem 'zendesk_api'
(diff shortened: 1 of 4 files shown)


mloughran:
@SteveMarshall I wonder whether a cleaner solution would be to serve ping.json from some middleware which is added to the middleware chain before the `ActiveRecord::QueryCache` middleware (see `rake middleware`)?

tatyree:
Closing this one as it is no longer an issue.

PR #104, opened 2016-01-13
Rationalise seed data

threedaymonk:
As we have now synchronised our seed data, we can allow the old and new app data to diverge. We need them to diverge in order to support localisation of prison names and addresses, and this is a good time to:

* Make the address a string, instead of joining lines
* Make the keys match those used internally
* Require that every enabled prison has full contact details (address, phone, email)

This simplifies things and means that when we add the translations, the translated addresses can match the default addresses without having to do some custom processing.


mloughran:
LGTM

PR #105, opened 2016-01-14
Add missing completion page

threedaymonk:
There *was* a page when a booking request was completed, but it had nothing more than a title on it. This restores the helpful information that was present in the old application.


tatyree:
LGTM. @mloughran?

mloughran:
Yes, looks good. Would be nice if that static text was markdown, but that can happen in a separate PR. I'll merge.

PR #106, opened 2016-01-14
Adds compilation of static pages on the container runtime

lukaszraczylo:


mloughran:
This shouldn't make any difference since the static page generation is triggered by `assets:precompile`... Does it?

PR #109, opened 2016-01-15
Reset booking request after 20 minutes

threedaymonk:
If someone abandons a booking request, the browser will return to the start of the booking process after 20 minutes in order to try to keep their data private.

After 15 minutes, a prompt is displayed, giving them 5 minutes to reset the counter.

### DIFF
--- a/config/locales/en/views.yml
+++ b/config/locales/en/views.yml
@@ -425,6 +425,10 @@ en:
         Please don’t bring anything restricted or illegal to the prison. The
         <a href="%{url}">%{prison} prison page</a> has more information about
         what you can bring.
+    timeout_prompt:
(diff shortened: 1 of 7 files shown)


config/locales/en/views.yml
```diff
@@ -425,6 +425,10 @@ en:
         Please don’t bring anything restricted or illegal to the prison. The
         <a href="%{url}">%{prison} prison page</a> has more information about
         what you can bring.
+    timeout_prompt:
+      title: Your session will expire in %{minutes_left} minutes
+      continue_question: Would you like to continue?
+      yes_continue: "Yes"
```
threedaymonk: "Why does it say 'true' instead of 'Yes'?! Oh, YAML."

PR #111, opened 2016-01-18
Move admin pages under /en/

threedaymonk:
This means that we can say that all pages on the new app (with the exception of `/ping.json` and `/healthcheck.json` and the `/pages/*id` route used only internally to generate static pages) are under a locale prefix of either `/en/` or `/cy/`, which simplifies routing to the old and new apps during the parallel running period.


tatyree:
LGTM.  @mloughran?

mloughran:
Makes sense

PR #116, opened 2016-01-20
Render error pages dynamically

mloughran:
The previous approach caused more problems than it was worth, and was unnecessarily complex.

An unstyled error page isn’t the end of the world.


config/routes.rb
```diff
@@ -4,6 +4,10 @@
 
   get '/', to: redirect(ENV.fetch('GOVUK_START_PAGE', '/en/request'))
 
+  %w[ 404 500 503 ].each do |code|
+    get code => 'errors#show', status_code: code
```
threedaymonk: This would be more consistent with other routes via the `on:` syntax:

```ruby
get code, to: 'errors#show', status_code: code
```

spec/controllers/errors_controller_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+  # Test that each static page is generated correctly
+  Dir.glob(Rails.root.join('app', 'views', 'errors', '*.erb')).each do |file|
+    status_code = file.split('/').last.split('.').first
+
+    describe "when rendering #{status_code} page" do
+      before do
+        get :show, status_code: status_code
+      end
+
+      it { should respond_with(status_code.to_i) }
```
tatyree: Could you change to the new syntax, please? 

```ruby
expect(response.status).to eq(status_code.to_i) 
```

spec/controllers/errors_controller_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'rails_helper'
+
+RSpec.describe ErrorsController do
+  render_views
+
+  # Test that each static page is generated correctly
+  Dir.glob(Rails.root.join('app', 'views', 'errors', '*.erb')).each do |file|
+    status_code = file.split('/').last.split('.').first
```
tatyree: I think it would be more declarative if it was just a list of the statuses we wanted covered: 

```
[404, 500, 503]. each do |status_code|...
```

spec/controllers/high_voltage/pages_controller_spec.rb
```diff
@@ -3,16 +3,13 @@
...
-    page_name = file.split('/').last.split('.').first
-
+  %w[ cookies terms_and_conditions unsubscribe ].each do |page_name|
     describe "when rendering #{page_name}" do
       before do
         get :show, id: page_name
       end
 
-      it { should respond_with(:success) }
+      specify { expect(response.status).to eq(200) }
```
dan-palmer: There is an rspec matcher for status codes that might read a bit better here. It takes the integer value of the response code(200, 201 etc) or the action name(:ok, :created etc).

So you could write:

```ruby
specify { expect(response).to have_http_status(:ok) }
```

threedaymonk: Agreed. We're already using these matchers elsewhere (in `healthcheck_controller_spec.rb`).

dan-palmer: The doc output from the test would read something like:

```
HighVoltage::PagesController
  when rendering cookies
    should respond with status code :ok (200)
```

mloughran: I agree the test output is a bit nicer, but really, the test reads pretty clearly as is, and I'd rather use 200 than `:ok`. Everyone knows what 200 means, `:ok` is unnecessary indirection.

threedaymonk: > Everyone knows what 200 means, `:ok` is unnecessary indirection.

For 200, yes. But can everyone remember which is which out of `301` and `302`? Or the various `4xx` codes? I have to check every time.

I don't think this is indirection so much as description. It expresses the intention of the test better, and reports the result more clearly. There's no harm in it, and we'll get more helpful error messages.

Compare this:

```ruby
expect(response.status).to eq(502)
```

```
Failure/Error: expect(response.status).to eq(502)

  expected: 502
       got: 200

  (compared using ==)
```

with this:

```ruby
expect(response).to have_http_status(:bad_gateway)
```

```
Failure/Error: expect(response).to have_http_status(:bad_gateway)
  expected the response to have status code :bad_gateway (502) but it was :ok (200)
```

I much prefer the latter.

PR #118, opened 2016-01-21
Implement basic prisons API

threedaymonk:
This API lists all enabled prisons, and returns details about an individual prison.

I've added links in an attempt to make it conform to HATEOAS. That's harmless, but I'm not entirely convinced that it's useful.

A [Swagger](http://swagger.io/) definition is included. The code generated by Swagger's client codegen is messy and buggy, so it's probably useful only for documentation.

Locale is determined by the `Accept-Language` header.

There is now working code using this API in https://github.com/ministryofjustice/prison-visits-public/pull/3, so it seems to be fit for purpose.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,6 +5,8 @@ gem 'rails', '~> 4.2.3'
 
 gem 'govuk_frontend_toolkit', '2.0.1'
 gem 'high_voltage'
+gem 'http_accept_language'
+gem 'jbuilder'
 gem 'kramdown'
 gem 'logstasher'
 gem 'moj_template', '0.21.0'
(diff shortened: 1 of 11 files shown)


tatyree:
LGTM :+1:

api.yaml
```diff
@@ -0,0 +1,65 @@
+swagger: "2.0"
```
tatyree: On reflection, I'm uncertain about including this.  It seems counter productive to include a swagger definition when we have reviewed swagger and decided it is not suitable.

threedaymonk: I think you might be right. I'll take it out.

PR #119, opened 2016-01-21
Moved assets generation to container build process

lukaszraczylo:


mloughran:
@lukaszraczylo I just removed `SECRET_KEY_BASE` – this is no longer required.

PR #123, opened 2016-01-22
Log to stdout in production

threedaymonk:
This makes it easier to deliver logs to the Elasticsearch/Logstash/Kibana stack. It does not affect environments other than production.


lukaszraczylo:
LGTM.

PR #129, opened 2016-01-26
Prevent sidekiq daemon from going into background

lukaszraczylo:
Sidekiq has been run as a daemon in app container. When I moved it to the separate container it was still in background mode which caused container to immediately quit.


tatyree:
@lukaszraczylo if you'll just stick the reason for this, as discussed on Hipchat, into the description, I'll merge.

PR #130, opened 2016-01-26
Report processing time by percentile

tatyree:
Display processing time metrics for all prisons and individual prisons and also broken down by calendar week and calendar date.


threedaymonk:
As far as the code is concerned, I think this is fine. However, we're still having excessive movement in `structure.sql`, so I wonder if we should try out that view schema plugin that we looked at before.

PR #131, opened 2016-01-26
Redirect mails to maildrop based on ENV variables

tatyree:
Currently, staging sends real emails.  This changes the seeder such that
it writes a `maildrop.dsd.io` address for each prison if the
`SMTP_HOSTNAME` environment variable is set to `maildrop.dsd.io`.  I
decided to do it this way as it prevents us from having to change the
smoketests to account for different environments. This would have been
the case with any of the other mail interceptor gems.

### DIFF
--- a/README.md
+++ b/README.md
@@ -255,6 +255,14 @@ These configure email delivery in the production environment. `SMTP_DOMAIN` is
 also used when generating the `no-reply@` address and the `feedback@` stand-in
 address used when submitting feedback without an email address to Zendesk.
 
+#### `SMTP_HOSTNAME` and staging
+
+If you set `SMTP_HOSTNAME` to `maildrop.dsd.io` and seed prisons, it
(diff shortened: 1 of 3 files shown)


threedaymonk:
If we set `SMTP_HOSTNAME` to `maildrop.dsd.io`, does Maildrop show emails there regardless of domain? If so, then we don't need the address mangling.

Does sending all email to Maildrop (via `SMTP_HOSTNAME`) stop the smoke tests from running on staging?

Having `maildrop.dsd.io` hard-coded in the middle of `SeedEntry` makes me feel unhappy.

You can send an email to a `mailbox.dsd.io` address through the internet, and [it will be delivered](http://maildrop.dsd.io/inbox/test-arbitrary-delivery). This presents an alternative course, which is to have another variable – `PRISON_EMAIL_DOMAIN`, say – and use that to generate prison email addresses.

PR #156, opened 2016-02-16
Use updated scenic gem for db view migrations

tatyree:
We are having a lot of trouble with ordering using hand-written queries
migrated via `structure.sql`. The percentile view migrations are also
failing on production on RDS, which I have not been able to reproduce  I
suspect this may be down to encoding.

We tried scenic before in order to try and deal with the ordering
problem. Unfortunately, the version we installed had a bug. That has
been fixed.

I am changing back to it in an effort to better manage the ordering
problem and to see if it solves the RDS production migtation issue.


dan-palmer:
:ship:

PR #160, opened 2016-02-18
Move each DROP VIEW to the relevant migration

tatyree:
To ensure accuracy of table name. The monolithic migration was failing
due to an incorrect pluralisation.


mloughran:
@tatyree agree this is better

mloughran:
@tatyree look like you've missed out a couple of migrations:

* https://github.com/ministryofjustice/prison-visits-2/blob/master/db/migrate/20160216110548_create_count_visits_by_prison_and_states.rb
* https://github.com/ministryofjustice/prison-visits-2/blob/master/db/migrate/20160216134620_create_distributions.rb

PR #161, opened 2016-02-18
Re order view migrations to sort failures

tatyree:
This one appears to have updated the version even though it broke.  I am changing the ordering in an attempt to fix this problem.


tatyree:
Closing this one as it has been fixed by another PR.

PR #163, opened 2016-02-18
Show validation error if dates are invalid

mloughran:
Previously, an error would be raised inside Virtus’ date coercion code.

Ideally we wouldn’t be using Virtus since it has significantly different behaviour (coersion and otherwise) to ActiveRecord – which adds significant extra complexity to the app.

See also related solution in MPS: https://github.com/ministryofjustice/moving-people-safely/blob/master/app/forms/form/date_handling.rb /ht @dan-palmer

### DIFF
--- a/lib/maybe_date.rb
+++ b/lib/maybe_date.rb
@@ -1,0 +1,25 @@
+UncoercedDate = Struct.new(:year, :month, :day)
+
+# Coerces date input into either a Date (if the date is valid), or an
+# UncoercedDate struct which can be redered back to the view for correction
+class MaybeDate < Virtus::Attribute
+  # This coercion is probably not as comprehensive as
+  # Virtus::Attribute::Date, but it is understandable and sufficient for
+  # our needs
+  # rubocop:disable Metrics/MethodLength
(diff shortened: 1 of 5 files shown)


lib/maybe_date.rb
```diff
@@ -0,0 +1,24 @@
...
+  # Virtus::Attribute::Date, but it is understandable and sufficient for
+  # our needs
+  # rubocop:disable Metrics/MethodLength
+  def coerce(value)
+    return value if value.is_a?(Date)
+
+    if value.is_a?(String)
+      Date.parse(value)
+    elsif value.respond_to?(:values_at)
+      ymd = value.values_at(:year, :month, :day).map(&:to_i)
```
dan-palmer: This will turn empty strings and non numeric characters to zeros - so any original user entered input will be lost when the view is rendered with errors. Is that ok?

@mloughran @tatyree

mloughran: Personally, I'm ok with this.

dan-palmer:
LGTM :ship: :ship: :boom: :rocket: :100:

PR #166, opened 2016-02-19
Allow prisons to accept booking requests while rejecting specific visitors

tatyree:
This is the behaviour of the old app.  I've been through the history and can see no reason why we changed it to work this way.  Prisons are requesting we fix it ASAP.

### DIFF
--- a/app/views/visitor_mailer/booked.html.erb
+++ b/app/views/visitor_mailer/booked.html.erb
@@ -16,7 +16,7 @@
   <strong><%= @visit.prisoner_number %></strong>
 </p>
 
-<% @visit.visitors.each_with_index do |visitor, index| %>
+<% @visit.visitors.allowed.each_with_index do |visitor, index| %>
   <p>
     <%= t('.visitor', n: index + 1) %>
     <strong><%= visitor.anonymized_name %></strong>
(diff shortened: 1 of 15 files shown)


app/views/visitor_mailer/booked.html.erb
```diff
@@ -32,6 +32,20 @@
   <p><%= t(".this_is_a_closed_visit_html") %></p>
 <% end %>
 
+<% if @visit.visitors.any?(&:banned?) %>
```
alan: Does line [19](https://github.com/ministryofjustice/prison-visits-2/pull/166/files#diff-0613bf27f26accef00d1ecc1a3b22690R19) need to be updated to exclude the banned visitors?

tatyree: Seems sensible.  I'll do it now.

app/views/visitor_mailer/booked.html.erb
```diff
@@ -32,6 +32,20 @@
...
+  <% @visit.visitors.select(&:banned?).each do |v| %>
+    <p><%= t('.visitor_banned', name: v.anonymized_name) %></p>
```
alan: I've tested locally and the people that are banned or not on the list are easy to miss in the email. Using the same formatting as the approved visitors could be a solution

screen shot 2016-02-25 at 09 54 22

PR #172, opened 2016-02-24
Improvements to staff email

alan:
Fixes https://trello.com/c/D9frS202/162-missing-s-in-the-staff-email

* Add missing `:`
* Introduce line break for email
* CSS tweaks to use the same font

screen shot 2016-02-24 at 10 54 14


alan:
Note the incorrect telephone number, is this a regression from the original app?

PR #180, opened 2016-02-26
Show the booked slot if visit has been approved

tatyree:
This was defaulting to the first slot requested even after the visit had
been accepted.  This logic show the booked slot if the visit is
accepted, and the first requested slot if it is not.

### DIFF
--- a/app/models/visit.rb
+++ b/app/models/visit.rb
@@ -35,9 +35,6 @@ class Visit < ActiveRecord::Base
         VisitStateChange.new(visit_state: visit.processing_state)
     end
 
-    after_transition any => %i[cancelled withdrawn],
-                     do: :send_cancellation_email
-
     event :accept do
       transition requested: :booked
     end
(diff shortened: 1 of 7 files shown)


mloughran:
Does it ever make sense to only show the first slot?

Why do staff need to receive a cancellation email if the visit was never booked?

tatyree:
The same template gets used for cancelled and withdrawn requests.  The granted_slot won't be set if the visit is withdrawn.

alan:
The use of withdrawn and cancelled for the cancel event has just caught me while fixing another bug. I think it would be clearer if they were split as they are sightly different, what do you think?

mloughran:
I have re-implemented this using a different template, and also fixed a further inconsistency in the email template. This has created a bit of duplication, but I think it's now simpler and clearer.

app/models/visit.rb
```diff
@@ -52,6 +49,14 @@ def total_number_of_visitors
...
       transition rejected: :rejected
     end
+
+    after_transition booked: :cancelled do |visit, _|
+      PrisonMailer.cancelled(visit).deliver_later
+    end
+
+    after_transition requested: :withdrawn do |visit, _|
+      PrisonMailer.withdrawn(visit).deliver_later
+    end
```
tatyree: Seems much clearer.

PR #181, opened 2016-02-29
Don't allow to re-cancel a visit

alan:
Updated the state machine to not allow transitions from cancelled to cancelled and moved the sending of the email inside the state machine.

I'd like to move the rest of the emails inside the state machine if you like this approach to have them consistent.

Fixes https://trello.com/c/Ql4Ip0dP/174-it-s-possible-to-resubmit-the-cancel-form-multiple-times-and-thereby-send-multiple-emails-to-a-prison

### DIFF
--- a/app/controllers/cancellations_controller.rb
+++ b/app/controllers/cancellations_controller.rb
@@ -1,9 +1,8 @@
 class CancellationsController < ApplicationController
   def create
     visit = Visit.find(params[:id])
-    if cancellation_confirmed?
+    if cancellation_confirmed? && visit.can_cancel?
       visit.cancel!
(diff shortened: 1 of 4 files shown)


mloughran:
@alan and I discussed moving the `visit.can_cancel?` logic inside the Visit model so that `visit.cancel!` would become idempotent, and agreed that in that case there should not be a transition from `cancelled -> cancelled`. However we could not decide on an API we were satisfied with so decided to merge this as is and potentially reconsider later. The same issue applies to other transitions which are currently defined to themselves, and this should be made consistent. Any thoughts @tatyree ?

PR #184, opened 2016-02-29
Full names on prison booking confirmed emails

alan:
For https://trello.com/b/lbxS986X/prison-visit-booking-v2

Start duplicating the email for booked visits so that we can have full names on the prison version.


alan:
Replaced by #189

PR #187, opened 2016-03-01
Configure on/off sengrid email validations

alan:
Use an environment variable `ENABLE_SENDGRID_VALIDATIONS` to enable / disable Sengrid email validations.

Until the variable is set it will default to `false` behaviour so it should be safe to deploy.

### DIFF
--- a/app/services/email_checker.rb
+++ b/app/services/email_checker.rb
@@ -56,7 +56,7 @@ private
   # rubocop:enable Metrics/MethodLength
 
   def override_sendgrid?
-    @override_sendgrid
+    @override_sendgrid || !Rails.configuration.enable_sendgrid_validations
   end
 
   def domain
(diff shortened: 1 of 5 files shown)


app/services/email_checker.rb
```diff
@@ -56,7 +56,8 @@ def compute_error
   # rubocop:enable Metrics/MethodLength
 
   def override_sendgrid?
-    @override_sendgrid
+    @override_sendgrid ||
+      ENV['ENABLE_SENDGRID_VALIDATIONS'].try(:downcase) != 'true'
```
tatyree: I find this syntax confusing and this ENV variable needs to be added to the readme.

alan: Spoke to @mloughran and agreed to move to Rails configuration to keep it consistent with the other environment variables.

Also the presence of the environment variable should be enough to enable the validations to keep consistency.

@tatyree I'll get working on that approach, let me know if that will still be confusing

PR #194, opened 2016-03-03
Timeout after 2s when using the Sendgrid API

alan:
Use `Timeout.timeout` around the Sendgrid client.

Currently SendgridToolkit uses HTTP Party without a timeout so if the Sendgrid
API becomes unresponsive booking a visit becomes unresponsive.

Ideally the timeout would be set on the HTTP client, however it is not possible
to set it by design. We should revisit this issue as the fix is not a good term
solution.

### DIFF
--- a/app/services/sendgrid_api.rb
+++ b/app/services/sendgrid_api.rb
@@ -1,6 +1,8 @@
 class SendgridApi
   extend SingleForwardable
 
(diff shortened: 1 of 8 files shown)


alan:
Will implement a very simple Sendgrid implementation to replace the existing gem as discussed

PR #195, opened 2016-03-04
Replace sendgrid gem and set timeout

alan:
Replace Sendgrid gem with a pool of custom Excon clients using a 2s timeout.

The existing specs around the Sendgrid usage are very thorough, they've been very useful to implement undocumented behaviour of the API.


app/services/email_checker.rb
```diff
@@ -83,4 +86,8 @@ def mx_records?
       Rails.configuration.mx_checker.records?(domain)
     end
   end
+
+  def sendgrid_api
+    @sendgrid_api ||= SendgridApi.new
```
mloughran: I would prefer that the API client was created by an initialiser (or at least only once). That way the persistent HTTP connection can be reused for multiple requests. This code opens a new connection for each rails request.

app/services/sendgrid_api.rb
```diff
@@ -1,39 +1,133 @@
 class SendgridApi
-  extend SingleForwardable
+  class Error < StandardError; end
 
-  def_single_delegators :new, :spam_reported?, :bounced?,
-    :remove_from_bounce_list, :remove_from_spam_list
+  TIMEOUT = 2 # seconds
+
+  def initialize
+    @connection = Excon.new(Rails.configuration.sendgrid_api_host,
```
mloughran: I think the configuration parameters (e.g. host) should be passed into the initializer. This code introduces unnecessary coupling with rails & the specific way we store the configuration.

app/services/sendgrid_api.rb
```diff
@@ -1,39 +1,133 @@
...
 
   def spam_reported?(email)
-    api { spam_reports.retrieve(email: email).any? }
+    route = 'spamreports.get.json'
+    query = { email: email }
+
+    response = get_request(route, query)
+
+    response.any?
+  rescue => e
```
mloughran: Generally not a fan of catching all exceptions... I don't think it's necessary in this case.

app/services/email_checker.rb
alan: Yes I know, went with this approach initially because I assume we are using Puma with forking, if that is the case we should use the `after_fork` method in the puma config

app/services/sendgrid_api.rb
alan: Good point, I'll change it

app/services/sendgrid_api.rb
```diff
@@ -1,39 +1,133 @@
...
+    route = 'bounces.delete.json'
+    body = { email: email }
+
+    response = post_request(route, body)
+
+    check_email_error(response)
+    true
+  rescue => e
+    Rails.logger.error("SendgridApi error: #{e.class} #{e}")
+    false
```
mloughran: This code is (unlike `bouced?` it would seem – but it's not obvious from the code) using exceptions for flow control and is using exceptions here for 2 completely separate things:

1. For flow control in the case that `check_email_error` raises an exception (because the email address doesn't exist). This shouldn't be implemented using exceptions in my opinion.
2. In the case of other exceptions – HTTP etc. Again same concern about catching everything.

Also, I think the consumer of this client should be the one to decide that HTTP errors, timeouts, etc. are not important. Returning false in the case of explicit failure (no email) and call failure (API down etc) is not good.

app/services/sendgrid_api.rb
```diff
@@ -1,39 +1,133 @@
...
     false
   end
 
-  def spam_reports
-    SendgridToolkit::SpamReports.new
+private
+
+  def credentials
+    if Rails.configuration.sendgrid_api_user.blank? ||
+       Rails.configuration.sendgrid_api_key.blank?
```
mloughran: Again, I suggest passing these into the initializer in order to decouple.

app/services/sendgrid_api.rb
```diff
@@ -1,39 +1,133 @@
...
+  end
+
+  # rubocop:disable Metrics/MethodLength
+  def get_request(route, query, timeout: TIMEOUT)
+    options = {
+      path: "api/#{route}",
+      expects: [200],
+      headers: {
+        'Accept' => 'application/json',
+        'Accept-Language' => 'en'
```
mloughran: I don't think we need `Accept-Language` header. Pretty sure that's just something you've copied over from pvb-public.

app/services/sendgrid_api.rb
```diff
@@ -1,39 +1,133 @@
...
+  # rubocop:disable Metrics/MethodLength
+  def post_request(route, body, timeout: TIMEOUT)
+    options = {
+      path: "/api/#{route}",
+      expects: [200],
+      headers: {
+        'Accept' => 'application/json',
+        'Accept-Language' => 'en'
+      },
+      query: URI.encode_www_form(body.merge(credentials)),
```
mloughran: We shouldn't be using query in the case of a POST request. I haven't looked at the sendgrid API design, but I'd be surprised if it doesn't accept `application/json`.

config/environments/test.rb
```diff
@@ -15,6 +15,9 @@
   config.active_support.test_order = :random
   config.active_support.deprecation = :stderr
 
+  config.sendgrid_api_host = ENV['SENDGRID_API_HOST'] || 'https://sendgrid.example.com/'
+  config.sendgrid_api_user = ENV['SMTP_USERNAME']
+  config.sendgrid_api_key = ENV['SMTP_PASSWORD']
```
mloughran: Why are these duplicated in each environment? Surely we can just configure in `application.rb`, or in a sendgrid specific initializer?

app/services/sendgrid_api.rb
alan: I agree with doing that generally, although I can see why the old code did it which I ported over (there are specs for catching `StandardError`).

By catching all it makes the client less likely to raise an error, and given we use it in the request cycle it seems to be a good idea in this case. Downside that there might be some errors caused by the app to be swallowed and go unnoticed, to mitigate this I could report any unexpected error to sentry? An unexpected error would be anything that is not an Excon or SendgridApi::Error.

app/services/email_checker.rb
mloughran: I'm not sure how we've got puma configured (I don't think we're using forking, but I may be wrong...). Can you investigate?

app/services/sendgrid_api.rb
alan: Yes, not keen on the use of errors as flow control. With this PR I wanted to use the existing specs as much as possible to be sure that I didn't miss anything and to keep the number of changes small.

Ideally I'd like to pull out the API calls to a separate class so that SendgridApi uses that and makes the decisions. Then if later on there is a good open source Sendgrid client the implementation of SendgridApi can be changed to use it

app/services/sendgrid_api.rb
alan: Good spot!

app/services/sendgrid_api.rb
alan: I thought the same, ended up using query because of existing specs, docs are unclear

config/environments/test.rb
alan: I'll move as my initial assumption to need different ways of initialisation was wrong

app/services/sendgrid_api.rb
mloughran: I think that's a good idea: separate the bits that are specific to this app, from a more generic client which appropriate design.

app/services/email_checker.rb
alan: We use the default configuration which is up to 16 threads, no workers

app/services/sendgrid_client.rb
```diff
@@ -0,0 +1,62 @@
...
+    @connection ||= Excon.new('https://api.sendgrid.com',
+      persistent: persistent,
+      connect_timeout: timeout)
+
+    @api_key = api_key
+    @api_user = api_user
+  end
+
+  def get_request(endpoint, query, timeout: TIMEOUT)
+    fail Error, 'no credentials set' if credentials_missing?
```
alan: This is to keep backwards compatibility. The current behaviour in `master` allows to run development without Sendgrid, and the rest of the test suite doesn't expect any API calls out to Sendgrid.

I'd like to refactor in a different PR so that Sendgrid is explicitly disabled in development and in test rework the specs so that it allows Sendgrid API calls.

app/services/sendgrid_api.rb
```diff
@@ -1,39 +1,97 @@
...
+    action = 'spamreports.get.json'
+    query = { email: email }
+
+    response = SendgridPool.instance.with { |conn|
+      conn.get_request(action, query)
+    }
+
+    return false if error_response?(response)
+
+    response.any?
```
alan: I removed the rescue of `StandardError` as suggested previously, however I'd like to put it back in to make the Sendgrid integration inside the request more robust.

I would do it a bit differently than before though, as part of rescuing the error I'd send the error to Sentry so that we get notified of unexpected errors. What are thoughts on doing this?

tatyree: :+1:

app/services/sendgrid_client.rb
tatyree: :+1:

tatyree:
LGTM

app/services/sendgrid_api.rb
```diff
@@ -1,39 +1,97 @@
...
+      true
+    else
+      false
+    end
   end
 
-  def bounces
-    SendgridToolkit::Bounces.new
+  def email_removed?(response)
+    if response.try(:key?, 'message') && response['message'] != 'success'
```
mloughran: Testing the negative condition feels awkward, and it misses cases (e.g. if there is no message at all). If we're going to rescue all errors I would suggest that we write this code with confidence about the expected types, so simply:

    if response['message'] == 'success'

app/services/sendgrid_api.rb
```diff
@@ -1,39 +1,97 @@
...
+    response = SendgridPool.instance.with { |conn|
+      conn.post_request(action, data)
+    }
+
+    return false if error_response?(response)
+
+    email_removed?(response)
+  rescue *RESCUABLE_ERRORS => e
+    Rails.logger.error("#{e.class.name}: #{e.message}")
+    false
```
mloughran: If feels like there's a lot of duplication in this code which we could abstract: namely fetching connection from the pool and rescuing errors.

config/application.rb
```diff
@@ -55,5 +55,8 @@ class Application < Rails::Application
     config.lograge.custom_options = lambda do |event|
       event.payload[:custom_log_items]
     end
+
+    config.sendgrid_api_user = ENV['SMTP_USERNAME']
+    config.sendgrid_api_key = ENV['SMTP_PASSWORD']
```
mloughran: Isn't it easier to put these in the sendgrid initializer? Then they don't even need adding to the rails config hash.

app/services/sendgrid_pool.rb
```diff
@@ -0,0 +1,21 @@
...
+        http_opts: client_attrs.fetch(:http_opts)
+      )
+    end
+  end
+
+  def with(timeout: nil)
+    @pool.with(timeout: timeout) do |conn|
+      yield conn
+    end
+  end
```
mloughran: Does the timeout inherit from the pool? Seems slightly odd to be passing in nil.

tatyree:
Overall :+1: Could you reword and add a bit of detail about the thinking behind the "Refactor initialization and connection pool" and "Refactor of SendgridClient" commits?

alan:
@tatyree updated those commit messages

mloughran:
@alan let me know when this is ready to be merged

alan:
@mloughran ready

tatyree:
Again.  LGTM.

PR #200, opened 2016-03-08
Re enable problematic views

tatyree:
These were broken because we were running Postgres 9.3 on our staging
and production RDS instances. We have requested that platforms upgrade
these to 9.4.  According to the AWS documentation, this is a trivial operation.

### DIFF
--- a/app/metrics/overdue.rb
+++ b/app/metrics/overdue.rb
@@ -1,0 +1,31 @@
+require 'support/counter_support'
+
+module Overdue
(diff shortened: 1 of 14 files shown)


tatyree:
This cannot be released to production until after the maintenance window on 10 March 2016 (01:36 - 02:06).  It is ready to go for staging.

alan:
Confirmed that we are running 9.4.4 in prod

PR #209, opened 2016-03-11
Report percentage of rejected visits

tatyree:
This is part of the PVB1 standard metrics table and is required by the
prisons.

### DIFF
--- a/app/metrics/rejections.rb
+++ b/app/metrics/rejections.rb
@@ -1,0 +1,30 @@
+require 'support/counter_support'
+
+module Rejections
+  class RejectionPercentage < ActiveRecord::Base
+    def self.fetch_and_format
+      pluck(:reason, :percentage).to_h
+    end
+  end
+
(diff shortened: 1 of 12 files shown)


app/metrics/support/counter_support.rb
```diff
@@ -7,6 +7,10 @@ def fetch_and_format(aggregate = nil)
     # The block on the deep_merge ensures that the final values
     # are added.  Without it, the value is the last value passed in.
     ordered_counters.each_with_object({}) do |values, result|
+      # Rails casts PG numeric as BD, which causes problems in the addition.
```
alan: How come is a problem? I'd have thought that floats would cause problems since they don't have precision

tatyree: Actually, it wasn't.  It looked like it was throwing error because it failed to convert the BigDecimal from its class representation.  However, the problem was with a failure elsewhere.  I left it in by accident.  Removing now.

PR #210, opened 2016-03-14
Update HVI-haverigg.yml

joseph-silverman:


joseph-silverman:
Tested in Staging, passed

PR #212, opened 2016-03-15
Metrics index UI

alan:
Index view of visit metrics for all time or the past week, similar to the one in pvb1

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -10,14 +10,14 @@ class ApplicationController < ActionController::Base
 
 private
 
+  # :nocov:
+
   def authorize_prison_request
     unless Rails.configuration.prison_ip_matcher.include?(request.remote_ip)
       Rails.logger.info "Unauthorized request from #{request.remote_ip}"
       fail ActionController::RoutingError, 'Not Found'
     end
   end
-
-  # :nocov:
 
   def append_to_log(params)
     @custom_log_items ||= {}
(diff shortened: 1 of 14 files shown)


app/controllers/application_controller.rb
```diff
@@ -12,6 +12,13 @@ class ApplicationController < ActionController::Base
 
   # :nocov:
 
+  def authorize_prison_request
+    unless Rails.configuration.prison_ip_matcher.include?(request.remote_ip)
+      Rails.logger.info "Unauthorized request from #{request.remote_ip}"
```
tatyree: :+1:

app/controllers/metrics_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+  end
+
+private
+
+  def weekly_counts(year: nil, week: nil)
+    counts = Counters::CountVisitsByPrisonAndCalendarWeek.
+             where(year: year, week: week).fetch_and_format
+
+    overdue = Overdue::CountOverdueVisitsByPrisonAndCalendarWeek.
+              where(year: year, week: week).ordered_counters
```
tatyree: If it's working for you, then great. ordered_counters was mainly meant to be a method to feed fetch_and_format, which in mixed from the various support modules when it isn't being declared explicitly.

alan: I wish I had known! I'll update after tomorrow

PR #213, opened 2016-03-15
Break down timely and overdue visits by prisions

tatyree:
Only for booked and rejected visits.  Use the overdue metric if you'd
like to see all status that are overdue.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160314160900) do
+ActiveRecord::Schema.define(version: 20160315143957) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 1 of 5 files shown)


db/schema.rb
```diff
@@ -172,39 +172,39 @@
   SQL
 
   create_view :distributions,  sql_definition: <<-SQL
-      SELECT percentile_disc((ARRAY[0.99, 0.95, 0.90, 0.75, 0.50, 0.25])::double precision[]) WITHIN GROUP (ORDER BY date_part('epoch'::text, (vsc.created_at - v.created_at))) AS percentiles
+      SELECT percentile_disc(ARRAY[(0.99)::double precision, (0.95)::double precision, (0.90)::double precision, (0.75)::double precision, (0.50)::double precision, (0.25)::double precision]) WITHIN GROUP (ORDER BY date_part('epoch'::text, (vsc.created_at - v.created_at))) AS percentiles
```
alan: Why did you have to add the `double precision` ?

tatyree: Oh, weird.  That must be a leftover from an old dev version of the DB that is still hanging around on this machine.   I'll fix it.

PR #215, opened 2016-03-16
CSV downlad the last 12 weeks of confirmed stats

alan:


spec/metrics/weekly_metrics_confirmed_csv_exporter_spec.rb
```diff
@@ -0,0 +1,38 @@
+require 'rails_helper'
+
+RSpec.describe WeeklyMetricsConfirmedCsvExporter do
+  let(:weeks) { 2 }
+  let(:instance) { described_class.new(weeks: weeks) }
+
+  describe '#to_csv' do
+    let(:prison1) { FactoryGirl.create(:prison, name: 'A Prison') }
```
tatyree: You don't need `FactoryGirl`, the build and create are already included.

PR #216, opened 2016-03-16
Brute-force the flip flopping sendgrid features

tatyree:
These are intermittent and impossible to reproduce. They are also wrong.
This just ensures that they get rerun when the occur.  Not intended as a
long term fix as these features will go when PVB2 becomes the API.


tatyree:
This was fixed using a workaround published on the poltergeist repo, instead.

PR #218, opened 2016-03-17
Try to fix the intermittently failing features

tatyree:
I'm not 100% certain about this, but it has *appeared* to fix a
semi-regularly failing seed of this test. The rationale is explained
here:

https://github.com/teampoltergeist/poltergeist#mouseeventfailed-errors


mloughran:
@tatyree Interesting – I don't pretend to understand the issue but will merge and we can see. Did you try taking a screenshot as suggested by the poltergeist page?

PR #231, opened 2016-03-21
Fix overdue count in metrics

tatyree:


tatyree:
This was opened so @alan could easily find the metrics feature spec.

PR #232, opened 2016-03-21
Catch Excon errors subclasses

alan:
Excon raises its own timeout class which means that it is getting reported to
Sentry

### DIFF
--- a/app/services/sendgrid_api.rb
+++ b/app/services/sendgrid_api.rb
@@ -112,14 +112,15 @@ private
     yield response
   rescue => e
     Rails.logger.error("#{e.class.name}: #{e.message}")
-    unless RESCUABLE_ERRORS.any? { |error_klass| e.is_a?(error_klass) }
-      Raven.capture_exception(e)
-    end
+    Raven.capture_exception(e) unless ignore_error?(e)
     false
   end
 
+  def ignore_error?(error)
+    RESCUABLE_ERRORS.any? { |error_klass| error.is_a?(error_klass) }
+  end
+
   def enabled?
-    Rails.logger.error('Sendgrid is disabled') unless @enabled
     @enabled
   end
 end
--- a/app/services/sendgrid_api.rb
+++ b/app/services/sendgrid_api.rb
@@ -105,21 +105,21 @@ private
   end
 
   def call_api(method, action, data)
-    unless enabled?
-      Rails.logger.error('Sendgrid is disabled')
-      return false
-    end
+    return false unless enabled?
 
     response = @pool.with { |client| client.request(method, action, data) }
 
     yield response
   rescue => e
     Rails.logger.error("#{e.class.name}: #{e.message}")
-    Raven.capture_exception(e) unless RESCUABLE_ERRORS.include?(e.class)
+    unless RESCUABLE_ERRORS.any? { |error_klass| e.is_a?(error_klass) }
+      Raven.capture_exception(e)
+    end
     false
   end
 
   def enabled?
+    Rails.logger.error('Sendgrid is disabled') unless @enabled
     @enabled
   end
 end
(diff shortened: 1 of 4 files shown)


spec/services/sendgrid_api/shared_examples.rb
```diff
@@ -18,6 +18,17 @@
       end
     end
 
+    context 'when times outs' do
```
tatyree: 'when timeouts occur' or something similar, please.

spec/services/sendgrid_api/shared_examples.rb
```diff
@@ -18,6 +18,17 @@
...
+      include_context 'sendgrid times out'
```
tatyree: 'timeout'

tatyree:
Sorry, @alan, niggly little grammatical things, s'all, otherwise :+1:

app/services/sendgrid_api.rb
```diff
@@ -105,21 +105,21 @@ def email_found?(response)
...
     Rails.logger.error("#{e.class.name}: #{e.message}")
-    Raven.capture_exception(e) unless RESCUABLE_ERRORS.include?(e.class)
+    unless RESCUABLE_ERRORS.any? { |error_klass| e.is_a?(error_klass) }
+      Raven.capture_exception(e)
+    end
     false
   end
 
   def enabled?
+    Rails.logger.error('Sendgrid is disabled') unless @enabled
```
mloughran: Is this really appropriate for ERROR loglevel?

mloughran: Also, not sure why you moved this out of the call api method? It makes more sense there doesn't it? You calling the API and want to warn that sendgrid is disabled?

mloughran:
@alan left a few comments on the commit

app/services/sendgrid_api.rb
```diff
@@ -105,21 +105,21 @@ def email_found?(response)
...
-    end
+    return false unless enabled?
 
     response = @pool.with { |client| client.request(method, action, data) }
 
     yield response
   rescue => e
     Rails.logger.error("#{e.class.name}: #{e.message}")
-    Raven.capture_exception(e) unless RESCUABLE_ERRORS.include?(e.class)
+    unless RESCUABLE_ERRORS.any? { |error_klass| e.is_a?(error_klass) }
```
mloughran: Is there a clearer way of expressing this line?

alan: I could pull it out to a private method? Any suggestions?

app/services/sendgrid_api.rb
alan: Rubocop alerted of the method size, then I thought that having in `enabled?` could be good because any use of `enabled?` would be worth logging as it indicates that someone wants to use Sendgrid in some way

alan: Looking at it again I would just log this once in the initializer, I don't think there is much value logging every time there is an attempted api call since it will remain enabled / disabled until the app boots again. What do you think?

spec/services/sendgrid_api/shared_examples.rb
alan: I'll update it

alan:
To be reviewed on the 29th of March if we want to silence Sendgrid errors. That way we'll have a weeks worth of data to make a decision

alan:
Today is the 29th so time to review the Sendgrid errors. Sendgrid errors are frequent,  we got over 300 instances.

I say we merge this PR so that we can resolve the sentry and create a ticket to implement Sendgrid metrics. Thoughts @tatyree @mloughran ?

tatyree:
Merging this seems sensible given that error rate.  Any further thoughts on the noted/requested changes, @alan?  @mloughran?

alan:
I'll update the grammatical mistakes when updating the other issues.

Need to get agreement on the logging for sendgrid being disables. I'd like to move it to the initializer, don't see much value of logging on every call. I guess it is worth noting that it behaves like this currently because of PVB1.

Open to suggestions for the code to detect error subclasses

tatyree:
I'd be fine with it being a private method with a declarative name, @alan.

PR #236, opened 2016-03-23
Fix prisoner number validation errors

tatyree:
Users are occassionally adding trailing spaces to otherwise valid
numbers, this causes the prisoner step to fail. I've added a pre
validation hook to the model to strip leading and trailing spaces from
the prisoner number to prevent this from happening.

### DIFF
--- a/app/models/prisoner_step.rb
+++ b/app/models/prisoner_step.rb
@@ -3,12 +3,15 @@ require 'maybe_date'
 class PrisonerStep
   include NonPersistedModel
   include Person
+  include ActiveModel::Validations::Callbacks
 
   attribute :first_name, String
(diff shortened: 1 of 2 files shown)


mloughran:
This is inconsistent with the way @alan implemented email stripping (https://github.com/ministryofjustice/prison-visits-2/blob/master/app/models/visitors_step.rb#L30-L32). I think we should be consistent.

alan:
I was thinking about this too.

I wonder if we'd be better off implementing a more generic solution like PVB1 does stripping all whitespace from any parameter that can be stripped. I can't think of any parameter for pvb that we'd want to keep leading / trailing whitespace

PR #237, opened 2016-03-23
Fix incorrect calculations in the weekly specs.

tatyree:
Ensure weekly rejection percentages are correct

Previously, the SQL used for the calculations would report the same
percentage for every type of rejection including the total. This was
because the date selection logic was in the wrong place.

This also removes the class and the specs for rejections done by
calendar date.  This was not being used anywhere and required additional
tuning to avoid the same problems.  If it is required, we can revisit
the SQL and migrations at a later date.

### DIFF
--- a/db/views/rejection_percentage_by_prison_and_calendar_weeks_v02.sql
+++ b/db/views/rejection_percentage_by_prison_and_calendar_weeks_v02.sql
@@ -1,0 +1,64 @@
+SELECT rejected.prison_name,
+       'total' AS reason,
+       rejected.year AS year,
+       rejected.week AS week,
+       ROUND((rejected_count::numeric / booked_count * 100),2) AS percentage
+FROM (SELECT prisons.name AS prison_name,
+             count(*) AS rejected_count,
+             extract(isoyear from visits.created_at)::integer AS year,
+             extract(week from visits.created_at)::integer AS week
+      FROM visits
+      INNER JOIN prisons ON prisons.id = visits.prison_id
+      WHERE visits.processing_state = 'rejected'
+      GROUP BY prison_name,
+             prison_id,
+               year,
+               week
+      ) rejected,
+     (SELECT prisons.name AS prison_name,
+             count(*) AS booked_count
+      FROM visits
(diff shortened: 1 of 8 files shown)


db/views/rejection_percentage_by_prison_and_calendar_weeks_v02.sql
```diff
@@ -0,0 +1,65 @@
...
+       rejected.year AS year,
+       rejected.week AS week,
+       ROUND((rejected_count::numeric / booked_count * 100),2) AS percentage
+FROM (SELECT prisons.name AS prison_name,
+             count(*) AS rejected_count,
+             extract(isoyear from visits.created_at)::integer AS year,
+             extract(week from visits.created_at)::integer AS week
+      FROM visits
+      INNER JOIN prisons ON prisons.id = visits.prison_id
+      INNER JOIN rejections ON rejections.visit_id = visits.id
```
alan: Do you need to join on rejections? As far as I can see visits and prisons should be enough?

db/views/rejection_percentage_by_prison_and_calendar_weeks_v02.sql
```diff
@@ -0,0 +1,65 @@
+SELECT rejected.prison_name,
```
alan: Have you tried using CTE's? Would help to break the query up to make them more readable

tatyree: I have now.  Unfortunately, adapting the existing queries simply causes the same bug to reappear.  I made some progress with them, but ultimately wasn't getting the desired results in a timely fashion.  This works for now, so I will leave it as is and revisit later.

db/views/rejection_percentage_by_prison_and_calendar_weeks_v02.sql
tatyree: It is required here as `rejections` is the only source of the rejection reason.  It isn't required in the similar subquery, above, and so I've removed it.

PR #240, opened 2016-03-24
'/status/:id' to catch pvb1 email url

alan:
Implements the pvb1 path that it is still being used. Users got a link on their
email in pvb1 that took them to a page with information about a visit.

It renders a page similar to the not found page but with a message that is more
helpful than just page not found.

Ommited the Welsh translation since it shoud always default to the english
locale.

### DIFF
--- a/app/views/pvb1_paths/status.html.erb
+++ b/app/views/pvb1_paths/status.html.erb
@@ -1,0 +1,11 @@
+<% content_for :header, t('.title') %>
+
+<div class='Grid'>
+  <div class='Grid-2-3'>
+    <article>
+      <%= markdown t('.body_md', {
+            new_booking_url: booking_requests_path,
+            prison_finder_url: link_directory.prison_finder}) %>
+    </article>
+  </div>
+</div>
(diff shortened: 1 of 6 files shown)


app/views/pvb1_paths/status.html.erb
```diff
@@ -0,0 +1,14 @@
+<% content_for :header, 'Visit expired' %>
+
+<div class='Grid'>
+  <div class='Grid-2-3'>
+    <article>
+      <p>This visit has expired and its information has been removed.<p>
```
tatyree: The copy in this view needs to go into en.yml.

kyliemulholland:
@alan Looks good to me

app/views/pvb1_paths/status.html.erb
alan: @tatyree 
This a pvb1 link, so has no locale param in the url. It would default to english if we used localisation. Given that the page is not going to be translated I decided to put the copy on the view to make it explicit that this copy is only in english.

I originally had changed `en.yml` and `cy.yml` so happy to move it to `en.yml` if it makes more sense.

tatyree: It's nor so much about have a localisation as it is that all of our other copy is in en.yml.

alan: Ok no problem, I'll move it then

PR #242, opened 2016-03-29
Change enabled to a scope and eager load estate

tatyree:
I noticed an n+1 query in the metrics when Prison was loading Estate.
The class method `enabled` seemed better expressed as a scope and I
added eager loading of Estate.

### DIFF
--- a/app/models/prison.rb
+++ b/app/models/prison.rb
@@ -18,9 +18,9 @@ class Prison < ActiveRecord::Base
     to: :parsed_slot_details
   delegate :finder_slug, to: :estate
 
-  def self.enabled
+  scope :enabled, lambda {
     where(enabled: true).order(name: :asc)
-  end
+  }
 
   def available_slots(today = Time.zone.today)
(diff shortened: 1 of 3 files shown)


app/models/prison.rb
```diff
@@ -18,9 +18,9 @@ class Prison < ActiveRecord::Base
     to: :parsed_slot_details
   delegate :finder_slug, to: :estate
 
-  def self.enabled
-    where(enabled: true).order(name: :asc)
-  end
+  scope :enabled, lambda {
+    where(enabled: true).order(name: :asc).includes(:estate)
```
alan: Do we always want to eager load estate? Is kind of surprising to see eager loading on a scope that checks a boolean flag.

My preference is to keep scopes simple so that they can easily reused, for example, if we wanted to pluck the id's of enabled prisons we wouldn't be forced to eager load estates

tatyree: A fair point.  I'll move the `includes` to somewhere more appropriate.

PR #243, opened 2016-03-29
Display 95th and median end to end processing times

tatyree:
The method called on the view model had changed to `fetch_and_format`,
which returns a hash. However, the presenter was still expecting an
array, which was what `ordered_counters` returned.  As a result, no
values were being found.

### DIFF
--- a/app/metrics/prison_summary_metrics_presenter.rb
+++ b/app/metrics/prison_summary_metrics_presenter.rb
@@ -1,12 +1,4 @@
 class PrisonSummaryMetricsPresenter
-  PERCENTILES_INDEX = {
-    '99th' => 0,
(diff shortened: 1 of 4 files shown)


app/metrics/prison_summary_metrics_presenter.rb
```diff
@@ -31,8 +23,8 @@ def processed_overdue
   def end_to_end_percentile(percentile)
     return 0 unless @percentiles
 
-    index = PERCENTILES_INDEX.fetch(percentile)
-    seconds = @percentiles[index]
+    key = percentile[/\d+/].to_i
```
alan: Nice, I hadn't seen this syntax for regexes before

PR #246, opened 2016-03-31
Intersect returned slots with prisoner availability from Nomis

mloughran:
This only applies when slots are requested via the slots API, and is therefore only visible from pvb-public.

### DIFF
--- a/app/controllers/api/slots_controller.rb
+++ b/app/controllers/api/slots_controller.rb
@@ -1,8 +1,15 @@
 module Api
   class SlotsController < ApiController
     def index
-      prison = Prison.enabled.find(params[:prison_id])
-      @slots = prison.available_slots
+      prison = Prison.enabled.find(params.fetch(:prison_id))
+
+      slot_availability = SlotAvailability.new(prison: prison)
+      slot_availability.restrict_by_prisoner(
+        prisoner_number: params.fetch(:prisoner_number),
+        prisoner_dob: Date.parse(params.fetch(:prisoner_dob))
+      )
+
+      @slots = slot_availability.slots
     end
   end
 end
--- a/app/models/slot_availability.rb
+++ b/app/models/slot_availability.rb
@@ -7,6 +7,7 @@ class SlotAvailability
   end
 
   # rubocop:disable Metrics/MethodLength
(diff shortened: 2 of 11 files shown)


app/models/slot_availability.rb
```diff
@@ -0,0 +1,36 @@
...
+      noms_id: prisoner_number,
+      date_of_birth: prisoner_dob
+    )
+
+    availability = Nomis::Api.instance.offender_visiting_availability(
+      offender_id: offender.id,
+      start_date: @prison.first_bookable_date,
+      end_date: @prison.last_bookable_date
+    )
+    offender_available_dates = availability.dates
```
tatyree: This feels convoluted.  Wouldn't offender and availability be better broken out into private methods?

app/controllers/api/slots_controller.rb
```diff
@@ -1,8 +1,15 @@
...
     def index
-      prison = Prison.enabled.find(params[:prison_id])
-      @slots = prison.available_slots
+      prison = Prison.enabled.find(params.fetch(:prison_id))
+
+      slot_availability = SlotAvailability.new(prison: prison)
+      slot_availability.restrict_by_prisoner(
+        prisoner_number: params.fetch(:prisoner_number),
+        prisoner_dob: Date.parse(params.fetch(:prisoner_dob))
+      )
```
tatyree: Am I missing something?  I feel it would be cleaner if the prisoner info were passed into the initializer and the `slots` method automatically restricted the slots based on that.

alan:
:+1:

app/models/slot_availability.rb
mloughran: Personally, I think the code would be less readable if there was more indirection. It's just 2 API calls.

app/controllers/api/slots_controller.rb
mloughran: The reason for this design is that we'll have more restrictions to apply as time continues. Adding all the arguments to the initialiser would make the code less flexible and increasingly convoluted.

A potential alternative design I discussed with @alan would be to make this method a functional filter, returning self with a filtered list of dates, rather than mutating slots. We decided that we'd revisit when we implement the next restriction.

PR #247, opened 2016-04-01
Proxy staff info pages from S3

tatyree:
The old app used a gem with an engine to include these pages. It was out
of date, and made updating content problematic. This proxies a static
website published to S3 in html, markdown, pdf and pngs/jpegs
directly.  

The content is published from the `prison_staff_info` private repo on
github.  Content changes should be made there.  Style change should be
made here.

### DIFF
--- a/app/controllers/staff_info_controller.rb
+++ b/app/controllers/staff_info_controller.rb
@@ -1,0 +1,87 @@
+class StaffInfoController < ApplicationController
+  rescue_from Excon::Errors::Error, with: :handle_excon_error
+  # Raised when Excon tries to parse the URI.  Will be raised if there is
+  # an error or ENV['STAFF_INFO_ENDPOINT'] is not set.
+  rescue_from URI::InvalidURIError, with: :not_available
+
+  TIMEOUT = 2
+
+  before_action :authorize_prison_request
+
+  def index
+    @content = fetch('/').body
+  end
+
+  # rubocop:disable Metrics/AbcSize
+  def show
+    page_path = params.values_at(:page, :format).compact.join('.')
+    resp = fetch(page_path)
+
+    if pdf?(resp)
+      send_data(resp.body, filename: page_path, type: 'application/pdf')
+    else
+      @nav = markdown(fetch('nav.md').body)
+      @content = markdown?(resp) ? markdown(resp.body) : resp.body
+    end
+  end
+# rubocop:enable Metrics/AbcSize
+
+private
+
+  def handle_excon_error(exception)
+    Rails.logger.
+      warn "Staff info pages cannot be served because of #{exception}."
+
+    if exception.class == Excon::Errors::NotFound
+      not_found
+    else
+      not_available
+    end
+  end
+
+  def not_found
+    fail ActionController::RoutingError, 'Not Found'
+  end
+
+  def not_available
+    render :not_available
+  end
+
+  def markdown?(resp)
+    resp.headers['Content-Type'] == 'text/markdown'
+  end
+
+  def pdf?(resp)
+    resp.headers['Content-Type'] == 'application/pdf'
+  end
+
+  def fetch(page_path)
+    options = {
+      expects: [200],
+      method: :get,
+      path: page_path,
+      read_timeout: TIMEOUT
+    }
+    connection.request(options)
+  end
+
+  def connection
+    @connection ||= Excon.new(
+      Rails.configuration.staff_info_endpoint,
+      persistent: true,
+      connect_timeout: TIMEOUT)
+  end
+
+  def markdown(content)
+    # Several of the existing pages have bad ASCII encoding. While we
+    # will fix these, it is highly likely others will appear in time as
+    # less experienced staff edit them using obsolete versions of Word.
+    unless content.encoding == Encoding::UTF_8
+      content.force_encoding('UTF-8')
+    end
+    content.scrub! unless content.valid_encoding?
+
+    Redcarpet::Markdown.new(Redcarpet::Render::HTML).
+      render(content)
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -17,6 +17,7 @@ gem 'netaddr'
 gem 'pg'
 gem 'premailer-rails'
 gem 'puma'
+gem 'redcarpet'
 gem 'sass-rails', '~> 5.0'
 gem 'scenic', '>= 1.2.0'
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -197,6 +197,7 @@ GEM
       thor (>= 0.18.1, < 2.0)
     rainbow (2.0.0)
     rake (10.5.0)
+    redcarpet (3.3.4)
     redis (3.2.2)
     rspec (3.4.0)
       rspec-core (~> 3.4.0)
@@ -355,6 +356,7 @@ DEPENDENCIES
   puma
   rails (~> 4.2.3)
   rake (< 11.0)
+  redcarpet
   rspec-rails (~> 3.0)
   rubocop
   rubocop-rspec
--- a/README.md
+++ b/README.md
@@ -259,6 +259,10 @@ These configure email delivery in the production environment. `SMTP_DOMAIN` is
 also used when generating the `no-reply@` address and the `feedback@` stand-in
 address used when submitting feedback without an email address to Zendesk.
 
+#### `STAFF_INFO_ENDPOINT`
+
+This is used to proxy the staff information pages static site.
+
 #### `OVERRIDE_PRISON_EMAIL_DOMAIN`
 
 In order to avoid sending confusing and irrelevant messages to real prisons and
--- a/app/views/staff_info/index.html.erb
+++ b/app/views/staff_info/index.html.erb
@@ -1,0 +1,1 @@
+<%= raw(@content) %>
--- a/app/views/staff_info/not_available.html.erb
+++ b/app/views/staff_info/not_available.html.erb
@@ -1,0 +1,1 @@
+<h1><%= t('.staff_info_service_unavailable') %></h1>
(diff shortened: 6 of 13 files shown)


tatyree:
As per the comment, I am not concerned about the AWS S3 endpoint; the information is not sensitive.  If it is an issue, we can update the policy to restrict it to our servers and the prison estates.

alan:
Changes look good.

Couple of questions:
- These pages won't be accessible in production until we switch off pvb1 right?
- What happens if there is a deadlink? I guess best thing would be render a 404?

mloughran:
Generally looks good.

However personally I would rather we do not include the S3 URL in this repo, nor use real content in the VCR tests. I don't know if it's sensitive, but since it's not open source, I'd rather simply not worry about it, and not bother with IP restricting.

R.e. the VCR tests fixture files, I'd rather keep them smaller if possible, so bit concerned to see the PDF in there. Is using VCR perhaps overkill for this? It's not like we're testing an API – it's simply a standard HTTP get. Maybe webmock would be simpler, smaller, and more revealing?

tatyree:
The info isn't secret at all, but I take your point.  VCR was just must simpler, but point taken on the pdf.

app/controllers/staff_info_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+  before_action :proxy_content, only: :show
+
+  def index
+    @content = proxy_content
+    render :index, status: req.status
+  end
+
+  def show
+    @content = markdown? ? convert_markdown : proxy_content
+    render :show, status: req.status
```
alan: If the page is not found in S3 it will show a blank page, should we be catching error codes differently to render our own error pages? I think as a user it would be surprising to see a page with no content.

mloughran: I agree, I don't think we should be forwarding status codes from S3. Unless it's a 200, it should go down an error path.

tatyree: It does not show a blank page.  It shows a 404, with the correct status: 

screen shot 2016-04-04 at 11 07 03

alan: Not as bad as I thought then.

I still think that would be nicer to show our content for errors rather than S3, up to you if you want to implement it, my main concern is gone now

tatyree: I decided to in the end as I also wanted to proxy the PDFs rather than redirect to them.  Thought I'd do them both at the same time.

app/controllers/staff_info_controller.rb
```diff
@@ -0,0 +1,80 @@
+class StaffInfoController < ApplicationController
+  before_action :authorize_prison_request
+
+  STAFF_INFO_ENDPOINT = ENV.fetch('STAFF_INFO_ENDPOINT', nil)
```
mloughran: This is inconsistent with the way configuration is fetched from ENV elsewhere. Can it be in `config/application.rb`?

db/schema.rb
```diff
@@ -452,7 +452,7 @@
       (date_part('isoyear'::text, v.created_at))::integer AS year
      FROM ((visits v
        JOIN prisons ON ((prisons.id = v.prison_id)))
-       JOIN visit_state_changes vsc ON (((v.id = vsc.visit_id) AND ((vsc.visit_state)::text = ANY ((ARRAY['booked'::character varying, 'rejected'::character varying])::text[])))))
+       JOIN visit_state_changes vsc ON (((v.id = vsc.visit_id) AND ((vsc.visit_state)::text = ANY (ARRAY[('booked'::character varying)::text, ('rejected'::character varying)::text])))))
```
mloughran: Looks like this shouldn't be here...

spec/spec_helper.rb
```diff
@@ -65,6 +65,6 @@
 
   config.ignore_request do |request|
     # Ignore capybara requests within feature tests
-    request.uri =~ /__identify__/
+    request.uri =~ /(__identify__|example\.com)/
```
mloughran: Why are we ignoring requests to example.com? It this related to the trouble you were having with Webmock @tatyree ?

app/controllers/staff_info_controller.rb
tatyree: Yep.

db/schema.rb
tatyree: Ah, that old chestnut.  God I love the schema.  Will fix.

app/controllers/staff_info_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+    render :index, status: req.status
+  end
+
+  def show
+    @content = markdown? ? convert_markdown : proxy_content
+    case req.status
+    when 200 then render :show
+    when 404 then render 'errors/404', status: 404
+    when 500 then render 'errors/500', status: 500
+    when 503 then render 'errors/503', status: 503
```
mloughran: I don't particularly see the purpose of proxying these statuses 1-1, and this code will render nothing in the case that the status is not in the list.

I would suggest that a non-200 error code simply raises an exception which is handled by the errors controller. I'd be happy with handling 5xx separately and returning an error, but I'm not fussed.

app/controllers/staff_info_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+  end
+
+private
+
+  def endpoint_set?
+    STAFF_INFO_ENDPOINT.present?
+  end
+
+  def not_found
+    render 'errors/404', status: 404
```
mloughran: Again, we shouldn't be rendering error pages in multiple places. I'd prefer `fail ActionController::RoutingError, 'Not Found'` which is caught be the router, and keep error rendering in one place.

app/controllers/staff_info_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+  def remote_endpoint(endpoint = page_params)
+    "#{STAFF_INFO_ENDPOINT}#{endpoint}"
+  end
+
+  def page_params
+    params.permit(:page, :format).values.compact.join('.')
+  end
+
+  def req
+    @req ||= Excon.get(remote_endpoint)
```
mloughran: This is a niggle, but it does mean we're not using a persistent TCP connection for S3 connections. It doesn't matter given the traffic volume, but it would have been better to create a long-lived connection like all the other code using excon.

app/controllers/staff_info_controller.rb
```diff
@@ -0,0 +1,80 @@
+class StaffInfoController < ApplicationController
+  before_action :authorize_prison_request
+
+  STAFF_INFO_ENDPOINT = ENV.fetch('STAFF_INFO_ENDPOINT', nil)
+
+  before_action :not_found, unless: :endpoint_set?
+  before_action :send_pdf, if: :pdf?
+  before_action :fetch_nav, only: :show
+  before_action :proxy_content, only: :show
```
mloughran: I don't like the fact that these are all before actions – it makes the code needlessly hard to follow in the action.

spec/spec_helper.rb
tatyree: No, because VCR hooks before webmock.  Ah, I'd forgotten you can turn it off...I'll remove and reimplement as an around block.

app/controllers/staff_info_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+    STAFF_INFO_ENDPOINT.present?
+  end
+
+  def not_found
+    render 'errors/404', status: 404
+  end
+
+  def fetch_nav
+    endpoint = "#{STAFF_INFO_ENDPOINT}nav.md"
+    @nav = markdown.render(Excon.get(endpoint).body.scrub!)
```
mloughran: I don't like that the http requesting code is included in multiple place.

Why are you calling `scrub!` on the result?

app/controllers/staff_info_controller.rb
tatyree: Ok

app/controllers/staff_info_controller.rb
tatyree: Ok

app/controllers/staff_info_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+    send_data(proxy_content, filename: page_params, type: 'application/pdf')
+  end
+
+  def pdf?
+    req.headers['Content-Type'] == 'application/pdf'
+  end
+
+  def convert_markdown
+    unless proxy_content.encoding == Encoding::UTF_8
+      proxy_content.force_encoding('UTF-8')
```
mloughran: Is this just defensive coding, or do we actually expect non-UTF8 content? I'd almost be tempted to treat this as an error case rather than trying to handle it.

app/controllers/staff_info_controller.rb
mloughran: In fact this is an issue even within 1 request – since a separate TCP conneection will be used to fetch the body and the nav.

tatyree: I considered it, but felt it was unnecessary optimisation considering the volume is around 100 hits per month.  This is simpler and will easily meet the need.

app/controllers/staff_info_controller.rb
tatyree: There is non-UTF8 encoding in at least one of the existing pages.

spec/spec_helper.rb
mloughran: Actually, I believe VCR uses webmock to intercept requests.

app/controllers/staff_info_controller.rb
mloughran: Can we fix that instead? Forcing encoding is likely to result in broken special characters.

config/locales/en/views.yml
```diff
@@ -338,3 +338,7 @@ en:
       title: You cancelled this visit
     withdrawn:
       title: You cancelled this visit request
+  staff_info:
+    not_available:
+      staff_info_service_unavailable: >-
+        Staff info service is temporarily unavailable. Please check back later.
```
mloughran: @tatyree this should be in `prison_views.yml` – so that it doesn't get requested as a translation.

PR #248, opened 2016-04-04
Sendgrid webhook to bounce incoming emails

alan:
Sendgrid does not provide mailboxes, all emails are accepted and is up to the
integrators to do something via the webhooks notifications.

This means that emails sent to the noreply email address don't bounce off
automatically, and need to be handled with a webhook so that an email can be
sent to the sender letting them know that the email address is unattended.

### DIFF
--- a/app/services/parsed_email.rb
+++ b/app/services/parsed_email.rb
@@ -1,0 +1,41 @@
+class ParsedEmail
+  include NonPersistedModel
+
+  ParseError = Class.new(StandardError)
+
+  attribute :from, String
+  attribute :to, String
+  attribute :subject, String
+  attribute :text, String
+
+  # rubocop:disable Metrics/AbcSize
+  def self.parse(hash)
+    hash = hash.dup
+    fail ParseError, 'Missing subject' unless hash[:subject]
+    fail ParseError, 'Missing email body' unless hash[:text]
+    charsets = JSON.parse(hash[:charsets]).with_indifferent_access
+
+    [:subject, :text].each do |field|
+      # Fields can have different encodings, normalize everything to UTF-8
+      encoding = charsets[field]
+
+      hash[field].force_encoding(encoding).encode!('UTF-8') if encoding
+    end
+
+    new new_parse(hash)
+  end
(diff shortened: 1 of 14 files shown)


alan:
If we decide to merge this the Sendgrid webhook and environment variable will need to be reinstated.

app/services/parsed_email.rb
```diff
@@ -0,0 +1,41 @@
+class ParsedEmail
```
alan: Ported from pvb1, mostly rubocop changes

app/services/parsed_email.rb
```diff
@@ -0,0 +1,41 @@
...
+    {
+      to: Mail::Address.new(hash[:to]),
+      from: Mail::Address.new(hash[:from]),
+      subject: hash[:subject],
+      text: hash[:text]
+    }
+  end
+
+  def source
+    from.domain.in?(%w[hmps.gsi.gov.uk noms.gsi.gov.uk]) ? :prison : :visitor
```
alan: Added 'noms' checking

config/routes.rb
```diff
@@ -20,6 +20,8 @@
   scope '/:locale', locale: /[a-z]{2}/ do
     get '/', to: redirect('/%{locale}/request')
 
+    post 'webhooks/email/:auth', to: 'webhooks#email'
```
alan: To be moved out of /en prefix once pvb1 is switched off

app/services/parsed_email.rb
tatyree: Will there be a need for additional domains?  Should these be in a constant?

app/services/parsed_email.rb
tatyree: It feels like a service to me.

app/services/parsed_email.rb
alan: Good question, I'm not sure, I noticed noms subdomain by chance recently by looking at the email addresses that prisons have. @kyliemulholland do you know if there is or will be anymore?

PR #249, opened 2016-04-04
Feedback submission API 'create' endpoint

alan:
The change in behaviour from the pvb2 html endpoint is that it returns a 422 if
the feedback object is invalid.


app/controllers/api/feedback_submissions_controller.rb
```diff
@@ -0,0 +1,21 @@
+module Api
+  class FeedbackSubmissionsController < ApiController
+    def create
+      @feedback = FeedbackSubmission.new(feedback_params)
+
+      if @feedback.save
+        ZendeskTicketsJob.perform_later(@feedback)
+      else
+        render status: 422, json: { messages: @feedback.errors.full_messages }
```
alan: Note that this renders `messages` instead of `message` like the visits API does. I think an array works better instead of rendering error messages individually. We should choose one or the other and keep it consistent.

Or we could consider using [ActiveModel::Serializer](https://github.com/rails-api/active_model_serializers) for serialisation. They were part of the `rails-api` project that eventually got merged into Rails

I had a good experience using it in the past, makes it easier to be consistent when writing an API. Decisions like how to render error are all handled by the serialiser which depends on what adapter you are using in the latest version ie json-api, custom one, etc. If we don't want to follow a particular standard we could then have a generic moj adapter to have consistent API's. What do you think?

app/views/api/feedback_submissions/create.json.jbuilder
```diff
@@ -0,0 +1,6 @@
+json.feedback_submission do
+  json.body @feedback.body
```
alan: Wasn't sure what to render for the success case, in theory a 201 response code should be all is needed, but returning some content seemed to me more consistent. Left the id out since there is no show action

app/controllers/api/feedback_submissions_controller.rb
```diff
@@ -0,0 +1,21 @@
+module Api
+  class FeedbackSubmissionsController < ApiController
```
alan: Maybe we should just call it `Feedback` ?

tatyree: :+1:

app/controllers/api/feedback_submissions_controller.rb
tatyree: :+1:

alan: @tatyree On using 'messages' or ActiveModel::Serializers ?

tatyree: Both, @alan

mloughran: @alan I've been loosely following the aspects I like from Github's API, in which there is a message and then an array of errors. See https://developer.github.com/v3/#client-errors.

So in this case the error would be something like "Validation Failed" and then there would be a list of errors.

_However_, I would prefer to do validation on the client side (i.e. in public) rather than using the API for this. This is the pattern for the booking process, in which case creating a booking is expected to succeed and failure is an error state. I see no reason to do differently for the API.

app/views/api/feedback_submissions/create.json.jbuilder
mloughran: @alan I'd prefer not to return anything from the API unless something is required. In terms of status code, I understand that technically 201 might be correct (although so might 202 since although the feedback has been created, it is asynchronously sent to zendesk). However, I've simply used 200 for other successful calls (including creating a booking), and unless there's a reason to differ I'd rather not.

app/controllers/api/feedback_submissions_controller.rb
mloughran: +1

app/controllers/api/feedback_submissions_controller.rb
mloughran: I haven't used `ActiveModel::Serializer` – in principle it looks good, but since this API is currently extremely simple I don't think it's necessary to add right now.

config/routes.rb
```diff
@@ -46,6 +46,7 @@
 
   namespace :api, constraints: { format: 'json' } do
     get '/', to: 'root#index'
+    resources :feedback_submissions, only: :create
```
mloughran: Purely for code consistency, `%i[ create ]`?

app/controllers/api/feedback_submissions_controller.rb
alan: What do you mean by 'error state' in your second last sentence? Do you mean something different than a 422?

I don't see it as a bad thing to return validations errors on create, mainly for 2 reasons:
- Some errors will happen on creation even after pre-validating, for example unique validations that are prone to race conditions.
- Instead of having 2 requests for creation (validation + create) we only need 1 request (If I understand your suggested pattern correctly). Less requests are better imo.

Good to know that you are following some form the Github's API, I didn't really know what to do with the errors. What do you think of trying to encapsulate this design decisions on a ActiveModel Serializer adapter, or writing them somewhere, perhaps on a wiki page on the project?

Out of curiosity, did you consider some API standards like json-api?

app/views/api/feedback_submissions/create.json.jbuilder
alan: OK, will update, I prefer not returning anything also.

app/controllers/api/feedback_submissions_controller.rb
mloughran: @alan I'm not suggesting creating 2 calls (using a separate one for validation) in the general case. I'm suggesting that the client does the validation wherever possible. I agree that if a validation error prevents saving an error should be returned, but I'm suggesting that that response doesn't need to be sufficiently rich to render an error page, and should be allowed to trigger a 500 response in pvb-public (that's what I mean by 'error state').

The validation APIs that do exist are special cases since they need to talk to NOMIS, and would be impossible to implement in the client.

I didn't consider using json-api. First impressions – pretty sensible, but potentially a little heavy (although a lot of the spec is optional). Happy to discuss!

alan: I realised afterwards that this is probably what you meant. I still think that is better to have one source for the validations where possible, that way we don't have to update more than 1 codebase when changing validations. Also 500 in my view are errors in the server side, a client shouldn't be able to trigger them. Lets go over some of the code I have when you are in.

Regarding json-api, my main worry is APIs consistency long term. I agree with you that initially might seem heavier than needed, however I think it can be beneficial long term:
- Have consistent API's across the the different micro services, which will make it easier to  build new services, understand the API of other services, or making changes to them.
- As this API evolves (or others), it might require design breaking changes. From experience it is more costly to introduce breaking changes later on. By using a standard we can piggy back on design decisions that have had a lot of thought on them.
- Code reuse as there is good open source support across different languages and frameworks.

Let's discuss when you are in.

alan: Thought more about after we talked, I agree with you on duplicating the validation. I prefer the explicit approach instead of relying on the API spec to proxy the errors, less magic.

I still would like to keep discussing how we are going to ensure API consistency as we start having more for new services

app/controllers/api/api_controller.rb
```diff
@@ -17,6 +17,10 @@ class ApiController < ActionController::Base
       render_error 404, 'Not found'
     end
 
+    rescue_from ActiveRecord::RecordInvalid do |e|
+      render_error 422, e.record.errors.full_messages
```
mloughran: @alan full messages will be an array, where elsewhere message is assumed to be a string. We're not using this programatically at this stage, so what do you think about keeping the types consistent but keeping it simple and simply returning `full_messages.join(', ')`?

alan: I just got notified about this message, 5 hours later. Good point, better to have them consistent.

PR #250, opened 2016-04-05
Show cancelled and withdrawn in overdue

tatyree:
Without these the total overdue counts do not add up.


db/schema.rb
```diff
@@ -503,4 +455,52 @@
...
   SQL
 
+  create_view :timely_and_overdues,  sql_definition: <<-SQL
+      SELECT count(*) AS count,
+      'overdue'::text AS status,
+      vsc.visit_state,
+      prisons.name AS prison_name
+     FROM ((visits v
+       JOIN prisons ON ((prisons.id = v.prison_id)))
+       JOIN visit_state_changes vsc ON (((v.id = vsc.visit_id) AND ((vsc.visit_state)::text = ANY ((ARRAY['booked'::character varying, 'rejected'::character varying, 'cancelled'::character varying, 'withdrawn'::character varying])::text[])))))
```
alan: I wonder if blacklisting might be better here in case we add a new state. I think we want to exclude the 'requested' state so might be safer to use that in the view?

PR #255, opened 2016-04-06
Update Excon to latest version

alan:
Bumping Excon to latest version so that pvb-public and pvb2 use the same version


alan:
https://github.com/ministryofjustice/prison-visits-public/pull/14

PR #260, opened 2016-04-07
Fix rejection calculations

tatyree:
The count of total visits included the rejected visits.  This was throwing the calculations off.  Compounding the problem was the fact that I had chosen odd numbers of visits when I was setting up the specs, so the error was not immediately obvious.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160406102326) do
+ActiveRecord::Schema.define(version: 20160407142305) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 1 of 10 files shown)


db/schema.rb
```diff
@@ -409,6 +346,54 @@
     GROUP BY rejected.prison_name, rejected.reason, round((((rejected.rejected_count)::numeric / (booked.booked_count)::numeric) * (100)::numeric), 2), (date_part('year'::text, rejected.created_at))::integer, (date_part('month'::text, rejected.created_at))::integer, (date_part('day'::text, rejected.created_at))::integer;
   SQL
 
+  create_view :timely_and_overdue_by_calendar_weeks,  sql_definition: <<-SQL
```
tatyree: I wish it would keep the same order.  I manually diffed it to fix some PG optimisations that only seem to happen on my machine around the distribution views.  I'm afraid that trying to get the move of this `create_view` diffed out manually is a bit too high-risk for my liking.

PR #262, opened 2016-04-08
Integrate nomis slot availability

mloughran:
This extends the existing get slots API to optionally make an additional API request into NOMIS to fetch slots, rather than using the hardcoded slots from this repo's YAML files.

In the case that NOMIS is not available or errors, the built in slots are still used.


spec/services/nomis/api_spec.rb
```diff
@@ -43,22 +43,43 @@
     let(:params) {
       {
         offender_id: 1_055_827,
-        start_date: Date.parse('2016-04-01'),
-        end_date: Date.parse('2016-04-21')
+        start_date: Date.parse('2016-05-01'),
+        end_date: Date.parse('2016-05-21')
```
tatyree: The date changes are down to re-recording the cassette?

tatyree:
LGTM.  @alan?

spec/services/nomis/api_spec.rb
mloughran: Yes, which we can't avoid because the NOMIS API is a black box from the perspective of these tests. It shouldn't be an issue because we won't need to re-record often.

tatyree: :+1:  As I assumed.  Just wanted to be sure.

app/models/slot_availability.rb
```diff
@@ -1,12 +1,32 @@
...
 
-  def initialize(prison:)
+  def initialize(prison:, use_nomis_slots: false)
     @prison = prison
-    @slots = prison.available_slots.to_a
+    @slots = lookup_slots_for_prison(prison, use_nomis_slots)
   end
 
   # rubocop:disable Metrics/MethodLength
+  def lookup_slots_for_prison(prison, use_nomis)
```
alan: There is a lot of branching going on this method, I'd find it clearer if it was something like:

```ruby
def initialize(prison:, use_nomis_slots: false)
  @prison = prison
  @use_nomis = use_nomis_lots && Nomis::Api.enabled?
  @slots = lookup_nomis_slots(prison) || prison.available_slots.to_a
end

def lookup_nomis_slots(prison)
  return nil unless @use_nomis

  begin
     Nomis::Api.instance.fetch_bookable_slots(
       prison: prison,
      start_date: prison.first_bookable_date,
      end_date: prison.last_bookable_date
    )
 rescue Excon::Errors::Error => e
   Rails.logger.warn "Error calling the NOMIS API: #{e.inspect}"
   nil
end
```

With something like this we also avoid having to put the default slots in the rescue and the else branch.

@tatyree @mloughran what do you think?

tatyree: :+1:

app/models/slot_availability.rb
```diff
@@ -3,30 +3,10 @@ class SlotAvailability
 
   def initialize(prison:, use_nomis_slots: false)
     @prison = prison
```
alan: I'm 👍 on the refactor.

However I've noticed that we are using `prison` as an instance variable and as a parameter, would be good to make it consistent.

mloughran: Great! It's an interesting point r.e. `@prison`, but I think it's best as is. The instance variable is set for the `restrict_by_prisoner` method, which is called externally, however passing the prison parameter to the private methods makes the code simpler since it makes explicitly that they do not depend on mutable state.

alan: 👍

PR #263, opened 2016-04-08
Use locale in the API from the headers

alan:
Uses the locale from the header in the API instead of using a redundant locale
parameter.

The specs uncovered some missing welsh localisation that I've added in English.

Also as discussed I've refactored out the gem we were using to parse the accept
language header as we don't use it's functionality.


config/locales/cy/errors.yml
```diff
@@ -3,14 +3,22 @@ cy:
...
               too_many_adults: Gallwch sicrhau lle i uchafswm o %{max} ymwelydd dros
                 %{age} oed ar yr ymweliad hwn
+        slots_step:
+          attributes:
+            option_0:
+              inclusion: is not included in the list
+            option_1:
+              inclusion: is not included in the list
+            option_2:
+              inclusion: is not included in the list
```
tatyree: Shouldn't these be in `en/errors.yml`?

config/locales/cy/errors.yml
```diff
@@ -3,7 +3,7 @@ cy:
     errors:
       messages:
         name: nid yw'n enw dilys
-        blank: 'yn ofynnol '
+        blank: 'yn ofynnol'
```
mloughran: @alan did you change this in transifex? Otherwise it will just come back next time I pull from there.

spec/controllers/api/visits_controller_spec.rb
```diff
@@ -15,6 +15,10 @@
     JSON.parse(response.body)
   }
 
+  before do
+    request.headers['Accept-Language'] = 'cy'
+  end
```
mloughran: @alan it seems a bit odd to be setting the Accept-Language to welsh since these tests are not about testing localisation, and it generally makes the tests more awkward to read for non-welsh-speakers.

app/controllers/api/visits_controller.rb
```diff
@@ -44,7 +42,7 @@ def create
       fail_if_invalid('slot_options', slots_step)
 
       @visit = BookingRequestCreator.new.create!(
-        prisoner_step, visitors_step, slots_step, locale
+        prisoner_step, visitors_step, slots_step, I18n.locale
```
mloughran: Seems cleaner :+1: – I assume you've checked that this is thread safe?

alan: Rails is threadsafe, if it isn't then we already have the same problem...

config/locales/cy/errors.yml
alan: Nope, don't know about transifex

spec/controllers/api/visits_controller_spec.rb
alan: It seemed like a good idea as it found missing translations, I didn't find it to be a problem and I don't know welsh...

Happy to create a subset of tests to cover localisation.

config/locales/cy/errors.yml
mloughran: See https://github.com/ministryofjustice/prison-visits-2#welsh-translation – I'll show you and we can update this string tomorrow.

spec/controllers/api/visits_controller_spec.rb
mloughran: In that case I don't mind. It's not causing me a problem ;)

PR #267, opened 2016-04-12
Handle unknown formats in request a booking

alan:
Requesting a random format will cause Rails to try to search for a template with
a matching file extension. Since the file does not exist it raises a 500.

Adding `respond_to` in an action will only match on the formats specified by the
block. If a request is of an unknown format it raises an
`ActionController::UnknownFormat` which gets translated with a request with an
error code 406.

While this works for these endpoints it doesn't catch all endpoints, for that
there are 2 possible approaches:
  - one is to use the `responders` gem that was extracted from Rails 3 and use
  it on on the ApplicationController
  - change the routes so that they match on the format

I'd leave investigating this solutions until we deploy the public app and cleanup
pvb2.

### DIFF
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -1,6 +1,7 @@
 class ErrorsController < ApplicationController
   SUPPORTED_ERRORS = {
     404 => :'404',
+    406 => :'404',
     500 => :'500',
     503 => :'503'
   }
@@ -8,6 +9,7 @@ class ErrorsController < ApplicationController
   # Otherwise erroring POST requests can fail the CSRF check when rendering the
   # error page...
   skip_before_action :verify_authenticity_token
+  before_action :set_html_format, only: :show
 
   def show
     status_code = params.fetch(:status_code).to_i
@@ -15,11 +17,19 @@ class ErrorsController < ApplicationController
     # Explicity protect against rendering an unexpected template (although such
     # a situation should not be possible with a correct routes definition)
     template_to_render = SUPPORTED_ERRORS.fetch(status_code)
-
-    render template_to_render, status: status_code
+    render template_to_render, status: status_code, format: :html
   end
 
   def test
     fail 'This is an test exception'
+  end
+
+private
+
+  # We always want to render html back. Without this the moj_template gem errors
+  # out trying to find a template of an unknown extension when requesting an
+  # unknown format.
+  def set_html_format
+    request.format = :html
   end
 end
--- a/app/controllers/booking_requests_controller.rb
+++ b/app/controllers/booking_requests_controller.rb
@@ -6,7 +6,10 @@ class BookingRequestsController < ApplicationController
     @steps = processor.steps
     @step_name = processor.step_name
     append_to_log booking_step_rendered: processor.step_name
-    render processor.step_name
+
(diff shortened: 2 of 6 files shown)


mloughran:
Rather than putting `respond_to` inside in the action, would this not work?

    class BookingRequestsController < ApplicationController
      respond_to :html

Definitely agree r.e. fixing this globally after deploying public.

alan:
Nope, tried that. It has been extracted into the `responders` gem so is not part of Rails 4.

I've tested this with prod like configuration (consider_all_requests_local to false) and not totally happy with the outcome. When requesting an unknown format it just renders a blank page. Working on a solution for this now.

mloughran:
Hmm... strange – I wonder why it was removed. The other option as you suggest is just to restrict the format in the routes – I wonder if that's a simpler solution?

alan:
Is a more generic solution, although less correct as it will respond with a 404 instead of a 406.

There is another solution, which is to override the format to be always html, I feel the same way as the routes option, simple but less correct.

None of the options are ideal. I like the idea of using respond_to in the application controller best, but would like to avoid having to add the gem as a dependency. Might be worth investigating the reasons why it was pulled into a gem as we might find other alternatives to achieve the same behaviour without using the gem.

mloughran:
I'm not particularly worried r.e. returning 404, although I take your point. Overriding the format would by my least favourite option.

app/controllers/errors_controller.rb
```diff
@@ -1,25 +1,37 @@
...
     fail 'This is an test exception'
   end
+
+private
+
+  # We always want to render html back in the case of a 406. Without this the
+  # moj_template gem errors out trying to find a template of an unknown
+  # extension.
+  def set_html_format
+    request.format = :html if status_code == 406
```
mloughran: Is it worth setting the format to html in all cases? Is there ever a case where we would not want it to be html?

mloughran: Also, why do this is a `before_action`? It's only used in one method, so doesn't it make the code more complex?

alan: Needs to happen in a before_action, inside the action it doesn't work.

I'm not sure if we want this for all cases, that is why I made it apply only to 406. Didn't want to break existing behaviour by introducing this change.

mloughran: Interesting, I did not know that.

I think it's safe to always set html. There's no template for anything else, so I don't see how it can break anything.

alan: @mloughran updated to be generic, can you take another look?

app/controllers/errors_controller.rb
```diff
@@ -1,25 +1,37 @@
...
+  # We always want to render html back. Without this the moj_template gem errors
+  # out trying to find a template of an unknown extension when requesting an
+  # unknown format.
+  def set_html_format
+    request.format = :html
+  end
+
+  def status_code
+    params.fetch(:status_code).to_i
+  end
```
mloughran: A niggle, but `status_code` doesn't need to be extracted out to a method any more.

alan: Good spot, I'll fix

alan: Pushed the fix

PR #269, opened 2016-04-12
404 for unknown locales, only allow 'en' and 'cy'

alan:
Deletes the 'xx' locale used for testing since specs break with the locale
restriction on the routes.


config/routes.rb
```diff
@@ -17,7 +17,7 @@
   # Old pvb1 link that users got in an email
   get 'status/:id', controller: :pvb1_paths, action: :status, as: :pvb1_status
 
-  scope '/:locale', locale: /[a-z]{2}/ do
+  scope '/:locale', locale: /en|cy/ do
```
alan: The bit that enforces a known locale in this big diff

mloughran:
@alan looks good, although there's a failing spec according to circle.

alan:
Yep, looking into it. When running with `rake` the spec rails, when using `rspec` it passes.

alan:
```
SPEC=spec/features/language_switching_spec.rb rake spec
```

failed before I bumped `rspec-rails`, however running the specs with `rake` still fails...

PR #275, opened 2016-04-14
Handle zendesk_client not configured

alan:
If a Rails configuration is not set, then the method is not defined. Inside the
Zendesk job we check the value of Rails.configuration.zendesk_client, but that
raises a NoMethodError if it does not exist so the code to raise an error is
never executed.

### DIFF
--- a/spec/jobs/zendesk_tickets_job_spec.rb
+++ b/spec/jobs/zendesk_tickets_job_spec.rb
@@ -14,16 +14,18 @@ RSpec.describe ZendeskTicketsJob, type: :job do
   let(:client) { double(ZendeskAPI::Client) }
   let(:ticket) { double(ZendeskAPI::Ticket, save!: nil) }
 
-  before :each do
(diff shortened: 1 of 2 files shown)


spec/jobs/zendesk_tickets_job_spec.rb
```diff
@@ -14,16 +14,18 @@
   let(:client) { double(ZendeskAPI::Client) }
   let(:ticket) { double(ZendeskAPI::Ticket, save!: nil) }
 
-  before :each do
+  before(:each) do
```
tatyree: Do we even need :each here?  That's the default after all...

alan: I didn't analyse it, it was just bugging me seeing it without parenthesis...

alan: I think it is required for the previous tests?

PR #276, opened 2016-04-14
Handle unsupported http methods gracefully

alan:
Rails raises an error in its middleware if the http method is not supported.
When this causes it notifies Sentry which is just noise as we don't care about
them.


config/application.rb
```diff
@@ -8,6 +8,7 @@
 require 'action_mailer/railtie'
 require 'action_view/railtie'
 require 'sprockets/railtie'
+require_relative '../app/middlewares/http_method_not_allowed'
```
tatyree: Should be 'middleware', not 'middlewares'.

PR #277, opened 2016-04-15
Propagate the request id to nomis API

alan:
Stores the request id coming from X-Request-Id header and uses for request
logging and to pass it through to the Nomis API.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -17,6 +17,7 @@ gem 'pg'
 gem 'premailer-rails'
 gem 'puma'
 gem 'redcarpet'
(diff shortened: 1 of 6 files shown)


alan:
Similar to https://github.com/ministryofjustice/prison-visits-public/pull/27

mloughran:
Looks good, albeit with the same niggle as the other PR.

tatyree:
@alan LGTM.  I'll refactor it after the fashion of whatever gets done on the refactor referenced in ministryofjustice/prison-visits-public/pull/27.

PR #278, opened 2016-04-15
Gracefully handle invalid dates in validation API

mloughran:
Discovered while implementing https://github.com/ministryofjustice/prison-visits-public/pull/28.

In the end that PR doesn't rely on this change, but still worth fixing.


app/controllers/api/validations_controller.rb
```diff
@@ -1,9 +1,16 @@
 module Api
   class ValidationsController < ApiController
+    # rubocop:disable Metrics/MethodLength
     def prisoner
+      begin
+        date = Date.parse(params.fetch(:date_of_birth))
+      rescue ArgumentError
+        raise ParameterError, 'date_of_birth'
+      end
```
tatyree: Wouldn't this be better abstracted into a method?  

```
def validate_dob(date) 
  Date.parse(date)
rescue ArgumentError
  raise ParameterError, 'date_of_birth'
end
```

app/controllers/api/validations_controller.rb
```diff
@@ -1,9 +1,16 @@
 module Api
   class ValidationsController < ApiController
+    # rubocop:disable Metrics/MethodLength
```
SteveMarshall: Disabling metrics inline seems… not good… to me. Did you intend to commit this?

(The method length cop disabled here is usually a great indicator for 'you could probably describe your intent better by extracting methods'.)

app/controllers/api/validations_controller.rb
mloughran: No, the code is clear – a method call just adds indirection.

app/controllers/api/validations_controller.rb
mloughran: Of course. Over-adherence to arbitrary 'rules' seems to be a favourite hobby of many, and generally results in terrible code. I know why the rule is there, but when I choose to break it, I do so with no qualms.

SteveMarshall: Well, I have to object to this change then.

If you disagree with the rules, you should work with the team to agree and implement a new set (or update the current one), not disable them ad-hoc as they get in your way. This is no different than disabling tests and committing them that way because you don't like the result (which, in case it's not clear, is also something one should not do).

Disabling style rules (or tests) ad-hoc also sets a bad precedent, and a bad example to others: you may know when to (and not to) disable them, and why, but will the person reading your code? What about a junior developer on the team?

Let's discuss this in person on Monday.

app/controllers/api/validations_controller.rb
alan: I'd go further. I think the parameter check would be better off handled in the validation object if possible, otherwise the validation logic is split between the controller and the validation object.

app/controllers/api/validations_controller.rb
```diff
@@ -1,18 +1,35 @@
...
+        date = Date.parse(params.fetch(:date_of_birth))
+      rescue ArgumentError
+        raise ParameterError, 'date_of_birth'
+      end
+
+      noms_id = params.fetch(:number)
+
+      # TODO: Discuss removing this cop; especially when there are multiple
+      # return values I find an explicit return much cleaner.
+      # rubocop:disable Style/RedundantReturn
```
alan: As discussed, I agree with rubocop on this one. I'd find the code more idiomatic without the explicit return on the last line

alan:
👍 for now. Would be good to revisit these classes once we have implemented more features

PR #279, opened 2016-04-18
Cardiff don't work weekends

kyliemulholland:
Attempt to resolve metrics queries - Cardiff believe their stats are setup to include weekends as working days.


mloughran:
This is the default, so it shouldn't make any difference.

kyliemulholland:
Thanks for clarifying @mloughran - I've since found a potential error in the metrics which may be the root of this confusion on Cardiff's part.

PR #281, opened 2016-04-18
Reject visits that only have minors

alan:
Reject visits that have only minors as allowed visitors

### DIFF
--- a/app/models/booking_response.rb
+++ b/app/models/booking_response.rb
@@ -49,6 +49,7 @@ class BookingResponse
   def reason
     return 'visitor_not_on_list' if visitor_not_on_list?
     return 'visitor_banned' if visitor_banned?
+    return 'no_adult' unless at_least_one_valid_visitor?
     selection
   end
 
--- a/app/models/booking_response.rb
+++ b/app/models/booking_response.rb
@@ -3,6 +3,10 @@ class BookingResponse
 
   SLOTS = %w[ slot_0 slot_1 slot_2 ]
 
+  # The adult age for accepting a booking is different from the configurable
+  # adult age in Prison
+  ADULT_AGE = 18
+
   attribute :visit
 
   attribute :selection, String, default: 'none'
@@ -71,8 +75,10 @@ class BookingResponse
 private
 
   def at_least_one_valid_visitor?
-    (visitors.map(&:id) -
-     [unlisted_visitor_ids, banned_visitor_ids].flatten).present?
+    visitors.
+      reject { |visitor| visitor.id.in? unlisted_visitor_ids }.
+      reject { |visitor| visitor.id.in? banned_visitor_ids }.
+      any? { |visitor| visitor.age >= ADULT_AGE }
   end
 
   def validate_checked_visitors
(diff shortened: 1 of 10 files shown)


app/models/booking_response.rb
```diff
@@ -71,8 +75,10 @@ def banned_visitors
 private
 
   def at_least_one_valid_visitor?
-    (visitors.map(&:id) -
-     [unlisted_visitor_ids, banned_visitor_ids].flatten).present?
+    visitors.
+      reject { |visitor| visitor.id.in? unlisted_visitor_ids }.
+      reject { |visitor| visitor.id.in? banned_visitor_ids }.
+      any? { |visitor| visitor.age >= ADULT_AGE }
```
mloughran: +1, much cleaner.

spec/models/booking_response_spec.rb
```diff
@@ -57,6 +57,18 @@
         end
       end
 
+      context 'and the allowed visitor is a child' do
+        let(:visit) { create(:visit) }
+
+        before do
+          visit.visitors.update_all(date_of_birth: 5.years.ago)
```
mloughran: Is it worth making this text explicit regarding the difference between a 'child' and 'under 18'? (i.e. this is a test about under 18, not about children)

I'm not sure if this class is currently `adult_age` aware, but the age in this test should be greater than `adult_age` but less than 18.

alan: Not sure if its `adult_age` aware, I'll double check, I'll add spec for that to make it more explicit.

tatyree:
Additional spec aside, LGTM 👍

alan:
I need to make changes to the email, otherwise it might not say the reason why is being rejected

config/locales/en/mailers.yml
```diff
@@ -88,6 +88,9 @@ en:
       allowance_will_renew: >-
         %{prisoner} will have their full visiting allowance (VO) renewed on
         %{date}.
+      no_adult:
```
alan: This will need welsh translation, let me know what I need to do. I haven't got Transifex configured on my machine

config/locales/en/mailers.yml
```diff
@@ -88,6 +88,9 @@ en:
...
+        There are no approved adults on the visit. There needs to be at least
+        one adult in the visit.
```
alan: Thoughts on the copy?

app/models/rejection.rb
```diff
@@ -1,6 +1,7 @@
 class Rejection < ActiveRecord::Base
   REASONS = %w[
     slot_unavailable
+    no_adult
```
tatyree: I'm afraid we're going to have to add a column to https://github.com/ministryofjustice/prison-visits-2/blob/master/app/views/metrics/_all_prisons.html.erb for this...

Will need to check with @kyliemulholland

kyliemulholland: @tatyree Yes, we will. We should re-evaluate metrics at some point, I agree, but that won't be for a while. In the meantime please add the column.

alan: I'll pick this up

PR #282, opened 2016-04-19
Use staff subdomain for prison urls

alan:
We want staff emails to use a subdomain so that requests can be routed to the
pvb2, without this they'd be routed to the public app once deployed.


alan:
This is the best approach I can think off, don't like that one must remember to use the shared email config method, if you know of a better way let me know!

alan:
Only to be merged once the subdomain has been set up

PR #286, opened 2016-04-21
Include visit id on visitor rejection email

alan:


mloughran:
I just realised that we add the visit ID inconsistently (sometimes we interpolate). Maybe it's worth taking the opportunity to make it consistent. Search for `visit_id: "Visit ID: %{visit_id}"`

PR #290, opened 2016-04-22
Fix ZendeskErrors due to stricter email validation

alan:
Zendesk has stricter email validations than just trying to parse the email
(a recent change for pvb-public). This re-introduces email checker, but without
using Sendgrid for validations.

I spent a good time trying to extend EmailChecker to be better at detecting
formatting errors, or only formatting errors (without MX checks or Sendgrid) but
it became a rabbit hole in terms of design and complexity of validating properly
the format of an email address and domain part.

At the end I decided to stick with the current formatting checks and MX
validation as is a more incremental change. In the future we could look at
improving the validation by validating the domain format.

### DIFF
--- a/app/models/feedback_submission.rb
+++ b/app/models/feedback_submission.rb
@@ -11,8 +11,13 @@ private
   end
 
   def email_format
-    Mail::Address.new(email_address)
-  rescue Mail::Field::ParseError
-    errors.add(:email_address, 'has incorrect format')
+    return if email_address.blank?
+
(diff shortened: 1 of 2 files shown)


alan:
I'll port the same validation checks to public so that they are in sync next.

Once deployed I'll clean the data in Redis so that the feedback can retried to be submitted to Zendesk

tatyree:
I don't think I've ever seen a fully effective email validator.

PR #298, opened 2016-04-26
Requesting visitor must be 18

tatyree:
The 'lead visitor' (i.e. the visitor making the request) must be over 18
years of age, even if the prison has a lower threshold for 'adult_age'.

### DIFF
--- a/app/models/prison.rb
+++ b/app/models/prison.rb
@@ -3,7 +3,7 @@ class Prison < ActiveRecord::Base
 
   MAX_VISITORS = 6
   MAX_ADULTS = 3
-  REQUESTING_VISITOR_MIN_AGE = 18
+  LEAD_VISITOR_MIN_AGE = 18
 
   has_many :visits, dependent: :destroy
   belongs_to :estate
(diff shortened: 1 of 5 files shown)


app/models/prison.rb
```diff
@@ -49,11 +49,14 @@ def confirm_by(today = Time.zone.today)
   end
 
   def validate_visitor_ages_on(target, field, ages)
+    return if ages.blank?
     adults, _children = ages.partition { |a| adult?(a) }.map(&:length)
-    if adults > MAX_ADULTS
+    # For simplicity’s sake, this operates on the possibly unsafe assumption
+    # that the first visitor is always going to be the requesting visitor.
+    if ages.first < REQUESTING_VISITOR_MIN_AGE
```
alan: Could you sort the ages to avoid the assumption?

tatyree: Sorting by age would defeat the purpose.  If the *requesting* (first) visitor is 12, but there is some over 18 in the group, then the request would be allowed.  This is to ensure that the person making the request is over 18, not that the group has an adult in it. I'm guessing there is a legal requirement regarding the minimum age of the requestor?  @kyliemulholland?

mloughran: @tatyree I see what you're saying and happy to go with the lead visitor being the first visit for now. We can always change later if needs be.

Separately, should both checks still be part of the same if-else block – aren't they independent? It's possible for the lead visitor to be a non-adult _and_ for there too many non-child visitors.

tatyree: Good point, @mloughran.  About to push a fix.

config/locales/en/errors.yml
```diff
@@ -16,6 +16,8 @@ en:
         visitors_step:
           attributes:
             general:
+              requestor_must_be_over_18:
+                The person requesting the visit must be over the age of 18
               too_few_adults:
```
mloughran: @tatyree can `too_few_adults` now be removed?

app/models/prison.rb
```diff
@@ -49,11 +49,16 @@ def confirm_by(today = Time.zone.today)
   end
 
   def validate_visitor_ages_on(target, field, ages)
+    return if ages.blank?
```
mloughran: @alan do you understand the reason for this return? It seems non-obvious. I think we should find a way to avoid this, or at least add a comment.

PR #299, opened 2016-04-26
Change date and name formatting for prisons

alan:
Changes dates to the same as used in nomis (dd/mm/yyyy) so that staff can copy
and paste.

Also changes the way that names are serialised to make it clear what is the
first name and last name.


alan:
**Email screenshot**

screen shot 2016-04-26 at 14 52 43

PR #301, opened 2016-04-28
Remove email validations using Sendgrid

alan:
Using the Sendgrid API ads significant latency to a request, anything in between
700ms and 4s.

It is unclear what value brings this validation so we've decided to remove it
and analyse the emails later on to see if it is worth reconsidering the problem.


alan:
I've left the Sendgrid client code so that we can analyse the emails later on. It can be deleted later on if we keep this approach.

mloughran:
How satisfying to remove all this code :)

PR #304, opened 2016-05-04
Show the day of week for visits to staff

alan:
Slots now show like 'Wednesday 04/05/2016 13:00 - 13:45'


kyliemulholland:
@alan looks good to me.

PR #316, opened 2016-05-09
Improve Rails error handling

alan:
Our exception app is an ActionController action (they are Rack apps), the
problem is that if an error happens inside that application it will be bubble
out of the error app resulting in a non-styled error page showed to the user and
the error not being reported to Sentry.

We've seen this happen with invaid query strings that involve invalid nested
attributes. To fix this we need to detect if the query string is invalid before
is processed by the exception app, otherwise the controller will try to parse
the query string before processing the main body of the action and raise an
error.

An alternative of doing this would be of decoupling our error pages from Rails
controllers / views, and have a Rack app that has the same behaviour while not
having to use the query string.

### DIFF
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -12,12 +12,16 @@ class ErrorsController < ApplicationController
   before_action :set_html_format, only: :show
 
   def show
-    status_code = params.fetch(:status_code).to_i
+    append_to_log(original_fullpath: request.original_fullpath)
+
+    status_code = env['PATH_INFO'][1..-1]
 
     # Explicity protect against rendering an unexpected template (although such
     # a situation should not be possible with a correct routes definition)
-    template_to_render = SUPPORTED_ERRORS.fetch(status_code)
-    render template_to_render, status: status_code, format: :html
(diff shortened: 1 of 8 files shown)


app/views/layouts/application.html.erb
```diff
@@ -76,7 +76,11 @@
   </li>
   <% alternative_locales.each do |locale| %>
     <li>
-      <%= link_to(t('language', locale: locale), locale: locale) %>
```
alan: This is here since the implicit link breaks from moving the `exception_app` to stop using the Rails routes.

Not sure about the value it has to let the users switch to another language from the error pages, I think we can probably remove the link if it is an error page.

app/controllers/errors_controller.rb
```diff
@@ -12,12 +12,16 @@ class ErrorsController < ApplicationController
   before_action :set_html_format, only: :show
 
   def show
-    status_code = params.fetch(:status_code).to_i
+    append_to_log(original_fullpath: request.original_fullpath)
```
alan: Without this lograge will only log that `/404` was rendered for example

app/views/layouts/application.html.erb
mloughran: I doubt anyone will miss it.

PR #321, opened 2016-05-11
Send the request_id to Sentry

alan:
Should make debugging errors easier. I've also double checked the Raven source
for that method and should be threadsafe.


mloughran:
Nice to see `append_to_log` taking a trip ;)

PR #326, opened 2016-05-16
Remove unused JS and CSS assets

chrishanes:
Remove unused frontend assets primarily used for the public app


mloughran:
@chrishanes if you rebase this onto the remove-public-code branch, then you can check that the tests pass. If you like you can create a PR into that branch, but I'd be happy for you just to push to that branch.

PR #327, opened 2016-05-16
Remove unused frontend JS and CSS assets

chrishanes:


mloughran:
@chrishanes surprisingly little. I would squash these 2 commits together but otherwise LGTM.

PR #329, opened 2016-05-17
Dashboards routes and controllers

alan:
Adds an index with links to each estate dashboard, and an empty dashboard.


alan:
Failing spec is a feature spec to book a visit. It can be merged as this code will be deleted soon, before this code makes it to 'master'

PR #334, opened 2016-05-18
Add message when no visits

chrishanes:
Add a message to the prison visit list when there are no visits, plus update the application template

### DIFF
--- a/app/views/prison/dashboards/show.html.erb
+++ b/app/views/prison/dashboards/show.html.erb
@@ -1,30 +1,37 @@
-<h2 class="heading-large">Latest booking requests</h2>
+<h2 class="heading-large">Latest visit requests</h2>
 
-<table>
-  <caption class="visuallyhidden">Visit inbox</caption>
-  <thead>
-    <tr>
-      <th class="" scope="col">Prisoner No.</th>
-      <th scope="col">Prisoner Name</th>
-      <th class="" scope="col">Visit Date</th>
-      <th class="" scope="col">Time slot</th>
-      <th scope="col">Lead Visitor</th>
-      <th scope="col">Tel.</th>
-      <th scope="col">Action</th>
-    </tr>
-  </thead>
-  <tbody>
-    <% @requested_visits.each do |visit| %>
+<% if @requested_visits.any? %>
+
+  <table>
+    <caption class="visuallyhidden">Visit inbox</caption>
+    <thead>
       <tr>
-        <td><%= visit.prisoner_number %></td>
-        <td><%= visit.prisoner_full_name %></td>
-        <td class=""><%= visit.slots.first.to_date %></td>
-        <td class=""><%= format_slot_times(visit.slots.first) %></td>
-        <td><%= visit.visitor_full_name %></td>
-        <td><%= visit.contact_phone_no %></td>
-        <td><%= link_to 'View Booking', prison_visit_path(visit) %></td>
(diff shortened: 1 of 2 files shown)


app/views/prison/dashboards/show.html.erb
```diff
@@ -1,30 +1,37 @@
...
-      <th class="" scope="col">Visit Date</th>
-      <th class="" scope="col">Time slot</th>
-      <th scope="col">Lead Visitor</th>
-      <th scope="col">Tel.</th>
-      <th scope="col">Action</th>
-    </tr>
-  </thead>
-  <tbody>
-    <% @requested_visits.each do |visit| %>
+<% if @requested_visits.exists? %>
```
alan: Using `any?` instead of `exists?` would be more idiomatic Rails / Ruby

app/views/layouts/application.html.erb
```diff
@@ -8,27 +8,45 @@
...
 <% end %>
 
 <% content_for :content do %>
-  <main>
-    <header>
-      <h1 class="heading-large">
-        <%= yield :header %>
-      </h1>
-    </header>
+<!--   <main id="wrapper" role="main">
```
alan: Can we delete these comments?

app/views/layouts/application.html.erb
```diff
@@ -8,27 +8,45 @@
   <%= stylesheet_link_tag('application', media: 'all') %>
 <% end %>
 
+<% content_for :header_class do "with-proposition" end %>
```
alan: Instead of using do / end on a single line you can use `{ }`, they are preferred when the block is in a single line

app/views/prison/dashboards/show.html.erb
```diff
@@ -1,30 +1,37 @@
...
-        <td><%= visit.prisoner_number %></td>
-        <td><%= visit.prisoner_full_name %></td>
-        <td class=""><%= visit.slots.first.to_date %></td>
-        <td class=""><%= format_slot_times(visit.slots.first) %></td>
-        <td><%= visit.visitor_full_name %></td>
-        <td><%= visit.contact_phone_no %></td>
-        <td><%= link_to 'View Booking', prison_visit_path(visit) %></td>
+        <th class="" scope="col">Prisoner No.</th>
+        <th scope="col">Prisoner Name</th>
+        <th class="" scope="col">Visit Date</th>
```
alan: Can we get rid of the empty classes?

app/views/layouts/application.html.erb
```diff
@@ -8,6 +8,18 @@
...
+<% content_for :proposition_header do %>
+  <div class="header-proposition">
+    <div class="content">
+      <nav id="proposition-menu">
+        <a href="/" id="proposition-name"><%= t('app_title') %></a>
+      </nav>
+    </div>
+  </div>
+<% end %>
+
```
alan: How are this changes in the layout related to messaging of no visits?

app/views/layouts/application.html.erb
alan: There is a small change though that `content_for` will only work with do / end

app/views/layouts/application.html.erb
chrishanes: I suppose it's not related but I noticed that the template being used was causing a problem. I will remove this and do in a separate branch.

PR #336, opened 2016-05-18
Visitors validation API

alan:
Takes care of validation the visitors for a visit:
- Lead visitor age
- Number of adults (prison defined age)
- Number of visitors

### DIFF
--- a/app/models/visitors_validation.rb
+++ b/app/models/visitors_validation.rb
@@ -1,0 +1,45 @@
+class VisitorsValidation
+  include NonPersistedModel
+
+  attribute :lead_date_of_birth, Date
+  attribute :dates_of_birth, Array[Date]
+  attribute :prison, Prison
+
+  validate :lead_visitor_age
+  validate :number_of_visitors
+  validate :number_of_adults
+
+  def lead_visitor_age
+    if age_calculator.age(lead_date_of_birth) < Prison::LEAD_VISITOR_MIN_AGE
+      errors.add(:general, 'lead_visitor_age')
+    end
+  end
+
+  def number_of_visitors
+    if dates_of_birth.size > Prison::MAX_VISITORS
+      errors.add(:general, 'too_many_visitors')
+    end
+  end
+
+  def number_of_adults
+    adults = dates_of_birth.select { |dob| considered_as_adult?(dob) }
+
+    if adults.size > Prison::MAX_ADULTS
+      errors.add(:general, 'too_many_adults')
+    end
+  end
+
+  def error_keys
+    errors[:general]
+  end
+
+private
+
+  def age_calculator
+    AgeCalculator.new
+  end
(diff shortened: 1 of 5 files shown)


alan:
Related to https://github.com/ministryofjustice/prison-visits-public/pull/54

app/models/visitors_validation.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def lead_visitor_age
+    if age_calculator.age(lead_date_of_birth) < MIN_LEAD_VISITOR_AGE
+      errors.add(:general, 'lead_visitor_age')
+    end
+  end
+
+  def number_of_visitors
+    if dates_of_birth.size > Prison::MAX_VISITORS
+      errors.add(:general, 'too_many_visitors')
```
alan: Using strings matters a lot. When using symbols the error message is localised when fetching the values on `errors`, had me confused for a while.

mloughran: Gotta love the magic...

spec/models/visitors_validation_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+        subject.valid?
+        expect(subject.error_keys).to eq(['too_many_visitors'])
+      end
+    end
+
+    context 'when there are too many adults' do
+      let(:dobs) do
+        (Prison::MAX_ADULTS + 1).times.map {
+          prison.adult_age.years.ago.to_date
+        }
```
mloughran: I wonder if it's clearer to write this? (ditto elsewhere)

    [prison.adult_age.years.ago.to_date] * (Prison::MAX_ADULTS + 1)

mloughran:
@alan what are your thoughts on the duplication of visitor validation between this code and e.g. `Prison#validate_visitor_ages_on`?

alan:
I didn't know of that method. I think the duplication, when we remove the steps code then we can delete `Prison#validate_visitor_ages_on`

mloughran:
@alan +1 for refactoring when we remove the steps. I think it's a bit more nuanced – the request booking method still needs to validate the booking (it can't rely on this validation API having been called) – but we'll find a solution.

app/models/visitors_validation.rb
```diff
@@ -0,0 +1,47 @@
+class VisitorsValidation
+  include NonPersistedModel
+
+  MIN_LEAD_VISITOR_AGE = 18 # years
```
mloughran: This is duplicated in Prison also. You're using `Prison::MAX_VISITORS` so probably best to be consistent.

alan: Good catch

alan:
Yes, I expect the request booking method (I guess you mean to create a visit) will run the same set of validations as the validations endpoints as a minimum.

PR #337, opened 2016-05-18
Staff interface to view requested visits

alan:
Shows the requested visits per Estate (estates with more than 1 prison use the same email for social visits).

This PR has 2 new views, one on `/prison` that lists links to all the estates dashboard and then the view for each estate `/prison/pentonville`. We need to decide final urls for these.

Not sure what's happened to the branch with the duplicate commits, we'll clean it up

PR #338, opened 2016-05-19
Staff interface for requested visits

alan:
Adds an index page with a list of all the links of each Estate dashboard and the
dashboard to view requested visits.

The dashboard is per Estate. Estates always use the same email for social visits
for their prisons.


alan:
This has a clean commit history compared to the original PR #337

PR #340, opened 2016-05-19
Add navigation bar to requested and processed visits

chrishanes:
Add a navigation bar to switch between requested and processed visits and back to the list of prisons


app/views/prison/dashboards/processed.html.erb
```diff
@@ -1,3 +1,5 @@
+<%= render('navigation', {estate: @estate, visits: @processed_visits}) %>
```
alan: The brackets and curly braces are normally omitted when used with render

PR #341, opened 2016-05-20
Styling and format amends

chrishanes:
Some styling, formatting and copy amends to the staff visit pages

### DIFF
--- a/app/views/prison/dashboards/show.html.erb
+++ b/app/views/prison/dashboards/show.html.erb
@@ -1,18 +1,18 @@
 <%= render 'navigation', estate: @estate, visits: @requested_visits %>
 
-<h2 class="heading-large">Latest visit requests</h2>
+<h2 class="heading-large"><%= t('.title') %></h2>
 
 <% if @requested_visits.any? %>
 
   <table>
-    <caption class="visuallyhidden">Visit inbox</caption>
+    <caption class="visuallyhidden"><%= t('.title') %></caption>
     <thead>
       <tr>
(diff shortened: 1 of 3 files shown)


app/views/prison/dashboards/show.html.erb
```diff
@@ -1,32 +1,32 @@
...
       <tr>
-        <th>Prisoner No.</th>
-        <th>Prisoner Name</th>
-        <th>Visit Date</th>
+        <th>Prisoner name</th>
+        <th>Prisoner no.</th>
+        <th>Visit date</th>
         <th>Time slot</th>
-        <th>Lead Visitor</th>
+        <th>Lead visitor</th>
```
mloughran: Are we planning to keep strings in the locale files for the staff interface? /cc @alan @chrishanes

chrishanes: @alan mentioned the staff interface was going to be only in English, but I had started to add copy in the locales file. Shall I add all text into the locale file?

mloughran: Before I joined the project the convention was established to use the locale files for everything, with the justification that it maintains a clean separation between content and code, and that content writers are comfortable editing yaml. Personally I'm not overly concerned either way, but for the sake of consistency I think we should put everything or nothing in locale files, and unless there's a good reason to change let's keep using locale files.

alan: I might have not been clear @chrishanes when discussing locales, @mloughran is right and everything should be on locale files.

PR #342, opened 2016-05-23
Staff visit view

chrishanes:


app/views/prison_mailer/request_received.html.erb
```diff
@@ -56,15 +56,15 @@
 
 <p class="cta">
   <%= link_to(t('.process_booking'),
-              prison_visit_url(@visit, locale: I18n.locale)) %>
+              prison_visit_process_url(@visit, locale: I18n.locale)) %>
```
alan: The only thing that has changed here is the name of the helper, the value generated is the same so old emails will keep working.

PR #353, opened 2016-05-27
Set a real noreply email address for prison emails

alan:
Some prisons reply to the emails they get thinking that they get sent to the
visitor, instead get sent to Sendgrid.

This is because with Sendgrid whitelabel settings they set up an MX record that
accepts emails to all inboxes, so emails sent there don't bounce.

This change fixes this with Sendgrid recommended 'solution', which is to send
the email from a fake subdomain from the one used in the whitelabel. This
preserves the whitelabel functionality with the correct anti spam signatures
and make the emails bounce if people reply to them.

### DIFF
--- a/app/mailers/prison_mailer.rb
+++ b/app/mailers/prison_mailer.rb
@@ -12,49 +12,42 @@ class PrisonMailer < ActionMailer::Base
 
   def request_received(visit)
     @visit = visit
+    preferred_date = visit.slots.first.begin_at
 
-    mail(
-      from: I18n.t('mailer.noreply', domain: smtp_settings[:domain]),
-      to: visit.prison_email_address,
-      subject: default_i18n_subject(
-        full_name: visit.prisoner_full_name,
-        request_date: format_date_without_year(visit.slots.first.begin_at)
-      )
-    )
+    mail_prison(visit,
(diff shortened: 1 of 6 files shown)


app/mailers/prison_mailer.rb
```diff
@@ -12,50 +12,43 @@ class PrisonMailer < ActionMailer::Base
...
+
+private
+
+  def mail_prison(visit, i18n_options = {})
     mail(
-      from: I18n.t('mailer.noreply', domain: smtp_settings[:domain]),
+      from: I18n.t('mailer.noreply', domain: noreply_domain),
       to: visit.prison_email_address,
-      subject: default_i18n_subject(cancelled_attributes(visit))
+      subject: default_i18n_subject(i18n_options)
```
mloughran: Rather than passing `i18n_options` into this method, which feels rather opaque, how about just passing in the subject string?

alan: I think that would be clearer, but would want to update the visitor mailer to keep them consistent

mloughran: Ah, hadn't realised that. In that case let's leave like this for now.

PR #355, opened 2016-05-31
Serialize visits consistently

alan:
Make the serialisation consistent regardless of the API action. This is needed
so that PVB Public can show more information on the visit show page.


mloughran:
This definitely seems sensible. Shall I merge?

PR #356, opened 2016-05-31
Serialise more fields on visit

alan:
To be used in PVB Public for the visit show page.

### DIFF
--- a/app/views/api/visits/show.json.jbuilder
+++ b/app/views/api/visits/show.json.jbuilder
@@ -1,4 +1,13 @@
 json.visit do
-  json.id @visit.id
+  json.prison_id @visit.prison.id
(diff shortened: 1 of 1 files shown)


mloughran:
Is there an argument for just adding this information to the get visit API, rather than adding a new API call?

alan:
I thought about it and decided to implement it like this because it is the simplest solution for the problem I was having and it keeps it consistent with the other endpoints as we don't embed sub resources anywhere else.

I would consider extending the visits api to conditionally load sub resources if we see a need to do that, for example for performance reasons. Something similar to 'include' in JSON-API (http://jsonapi.org/format/#fetching-includes)

alan:
Reworked to include visitors and other required fields when serialising visit

PR #359, opened 2016-06-01
Serialize visits consistently

alan:
Make the serialisation consistent regardless of the API action. This is needed
so that PVB Public can show more information on the visit show page.

It also adds additional fields that Public needs.

### DIFF
--- a/app/views/api/visits/show.json.jbuilder
+++ b/app/views/api/visits/show.json.jbuilder
@@ -1,4 +1,14 @@
 json.visit do
   json.id @visit.id
   json.processing_state @visit.processing_state
+  json.prison_id @visit.prison.id
+  json.confirm_by @visit.confirm_by
+  json.contact_email_address @visit.contact_email_address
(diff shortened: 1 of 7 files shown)


app/views/api/visits/show.json.jbuilder
```diff
@@ -1,4 +1,14 @@
...
   json.processing_state @visit.processing_state
+  json.prison_id @visit.prison.id
+  json.confirm_by @visit.confirm_by
+  json.contact_email_address @visit.contact_email_address
+  json.slots @visit.slots.map(&:iso8601)
+  json.slot_granted @visit.slot_granted&.iso8601
+
+  json.visitors @visit.visitors do |visitor|
+    json.anonymized_name visitor.anonymized_name
+    json.allowed !visitor.banned && !visitor.not_on_list
```
mloughran: Is it worth pulling `allowed` into a method on visitor so that it can be tested? Feels like something that should be modelled on the domain rather than just in the view.

PR #360, opened 2016-06-03
Visit cancellations by staff

alan:
We know that prison staff cancel visits, we are unsure how they do this since there is no ability to cancel as staff. One possibility is that they use the 'cancel' link in the copy of the visitors email that they get.

This PR changes the link url that gets generated on the copied visitor view to a new page where the staff can cancel a visit. This cancellation then gets email to the visitor instead of the prison.

The page where the cancellation is rendered is a new show page adapted from the 'updated' template. Ideally we'd use the new show page we've done for the dashboards. We should switch the redirect to the proper show page once we 'launch' the dashboards (as well as adding the ability to cancel from the dashboard show page)

### DIFF
--- a/app/assets/stylesheets/_notification.scss
+++ b/app/assets/stylesheets/_notification.scss
@@ -1,0 +1,4 @@
+.notification {
+  border: 4px solid $button-colour;
+  padding: $gutter-half;
+}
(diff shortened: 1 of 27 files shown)


alan:
To finish it needs a view of the copy of the visitor email and the temporary show page. The temporary show page could with some minor styling, specially the 'notice' message and the cancellation form.

alan:
Once deployed we should backfill cancellation records for old visit cancellations to keep the data consistent.

PR #363, opened 2016-06-06
Enable Downview

SteveMarshall:
Also fix order of recurring visit days, and complete address.

This fixes #361 by reverting #313, which introduced an invalid file and filename.


kyliemulholland:
Tested in staging, booking processed end-to-end without issue. Will deploy to production.

PR #370, opened 2016-06-10
Add show/hide request details

chrishanes:
Show more of the visit request details with a show/hide functionality

### DIFF
--- a/app/views/prison/dashboards/_request_details_row.html.erb
+++ b/app/views/prison/dashboards/_request_details_row.html.erb
@@ -1,0 +1,43 @@
+<tr class="hidden-row hide">
+  <td colspan="9" class="request-details">
+    <div class="request-row-details">
+      <dl>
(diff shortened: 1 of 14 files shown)


app/views/prison/dashboards/_request_details_row.html.erb
```diff
@@ -0,0 +1,43 @@
...
+      <dl>
+        <dt><%= t('email_address', scope: :shared) %></dt>
+        <dd><%= visit.contact_email_address %></dd>
+      </dl>
+      <dl>
+        <dt><%= t('tel', scope: :shared) %></dt>
+        <dd><%= visit.contact_phone_no %></dd>
+      </dl>
+    </div>
+    <% if visit.visitors.count > 1 %>
```
alan: I'd change this `count` to a `size`, that way we avoid a n+1 DB query. Size does the count in ruby

PR #373, opened 2016-06-13
Temporary login

alan:
Introduces users for login, these are associated to an estate.

Also removes the list of prisons since the data is scoped by the users estate, it is no longer possible to view any estate logged out or when logged in.


config/locales/en/prison_views.yml
```diff
@@ -228,3 +228,7 @@ en:
       slot_unavailable: None of the dates and times chosen were available
       visitor_banned: A visitor has been banned
       visitor_not_on_list: A visitor is not on the contact list
+  devise:
```
alan: This could go on the devise locale file, that way all the devise copy is in the same file

PR #395, opened 2016-06-20
Format Sidekiq logs in logstash format

alan:
Logstash only receives http app logs which means that we have no visibility on
how the asynchronous jobs have performed when debugging issues.

It works by broadcasting the Sidekiq logs to the logstash log file and merging
event data about the processing from the Sidekiq logs and ActiveJob
instrumentation. This is necessary because Sidekiq and ActiveJob do not log
enough data by themselves. This approach is similar to the Logstasher gem that
currently formats http requests for logstasher.

### DIFF
--- a/lib/logstash_sidekiq_logger.rb
+++ b/lib/logstash_sidekiq_logger.rb
@@ -1,0 +1,60 @@
+# Sets up a new sidekiq file logger for logstash consumption.
+#
+# To do this it combines data from ActiveJob notifications and from Sidekiq
+# logs.
+#
+# ActiveJob notifications provides data on what job ran with what parameters,
+# this is obtained by setting up a subscription to the ActiveJob instrumentaton
+# that Rails provides.
+#
+# Sidekiq provides no instrumentation so the ActiveJob data is combined on a
+# custom log formatter.
+#
+# The final output looks like:
+#
+# (as single lines in the log file, output generated from
+# LogstashSidekiqLogger::Formatter#extract_log_data)
+#
+# {
+#   "job_name":"prison_mailer_request_received",
+#   "arguments":["gid://prison-visits/Visit/0125d92d-3e27-4835-bc35-7fd5fb"],
+#   "queue_name":"mailers",
+#   "status":"completed",
+#   "active_job_duration":640.827,
+#   "total_duration":691.0,
+#   "message":"[completed] 691.0 ms prison_mailer_request_received arguments
+#   "@timestamp":"2016-06-17T13:22:58.096Z",
+#   "@version":"1"
+# }
+#
+# {
+#   "job_name":"prison_mailer_request_received",
+#   "arguments":["gid://prison-visits/Visit/0125d92d-3e27-4835-bc35-7fd5fb"],
+#   "queue_name":"mailers",
+#   "status":"to_be_retried",
+#   "active_job_duration":1.194,
+#   "total_duration":4.0,
+#   "message":"[to_be_retried] 4.0 ms prison_mailer_request_received arguments
+#   "retry_number":1,
+#   "@timestamp":"2016-06-17T13:00:26.298Z",
+#   "@version":"1"
+# }
+
+require_relative 'logstash_sidekiq_logger/active_job_subscriber'
+require_relative 'logstash_sidekiq_logger/formatter'
+
+module LogstashSidekiqLogger
+  def self.setup(log_file)
+    # Listen to Rails ActiveJob instrumentation
+    ActiveJobSubscriber.attach_to :active_job
+
+    custom_sidekiq_logger = ActiveSupport::Logger.new(log_file)
+    custom_sidekiq_logger.formatter = Formatter.new
+
+    # ActiveSupport::Logger.broadcast wraps an existing logger so that logs
+    # going to that logger get broadcasted to a different logger
+    Sidekiq.
+      logger.
+      extend(ActiveSupport::Logger.broadcast(custom_sidekiq_logger))
+  end
+end
--- a/lib/logstash_sidekiq_logger/formatter.rb
+++ b/lib/logstash_sidekiq_logger/formatter.rb
@@ -1,0 +1,105 @@
+module LogstashSidekiqLogger
+  class Formatter < Logger::Formatter
+    def call(_severity, _time, _program_name, message)
+      store_message_failure_metadata(message)
+
+      performed_job = RequestStore.store[:performed_job]
+      return unless performed_job && message_needs_logging?(message)
+
+      data = extract_log_data(performed_job, message)
+      event = LogStash::Event.new(data)
+
+      RequestStore.clear!
+
+      "\n" + event.to_json
+    end
+
+  private
+
+    def store_message_failure_metadata(message)
+      return unless message =~ /^fail/
+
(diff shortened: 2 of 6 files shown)


lib/logstash_sidekiq_logger.rb
```diff
@@ -0,0 +1,37 @@
...
+#
+# Sidekiq provides no instrumentation so the ActiveJob data is combined on a
+# custom log formatter.
+#
+# The final output looks like this:
+#
+# rubocop:disable LineLength
+# {"job_name":"prison_mailer_request_received","arguments":["gid://prison-visits/Visit/0125d92d-3e27-4835-bc35-7fd5fb75033e"],"queue_name":"mailers","status":"completed","active_job_duration":640.827,"total_duration":691.0,"@timestamp":"2016-06-17T13:22:58.096Z","@version":"1"}
+# {"job_name":"prison_mailer_request_received","arguments":["gid://prison-visits/Visit/0125d92d-3e27-4835-bc35-7fd5fb75033e"],"queue_name":"mailers","status":"to_be_retried","active_job_duration":1.194,"total_duration":4.0,"retry_number":1,"@timestamp":"2016-06-17T13:00:26.298Z","@version":"1"}
+# rubocop:enable LineLength
```
SteveMarshall: What're these for?

If it's for readers of this documentation to see a realistic sample of the fields in the log lines, I might be tempted to break it over multiple lines so it's easier to see which bits are keys and which are values (and that way, you can remove the `rubocop:disable`). That said, I worry that this will become out of sync with the actual code, particularly given that the code is in an entirely other file.

Maybe instead say something like “Final output contains the fields specified in LogstashSidekiqLogger.Formatter#base_log_data, compressed onto one line”?

lib/logstash_sidekiq_logger/formatter.rb
```diff
@@ -0,0 +1,111 @@
...
+
+      performed_job = RequestStore.store[:performed_job]
+      return unless performed_job && message_needs_logging?(message)
+
+      data = extract_log_data(performed_job, message)
+      event = LogStash::Event.new(data)
+
+      RequestStore.clear!
+
+      # Is there a better way to make the logs appear on new lines??
```
SteveMarshall: This comment feels kinda pointless (unless it's a comment to reviewers); either remove it, or find the solution?

lib/logstash_sidekiq_logger/formatter.rb
```diff
@@ -0,0 +1,111 @@
...
+  private
+
+    def store_message_failure_metadata(message)
+      return if message !~ /^fail/
+
+      extract_failure_duration(message)
+    end
+
+    def extract_failure_duration(message)
+      return if message !~ /^fail/
```
SteveMarshall: This method appears to only be used in one place (`store_message_failure_metadata`, directly above), which *also* ensures that the message starts with `fail`. Maybe merge the two?

lib/logstash_sidekiq_logger/formatter.rb
```diff
@@ -0,0 +1,111 @@
...
+      RequestStore.clear!
+
+      # Is there a better way to make the logs appear on new lines??
+      "\n" + event.to_json
+    end
+
+  private
+
+    def store_message_failure_metadata(message)
+      return if message !~ /^fail/
```
SteveMarshall: While this says exactly what you intend, it might be a bit more explicit if you invert the `if` and `!~` to be `unless`/`=~`?

lib/logstash_sidekiq_logger/formatter.rb
```diff
@@ -0,0 +1,111 @@
...
+      data.merge!(failure_data(message)) unless data[:status] == 'completed'
+      data[:message] = log_message(data)
+      data
+    end
+
+    def log_message(data)
+      msg = "[#{data[:status]}] (#{data[:total_duration]} ms) "
+      msg += "#{data[:job_name]} args: #{data[:arguments]}"
+      msg
+    end
```
SteveMarshall: 😎

lib/logstash_sidekiq_logger/formatter.rb
```diff
@@ -0,0 +1,111 @@
...
+
+    def calculate_status(message)
+      if message.is_a?(Hash)
+        if message['retry']
+          'to_be_retried'
+        else
+          'failed'
+        end
+      else
+        'completed'
```
SteveMarshall: What dictates that a message that's not a Hash is complete? Is that guaranteed?

lib/logstash_sidekiq_logger/formatter.rb
```diff
@@ -0,0 +1,111 @@
...
+    def active_job_duration(job)
+      job.duration # milliseconds
+    end
+
+    def total_duration(message)
+      RequestStore.store[:duration] || extract_duration(message)
+    end
+
+    def extract_duration(message)
+      message.match(/\s(.+)\s/).captures.first.to_f * 1000 # milliseconds
```
SteveMarshall: This isn't clear: I *assume* it means “the value matched in the message is in milliseconds, so correct for that”, but it could also mean “convert the value in the message to milliseconds”.

lib/logstash_sidekiq_logger/formatter.rb
SteveMarshall: Oh, it's a Sidekiq thing (from your comment below); never mind.

SteveMarshall: Wait… but `message_needs_logging?` requires `message` to be a hash, so will anything ever actually get here?

SteveMarshall: Oh, never mind: I see that the logic there is broken over two lines 😵

SteveMarshall:
It'd be good if this PR and the commit within it followed the [alphagov git styleguide](https://github.com/alphagov/styleguides/blob/master/git.md).

Other than that, and a few bits in need of clarification, it looks good to me!

lib/logstash_sidekiq_logger.rb
alan: I am happy with your suggestion of using the method name.

My idea when writing this is that it could be extracted into a gem eventually and this description would become the Readme.

lib/logstash_sidekiq_logger/formatter.rb
alan: Good spot, I'll merge them

lib/logstash_sidekiq_logger/formatter.rb
alan: Means the later interpretation: “convert the value in the message to milliseconds”.

Would deleting the comment be less confusing?

lib/logstash_sidekiq_logger.rb
SteveMarshall: Yeah, I can see that that'd be useful… Maybe break the example over multiple lines and make the comment clear where the actual thing comes from? 😕

lib/logstash_sidekiq_logger/formatter.rb
SteveMarshall: I think having some comment is useful… Maybe move the comment to the line above and say something like `durations in message are in microseconds; convert to milliseconds`?

SteveMarshall:
💥

alan:
Thanks!

PR #412, opened 2016-06-27
Remove devise & implement SSO simply via omniauth

mloughran:
This is not ready to merge yet, and the tests are broken, but it does work.

@alan I think you'll agree it is vastly simpler than the layers of complexity introduced by devise + omniauth.


spec/controllers/prison/dashboards_controller_spec.rb
```diff
@@ -15,7 +15,7 @@
 
     context "when logged in" do
       before do
-        sign_in user
+        allow(controller).to receive(:current_user).and_return(user)
```
mloughran: @alan could we redefine `sign_in` to do this, rather than repeating `allow(controller).to receive(:current_user).and_return(user)` in a zillion places?

alan: We aren't signing in though, we are just stubbing out `current_user` so would prefer a different name. How about `stub_current_user` ?

mloughran:
This branch is now functionally complete, but tests need attention, and the code could be refactored. Refactoring can wait from my perspective.

PR #416, opened 2016-06-29
Report API errors more clearly

mloughran:
Same approach as used in prison-visits-public, but for calls into NOMIS.


StupidCodeFactory:
I'm closing this. We've already done some work around API errors + monitoring

PR #419, opened 2016-07-01
Content amends

chrishanes:


mloughran:
@chrishanes a number of these changes relate to email copy, but this PR only changes the copy in the staff UI (not in the mailers) – is that intentional? [p.s. you'd be right to question why the duplication]

PR #420, opened 2016-07-05
Show closed visits

chrishanes:
Show if a booking is a 'closed' visit on the visit status, print page and in the CSV file

### DIFF
--- a/app/services/booked_visits_csv_exporter.rb
+++ b/app/services/booked_visits_csv_exporter.rb
@@ -19,9 +19,10 @@ class BookedVisitsCsvExporter
 
   def headers
     [
-      'Prisoner name', 'Prisoner number', 'Slot granted', 'Lead visitor',
-      'Lead visitor dob', 'Lead visitor allowed?', 'Phone number',
-      'Email address', 'Visitor 2 name', 'Visitor 2 dob', 'Visitor 2 allowed?',
+      'Prisoner name', 'Prisoner number', 'Slot granted', 'Closed visit',
+      'Lead visitor', 'Lead visitor dob', 'Lead visitor allowed?',
+      'Phone number', 'Email address',
+      'Visitor 2 name', 'Visitor 2 dob', 'Visitor 2 allowed?',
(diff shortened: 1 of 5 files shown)


app/services/booked_visits_csv_exporter.rb
```diff
@@ -19,9 +19,10 @@ def to_csv
 
   def headers
     [
-      'Prisoner name', 'Prisoner number', 'Slot granted', 'Lead visitor',
-      'Lead visitor dob', 'Lead visitor allowed?', 'Phone number',
-      'Email address', 'Visitor 2 name', 'Visitor 2 dob', 'Visitor 2 allowed?',
+      'Prisoner name', 'Prisoner number', 'Slot granted', 'Closed visit',
```
alan: Can you add a spec (or modify) to the csv exporter for this change?

spec/services/booked_visits_csv_exporter_spec.rb
```diff
@@ -40,6 +40,7 @@
       expect(row['Prisoner number']).to eq(booked_visit.prisoner_number)
       expect(row['Slot granted']).
         to eq(instance.format_slot_for_staff(booked_visit.slot_granted))
+      expect(row['Closed visit']).to be_nil
```
alan: I think it would be better to keep it consistent with the visitor `_allowed?` methods and return `true` or `false` instead of `nil`

chrishanes: If it's not a closed visit it doesn't return false, just a nil value. An error is return if I use:
expect(row['Closed visit']).to eq('false')

app/views/prison/visits/show.html.erb
```diff
@@ -18,6 +18,9 @@
       <% if @visit.rejection %>
         <small class="font-xsmall">(<%= t(@visit.rejection_reason, scope: 'shared.rejection_reason') %>)</small>
       <% end %>
+      <% if (@visit.closed && @visit.processing_state == 'booked') %>
```
alan: this could be: 

```ruby
@visit.closed? && @visit.booked?
```

PR #424, opened 2016-07-11
Add a task to backfill cancellation reasons

alan:
We need to backpopulate cancellation reasons because we want to show the
cancellation reason in the Admin UI. Only recent cancellations created from the
10th of June have a cancellation reason. This means that if a user tries to view
an old cancelled visit without backfilling the cancellation reasons it would
raise an error.

Cancellation reasons were introduced on the 10th of June with the ability of
staff to cancel requests in PVB.

### DIFF
--- a/lib/tasks/one_off.rake
+++ b/lib/tasks/one_off.rake
@@ -138,4 +138,40 @@ namespace :pvb do
     FileUtils.rm(password_filename)
     STDOUT.puts "\nRemoved passwords."
   end
+
+  desc 'Backpopulate cancellation reasons'
+  task backpopulate_cancellation_reasons: :environment do
+    missing_cancellation = Visit.
+                           preload(:cancellation).
+                           with_processing_state('cancelled').
+                           joins(<<-EOS).
+        LEFT OUTER JOIN cancellations \
+          ON cancellations.visit_id = visits.id
+      EOS
+                           where(cancellations: { id: nil })
+
+    batch = missing_cancellation.limit(1000).to_a
+
+    while batch.any?
+      batch.each do |visit|
+        if visit.cancellation
+          fail "Visit #{visit.id} already has a cancellation!"
+        end
+
+        Cancellation.create!(visit_id: visit.id,
+                             reason: Cancellation::VISITOR_CANCELLED,
+                             created_at: visit.updated_at,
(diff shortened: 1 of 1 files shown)


lib/tasks/one_off.rake
```diff
@@ -138,4 +138,35 @@ namespace :pvb do
...
+    missing_cancellation = Visit.
+                           preload(:cancellation).
+                           with_processing_state('cancelled').
+                           joins(<<-EOS).
+        LEFT OUTER JOIN cancellations \
+          ON cancellations.visit_id = visits.id
+      EOS
+                           where(cancellations: { id: nil })
+
+    batch = missing_cancellation.limit(1).to_a
```
SteveMarshall: Why `limit(1)`?

SteveMarshall: Oh, right: because you're getting a new `batch` later… This seems confusing; what's the benefit over getting them all here?

lib/tasks/one_off.rake
```diff
@@ -138,4 +138,35 @@ namespace :pvb do
...
+        Cancellation.create!(visit_id: visit.id,
+                             reason: Cancellation::VISITOR_CANCELLED,
+                             created_at: visit.updated_at,
+                             updated_at: visit.updated_at)
+      end
+
+      batch = missing_cancellation.limit(1).to_a
+    end
+
+    STDOUT.puts "Done. Missing cancellation: #{missing_cancellation.count}"
```
SteveMarshall: This sounds like it's counted the number of items with missing cancellations, when in fact it's created cancellation objects for visits without, right? Maybe say `Created Cancellations for #{missing_cancellation.count} Visits`?

lib/tasks/one_off.rake
alan: My bad, I put `1` to test it locally, should be `1000` (which is the same default as Rails `.find_each` method).

We have about 14k records with missing cancellations, if I load them into memory all at once it might cause some problems with swapping.

lib/tasks/one_off.rake
alan: Mmmh, that will return `0` when it runs successfully. I want to check that there are no visits that have been missed. I'll tweak the message to make it clear.

PR #425, opened 2016-07-11
Add the cancelled reason to the visit details status

chrishanes:
This will show the reason for the cancellation including visits cancelled by the visitor

### DIFF
--- a/app/views/prison/visits/show.html.erb
+++ b/app/views/prison/visits/show.html.erb
@@ -16,10 +16,13 @@
     <h2 class="bold-medium">
       <%= @visit.processing_state.capitalize %>
       <% if @visit.rejection %>
-        <small class="font-xsmall">(<%= t(@visit.rejection_reason, scope: 'shared.rejection_reason') %>)</small>
+        <small class="font-xsmall">(<%= t(@visit.rejection_reason, scope: :shared) %>)</small>
       <% end %>
       <% if (@visit.closed? && @visit.booked?) %>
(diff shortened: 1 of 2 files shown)


app/views/prison/visits/show.html.erb
```diff
@@ -21,6 +21,9 @@
       <% if (@visit.closed? && @visit.booked?) %>
         <small class="font-xsmall">(<%= t('.closed_visit') %>)</small>
       <% end %>
+      <% if @visit.cancelled? %>
+        <small class="font-xsmall">(<%= t('rejection_reason.'+@visit.cancellation_reason, scope: :shared) %>)</small>
```
alan: You could use string interpolation instead of `+`.

Where is the other place where the context is shared?

alan:
Before merging this should be https://github.com/ministryofjustice/prison-visits-2/pull/424 being merged, deployed and run

app/views/prison/visits/show.html.erb
```diff
@@ -16,11 +16,14 @@
     <h2 class="bold-medium">
       <%= @visit.processing_state.capitalize %>
       <% if @visit.rejection %>
-        <small class="font-xsmall">(<%= t(@visit.rejection_reason, scope: 'shared.rejection_reason') %>)</small>
+        <small class="font-xsmall">(<%= t(@visit.rejection_reason, scope: 'shared') %>)</small>
```
alan: `'shared'` should be a symbol to keep it consistent

PR #426, opened 2016-07-11
Add actionable cancellations to the inbox

alan:
Adds cancellations that are not cancelled in Nomis to the top of the inbox.

### DIFF
--- a/app/views/prison/dashboards/inbox.html.erb
+++ b/app/views/prison/dashboards/inbox.html.erb
@@ -4,18 +4,69 @@
   <%= @estate.name %> <%= t('.title') %>
 <% end %>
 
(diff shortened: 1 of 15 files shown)


app/views/prison/dashboards/inbox.html.erb
```diff
@@ -4,10 +4,66 @@
...
+  </table>
+<% else %>
+  <p><%= t('.no_cancellations') %></p>
+<% end %>
+
 <%= render 'search_form', locals: {search_path: prison_inbox_path} %>
 
 <% if @requested_visits.any? %>
-
+  <h2 class="bold-medium">Requests</h2>
```
alan: Should this be in a locale file?

PR #428, opened 2016-07-12
Reuse logstasher json file for sidekiq

alan:
Starts using the same file that gets used to send data to logstash.

Tested in staging with the .log file and works as expected.


SteveMarshall:
Did you mean to add the `.env` file here?

alan:
Nope, I've taken it out.

PR #440, opened 2016-07-21
Add cancelled visits to print page and csv

chrishanes:


app/views/prison/dashboards/print_visits.html.erb
```diff
@@ -36,15 +34,13 @@
       <% end -%>
     <% end -%>
 
-    <h2>Cancelled visits</h2>
     <% if @data['cancelled'].nil? -%>
-      No cancellations
+      <p class="lede push-top">No cancellations</p>
```
alan: Should no bookings have the same styling?

PR #445, opened 2016-07-26
Add cancellation reason, timestamps to visit API

alan:
So that PVB Public can build a visit show page with the reason.


mloughran:
What are the timestamps used for?

alan:
We use `updated_at` to show when it was cancelled. I've added `created_at` as well as it would be weird only having `updated_at`

mloughran:
To decouple the API from the internals of PVB2, and to make this use explicit, it might be worth calling at attribute `cancelled_at` and only including it if the visit was cancelled. I'm not keen on just exposing the raw timestamps on the visit table, which would be fragile to changes in PVB's data model.

alan:
I like the idea of making explicit, although I would always add that field and serialise `null` to have well known responses, not a fan of returning fields conditionally

PR #450, opened 2016-07-26
Store user first and last names in the session

alan:


alan:
Changes are already in master

PR #451, opened 2016-07-26
Add an auto refresh to inbox page

chrishanes:
This is to prevent users trying to process a visit that has already been actioned by another user

### DIFF
--- a/app/assets/javascripts/modules/moj.autorefresh.js
+++ b/app/assets/javascripts/modules/moj.autorefresh.js
@@ -1,0 +1,43 @@
+(function () {
+  'use strict';
+
+  moj.Modules.AutoRefresh = {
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/modules/moj.autorefresh.js
```diff
@@ -0,0 +1,39 @@
...
+      this.$el = $(this.el);
+    },
+
+    render: function(){
+     var time = new Date().getTime();
+     $(document.body).bind("mousemove keypress", function(e) {
+       time = new Date().getTime();
+     });
+     function refresh() {
+       if(new Date().getTime() - time >= 180000){
```
alan: Can you add a comment saying how long is this timeout in seconds?

PR #465, opened 2016-08-11
Add global search and results page

chrishanes:


himal-mandalia:
Looks like this branch needs to be brought up to date with master first. Still mergeable.

PR #469, opened 2016-08-15
Add footer and header links to staff pages

chrishanes:


app/views/layouts/application.html.erb
```diff
@@ -58,9 +58,26 @@
...
-  </li>
+  <ul>
+    <li>
+      <%= link_to(t('.staff_guide'), '/staff/staffuiguide.md') %>
+    </li>
+    <li>
+      <%= link_to(t('.recent_changes'), '/staff/changes.md') %>
+    </li>
+    <li>
+      <%= link_to(t('.stats'), 'https://staff.prisonvisits.service.gov.uk/en/metrics') %>
```
alan: This wont work as expected in staging as it will linking to prod. Instead we can use a url helper. I think `metrics_path` should fix it

app/views/layouts/application.html.erb
```diff
@@ -58,9 +58,26 @@
...
+    </li>
+    <li>
+      <%= link_to(t('.contact_us'), 'https://www.prisonvisits.service.gov.uk/en/feedback/new?referrer=staff') %>
```
alan: Same here, how were we linking to the public feedback page before? This link will change soon anyway when we have the contact page.

chrishanes: The current staff index page links using the same url.

app/views/layouts/application.html.erb
chrishanes: I took the url from the staff info index page, but I'll add the url helper

PR #473, opened 2016-08-17
Staff contact page

alan:


app/views/prison/feedbacks/new.html.erb
```diff
@@ -0,0 +1,32 @@
...
+</div>
+
+<%= form_for @feedback, url: prison_feedbacks_path do |f| %>
+
+  <fieldset>
+    <%= f.hidden_field :referrer %>
+
+    <%= single_field f, :body, :text_area, cols: 50, rows: 5, class: 'form-control' %>
+
+    <%= single_field f, :email_address, :email_field, class: 'form-control' %>
```
mloughran: If the user is logged in then we will already know their email address from SSO – should this be used? Should staff be able to submit feedback if they're not logged in?

db/schema.rb
```diff
@@ -39,12 +39,23 @@
...
     t.string   "user_agent"
-    t.datetime "created_at",    null: false
-    t.datetime "updated_at",    null: false
+    t.datetime "created_at",                         null: false
+    t.datetime "updated_at",                         null: false
+    t.uuid     "prison_id"
+    t.boolean  "submitted_by_staff", default: false, null: false
+  end
+
+  create_table "messages_tables", force: :cascade do |t|
```
mloughran: Has this table beed committed in error?

app/controllers/prison/feedbacks_controller.rb
```diff
@@ -0,0 +1,28 @@
+class Prison::FeedbacksController < ApplicationController
+  before_action :authorize_prison_request
+  before_action :set_inbox_navigation_count, if: -> { current_user }
```
mloughran: I think we can avoid using a before action for this, and thereby cluttering up all the controllers with unrelated concerns.

app/views/prison/feedbacks/new.html.erb
alan: I think staff will be able to submit feedback even when logged out, it is part of the footer which is available on the login page.

db/schema.rb
alan: Yes, I'll take it out.

app/controllers/prison/feedbacks_controller.rb
alan: Yes, but only worth doing when the dashboards have been rolled out. The prison/visits controller for example loads the counts depending if they are in the trial. I think once it has been rolled out there'll be enough logic to create a dashboard controller super class that has all of the dashboards controller defaults.

mloughran: Good – as long as you're keeping this in mind.

app/views/prison/feedbacks/new.html.erb
mloughran: The login page is now a separate service so it can't be there. I think you mean the staff guide. Must say I'd be more comfortable long term putting this behind the login, but we can discuss.

PR #478, opened 2016-08-23
Inbox switch

alan:


chrishanes:
prison_switcher

The list of prisons should only appear on the inbox, processed and print visits lists and not in the visit details or process page.

app/models/signon_identity.rb
```diff
@@ -15,63 +15,66 @@ def from_omniauth(omniauth_auth)
...
 
   private
 
-    # rubocop:disable Metrics/MethodLength
     def find_or_create_authorized_user(info)
       email = info.fetch('email')
       permissions = info.fetch('permissions')
 
-      sso_orgs = permissions.map { |p| p.fetch('organisation') }
-      estates = Estate.where(sso_organisation_name: sso_orgs)
```
mloughran: What happens if one of a users permissions does not have a corresponding estate?

config/initializers/sso.rb
```diff
@@ -0,0 +1,2 @@
+Rails.configuration.sso_admin_org = ENV.
+fetch('sso_admin_org', 'pvb.digital.headquarters.moj')
```
mloughran: The default should be `digital.noms.moj` – because PVB is only available for teams within the `noms.moj` organisation.

mloughran: Honestly I'd be happy just hard-coding this in the code – it will change that rarely.

mloughran:
I suspect the code would be simpler if the current estate was added to the URL path, rather than storing in the session. I'm not entirely with the fact that SSOIdentity has now become mutable.

app/models/signon_identity.rb
```diff
@@ -85,7 +88,32 @@ def to_session
...
+      'available_organisations' => @available_organisations,
+      'current_organisation' => @current_organisation
     }
   end
+
+  def current_estate
+    available_estates.find do |estate|
+      estate.sso_organisation_name == @current_organisation
+    end
+  end
```
mloughran: If we take this approach we need to protect against current_estate being nil – which I think it can become right now.

alan:
Yes, would be simpler for the inbox switching but would require most of the urls so I would do it as a separate card to decide on the new url scheme and ensure that it doesn't break behaviour for existing users.

app/models/signon_identity.rb
```diff
@@ -15,63 +17,73 @@ def from_omniauth(omniauth_auth)
...
+    def full_name_from_additional_data(info)
+      first_name = info.fetch('first_name')
+      last_name = info.fetch('last_name')
+      [first_name, last_name].reject(&:empty?).join(' ')
+    end
+
+    def valid_pvb_orgs(info)
+      sso_orgs = info.fetch('permissions').map { |p| p.fetch('organisation') }
+      pvb_orgs = Estate.pluck(:sso_organisation_name)
+      sso_orgs.select { |org| pvb_orgs.include?(org) }
```
mloughran: Is it cleaner to write `Estate.where(sso_organisation_name: sso_orgs)` ?

app/models/signon_identity.rb
```diff
@@ -85,7 +97,32 @@ def to_session
...
+    available_estates.find do |estate|
+      estate.sso_organisation_name == @current_organisation
+    end
+  end
+
+  def available_estates
+    @available_estates ||=
+      begin
+        if @available_organisations.include?(DIGITAL_ORG)
+          Estate.all
```
mloughran: Is there a bug here? By my logic `sso_orgs` will not contain `DIGITAL_ORG` because there's no matching estate, which means that the `available_estates` instance method will never match `if @available_organisations.include?(DIGITAL_ORG)`.

In any case I must say I find this code extremely confusing. I think we can find a clearer way of encapsulating the logic of which orgs are made available.

app/models/signon_identity.rb
```diff
@@ -15,63 +17,73 @@ def from_omniauth(omniauth_auth)
...
+      sso_orgs = valid_pvb_orgs(info)
 
       {
-        full_name: full_name,
+        full_name: full_name_from_additional_data(info),
         profile_url: links.fetch('profile'),
-        logout_url: links.fetch('logout')
+        logout_url: links.fetch('logout'),
+        available_organisations: sso_orgs,
+        current_organisation: sso_orgs.first
```
mloughran: This could do with a comment explaining that we default to the first organisation.

I'm assuming this is what we want? That if they log out and back in again their default estate is reset?

app/models/signon_identity.rb
```diff
@@ -33,7 +33,7 @@ def from_session_data(data)
 
     def find_or_create_authorized_user(info)
       email = info.fetch('email')
-      sso_orgs = valid_pvb_orgs(info)
+      sso_orgs = info.fetch('permissions').map { |p| p.fetch('organisation') }
 
       unless sso_orgs.any?
```
alan: The overall validation for the estate is that the signon class checks that there is at least an organisation and a new controller before action checks that the current estate is valid: https://github.com/ministryofjustice/prison-visits-2/pull/478/commits/732f575e3307beb210cadebdc830bd8a6a22c9c4#diff-55c5b7aecfb519d0e4880eaf2788eb6eR67

app/models/signon_identity.rb
```diff
@@ -103,26 +102,7 @@ def to_session
...
-          Estate.where(sso_organisation_name: @available_organisations)
-        end.to_a
-      end
-  end
-
-  def change_current_estate(new_estate_id)
-    @current_organisation = available_estates.find { |estate|
-      estate.id == new_estate_id
-    }.sso_organisation_name
+  def change_current_organisation(sso_org)
```
alan: Decided to keep the current organisation logic here until we do the url changes

app/controllers/application_controller.rb
```diff
@@ -10,13 +10,29 @@ class ApplicationController < ActionController::Base
...
+    available_estates.find do |estate|
+      estate.sso_organisation_name == sso_identity&.current_organisation
+    end
+  end
+
+  def available_estates
+    @available_estates ||=
+      if sso_identity.
+         available_organisations.
+         include?(SignonIdentity::DIGITAL_ORG)
```
alan: Couldn't think of a better place to put the the `DIGITAL_ORG` constant, any ideas?

mloughran:
@alan I've refactored to my taste (critically making SignonIdentity immutable, which makes it simpler, and it shouldn't need to change if we move the estate into the path). There's no need to validate the estate, because a default is always available. Also, by storing the estate separately to the identity information, the user's default estate can persist across logins.

If this makes sense to you I'll address the tests, which are currently broken.

app/controllers/application_controller.rb
```diff
@@ -10,11 +10,24 @@ class ApplicationController < ActionController::Base
...
+  helper_method :current_estate
 
   def current_user
     sso_identity&.user
   end
 
+  def current_estate
+    return unless sso_identity
+    @_current_estate ||= begin
+      estate_id = session[:current_estate]
```
alan: This could result in a 404 which would be very confusing to the user and when debugging if the current_estate is removed from the user sso settings and then the user logs in again.

app/controllers/prison/dashboards_controller.rb
```diff
@@ -35,23 +35,36 @@ def search
     cancellations
     processed_visits
 
-    @estate = user_estate
+    @estate = current_estate
+  end
+
+  def switch_estate
+    estate = Estate.find(params[:estate_id])
```
alan: Again, this could result in a 404 which would be confusing to the user and when debugging if the estate_id is not valid.

app/models/signon_identity.rb
```diff
@@ -79,13 +88,26 @@ def logout_url(redirect_to: nil)
...
+  def accessible_estates
+    @_ae ||= self.class.accessible_estates(@permissions).order(:nomis_id).to_a
+  end
+
+  def accessible_estate?(estate)
+    accessible_estates.include?(estate)
+  end
+
+  def default_estate
+    accessible_estates.first || fail('Should never be nil')
```
alan: This could be nil if SSO gets set up with a new organisation with access to PVB that has not yet been setup in PVB. With the current version if this happens the user will get a 500

alan:
Have you tried to see if the current estate persists between sessions for the same user? I would have thought it wouldn't work because the second time the user logs in it will be a different session.

There is still a few edge cases to deal with with your approach, mainly around better handling of unwanted 404's and 500s.

Regarding the overall approach I don't like having current estate stored separately from the rest of the of the session data, so overall I prefer my approach. From what I remember it handled the edge cases better.

app/models/signon_identity.rb
mloughran: It won't be nil in that case because the `self.class.accessible_estates` method is called on login and validated not to be empty. So in that case the user won't be logged in.

There is an edge case if an estate is removed from PVB after the user has logged in, but removing estates is not something we've ever done. I raised an exception so that the issue would be clear if it ever happened, rather than passing nils elsewhere. It could be handled properly but I'm not sure it's worthwhile.

mloughran:
I have validated that the estate is persisted. When then user logs out the session isn't cleared – only the `sso_data` key is deleted.

I'll take a proper look at the edge cases, but I think the approach is sound.

app/controllers/application_controller.rb
mloughran: I think you're right – I need to additionally check that the estate is still accessible.

app/controllers/prison/dashboards_controller.rb
mloughran: How would this happen in practice? Where do the non-existing ids come from?

app/controllers/prison/dashboards_controller.rb
```diff
@@ -35,23 +35,36 @@ def search
...
+
+    if sso_identity.accessible_estate?(estate)
+      session[:current_estate] = estate.id
+    else
+      # This should never happen
+      fail 'User attempted to switch to inaccessible estate'
```
mloughran: Will change this to a `flash` notice.

app/controllers/application_controller.rb
mloughran: The additional check is in place

app/controllers/prison/dashboards_controller.rb
mloughran: I decided not to worry about this.

However switching to a now inaccessible estate is handled more gracefully (only possible via an out of date browser tab) – see below.

app/models/user.rb
```diff
@@ -1,3 +1,8 @@
 class User < ActiveRecord::Base
   belongs_to :estate
+
+  # TODO: Delete me when the column has dropped
+  def self.columns
+    super.reject { |c| c.name == 'estate_id' }
+  end
```
mloughran: @alan why is this necessary? Can the column be dropped as a part of this PR?

spec/support/helpers/controller_helper.rb
```diff
@@ -1,6 +1,16 @@
 module ControllerHelper
-  def stub_logged_in_user(user)
-    identity = SignonIdentity.new(user, full_name: FFaker::Name.name, profile_url: '', logout_url: '')
-    allow(controller).to receive(:sso_identity).and_return(identity)
+  def stub_logged_in_user(user, estate, available_estates: [estate])
```
mloughran: Being pedantic, this method is no longer accurately named. It doesn't stub the login, it actually logs in the user.

app/models/user.rb
alan: This is a well known pattern to avoid downtime when a column needs to be dropped.

When the migration runs to drop the column and the app servers haven't yet been restarted AR still tries to write to the column because it uses prepared statements. To avoid this short downtime the first step is to tell ActiveRecord to ignore the column and deploy, then drop the column.

app/controllers/prison/dashboards_controller.rb
alan: Which part below?

PR #479, opened 2016-08-23
Add staff messages

chrishanes:


app/services/booking_responder.rb
```diff
@@ -84,4 +86,17 @@ def mark_banned_visitors
...
     end
   end
+
+  def create_message
+    return nil if booking_response.message_body.blank?
+
+    Message.create!(body: booking_response.message_body,
+                    user: booking_response.user,
+                    visit: visit,
+                    visit_state_change: visit.last_visit_state)
```
mloughran: This is a bit nasty. Although I can see that currently you're calling this method inside a transaction which also changes the visit state, this won't always be the case when creating messages, and seems rather non-obvious. Can it be more explicit?

I still think it would be cleaner if acceptance or rejections were first class objects which may have associated messages, and associated state changes – rather than tying the message to the state change directly.

app/mailers/visitor_mailer.rb
```diff
@@ -15,12 +15,16 @@ def request_acknowledged(visit)
 
   def booked(visit)
     I18n.locale = visit.locale
+    @message = visit.acceptance_message
```
mloughran: For example if the acceptance was a first-class event, it could be passed here – rather than the visit, which does a rather horrible lookup of acceptance message via its state change.

mloughran:
I think we can improve the way we represent messages in the domain, but happy to merge this and consider refactoring as part of introducing the timeline.

app/services/booking_responder.rb
alan: When creating a message within the booking responder it is always related to the last visit state change. Other messages that aren't created via the booking responder won't be using this method, to me it is pretty obvious so don't know what to change it to. Should we leave it as it is and wait for the events work?

mloughran: Yes, I had missed the fact that this was the booking responder class. I would be happier if the state change was passed into the method, but this is definitely less of an issue than I first claimed.

alan: Updated to pass in the visit state change

PR #488, opened 2016-09-05
Validate and sanitise visit phone number

alan:
It ensures that the phone number is valid, previous validation on PVB2 was that it had to be present, and in Public that it had be of minimum length 9.

This PR should be deployed after https://github.com/ministryofjustice/prison-visits-public/pull/72


alan:
Reviewed by:
- [ ] @kyliemulholland 
- [ ] @mloughran

PR #500, opened 2016-09-19
fix invalid locale exception

StupidCodeFactory:
**CHANGES:**
Check if the locale received through the request is available otherwise
set it to the default locale

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -78,7 +78,12 @@ private
   end
 
   def set_locale
-    I18n.locale = params.fetch(:locale, I18n.default_locale)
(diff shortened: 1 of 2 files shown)


app/controllers/application_controller.rb
```diff
@@ -78,7 +78,12 @@ def default_url_options(*)
   end
 
   def set_locale
-    I18n.locale = params.fetch(:locale, I18n.default_locale)
+    locale = params[:locale]
+    I18n.locale = if locale && I18n.available_locales.include?(locale)
+                    params[:locale]
```
alan: Could we use the `locale` variable here?

PR #499, opened 2016-09-19
Add APVS survey link to visitor email

kyliemulholland:
I copied a closed PR:
https://github.com/ministryofjustice/prison-visits-2/pull/292/files

PR #504, opened 2016-09-21
Add APVS survey to booked emails

kyliemulholland:
Relates to https://github.com/ministryofjustice/prison-visits-2/pull/499/files


app/views/visitor_mailer/booked.html.erb
```diff
@@ -97,3 +97,10 @@
 <p>
   <small><%= t(".visit_id") %> <%= @visit.id %></small>
 </p>
+</p>
+<div class="apvs_survey">
+   <p><strong><%= t('.survey_question1') %></strong></p>
```
alan: The locales wont work because it will try to find the text under the `booked` section of the locales. I think this should work:

```ruby
t('survey_question1', scope: %i(visitor_mailer request_acknowledged))
```

PR #506, opened 2016-09-21
Authentication for NOMIS API

mloughran:
This PR also rolls in my earlier (unmerged) PR to improve API errors (#416).

### DIFF
--- a/app/services/nomis/client.rb
+++ b/app/services/nomis/client.rb
@@ -1,6 +1,8 @@
 require 'excon'
 
 module Nomis
+  APIError = Class.new(StandardError)
+
   class Client
     TIMEOUT = 1 # seconds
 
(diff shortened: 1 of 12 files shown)


app/services/nomis/client.rb
```diff
@@ -64,5 +87,20 @@ def params_options(method, params)
         #   }
       end
     end
+
+    def auth_header
+      return unless @client_token && @client_key
+
+      token = auth_token(@client_token, @client_key)
```
StupidCodeFactory: Since the ```@client_token``` and ```@client_key``` are already set, there is need to pass them as arguments to ```#auth_token``` they can be directly accessed from within the ```#auth_token``` method.

mloughran: Definitely a reasonable point, however I prefer to make it explicit that `auth_token` is a function of the client token and key. Does that make sense?

PR #520, opened 2016-09-26
Process page ui layout change

StupidCodeFactory:
Reverts ministryofjustice/prison-visits-2#519


app/views/prison/visits/_issue_with_prisoner_section.html.erb
```diff
@@ -1,106 +1,142 @@
...
+        </div>
+        <%=
+          single_field(
+            f, :privileged_allowance_available, :check_box,
+            class: 'js-Conditional',
+            data: { 'conditional-el': 'privileged_allowance_available_details' }
+          )
+        %>
+        <div id="privileged_allowance_available_details" class="form-group">
+          <%= composite_field(f, :privileged_allowance_expires_on) do %>
```
StupidCodeFactory: I've modified this field name it use to be:
```ruby
<%= composite_field(f, :privileged_allowance_available_on) do %>
```
which is incorrect as the attribute is ```privileged_allowance_expires_on``` and was causing the validation errors not to appear.
I'm not sure if this could have any consequences elsewhere. @chrishanes ?

PR #532, opened 2016-09-30
Email preview

StupidCodeFactory:
**NOTES:**
Content to of emails should be thoroughly checked again

### DIFF
--- a/app/models/visit.rb
+++ b/app/models/visit.rb
@@ -143,11 +143,19 @@ cancellations.id IS NULL OR cancellations.nomis_cancelled = :nomis_cancelled
   end
 
   def banned_visitors
-    visitors.banned
+    if visitors.loaded?
+      visitors.select(&:banned)
+    else
+      visitors.banned
+    end
   end
 
   def unlisted_visitors
-    visitors.unlisted
+    if visitors.loaded?
+      visitors.select(&:not_on_list)
+    else
+      visitors.unlisted
+    end
   end
 
   def acceptance_message
--- a/app/models/visitor.rb
+++ b/app/models/visitor.rb
@@ -4,6 +4,7 @@ class Visitor < ActiveRecord::Base
 
   scope :banned,   -> { where(banned: true) }
   scope :unlisted, -> { where(not_on_list: true) }
+  scope :allowed,  -> { where(banned: false, not_on_list: false) }
 
   belongs_to :visit
   validates :visit, :sort_index, presence: true
@@ -20,9 +21,5 @@ class Visitor < ActiveRecord::Base
     return 'banned' if banned?
     return 'not on list' if not_on_list?
     'allowed'
-  end
-
-  def self.allowed
-    where(banned: false, not_on_list: false)
   end
 end
--- a/app/models/visitor.rb
+++ b/app/models/visitor.rb
@@ -2,6 +2,9 @@ class Visitor < ActiveRecord::Base
   include Person
   extend FreshnessCalculations
 
+  scope :banned,   -> { where(banned: true) }
+  scope :unlisted, -> { where(not_on_list: true) }
+
   belongs_to :visit
   validates :visit, :sort_index, presence: true
 
@@ -21,13 +24,5 @@ class Visitor < ActiveRecord::Base
 
   def self.allowed
     where(banned: false, not_on_list: false)
-  end
-
-  def self.banned
-    where(banned: true)
-  end
-
-  def self.unlisted
-    where(not_on_list: true)
   end
 end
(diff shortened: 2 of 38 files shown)


app/models/visit.rb
```diff
@@ -9,6 +9,14 @@ class Visit < ActiveRecord::Base
   has_one :rejection, dependent: :destroy
   has_one :cancellation, dependent: :destroy
 
+  has_many :banned_visitors, lambda {
```
alan: How about keeping `visitors` as the main association as all of the visitors need to be loaded always and changing the methods to something like this:

```ruby
def banned_visitors
  if visitors.loaded?
    visitors.select(&:banned?)
  else
    visitors.banned
  end
end
```

That way when we eagerload the visitors on a visit the methods will use the eager loaded data instead of triggering more DB calls

app/models/visit.rb
```diff
@@ -9,6 +9,14 @@ class Visit < ActiveRecord::Base
...
+    where(visitors: { banned: true })
+  }, class_name: 'Visitor'
+
+  has_many :unlisted_visitors, lambda {
```
alan: Same as above

app/services/booking_responder/booking_request_processor.rb
```diff
@@ -0,0 +1,57 @@
...
+      visit_state_change.update!(processed_by: booking_response.user)
+    end
+
+    def mark_disallowed_visitors
+      mark_unlisted_visitors
+      mark_banned_visitors
+    end
+
+    def mark_unlisted_visitors
+      booking_response.unlisted_visitors.each do |visitor|
```
alan: Potential minor improvement could be to batch update banned / not on list visitors in a single DB query.

app/views/visitor_mailer/booked.html.erb
```diff
@@ -1,22 +1,22 @@
...
 </p>
 
 <p>
   <%= t('.number') %>
-  <strong><%= @visit.prisoner_number %></strong>
+  <strong><%= @booking_response.prisoner_number %></strong>
 </p>
 
-<% @visit.visitors.allowed.each_with_index do |visitor, index| %>
+<% @booking_response.visitors.allowed.each_with_index do |visitor, index| %>
```
alan: This will be wrong if some visitors are rejected as it will show all of the visitors. We need a method on `BookingResponse` that returns only the allowed visitors

config/routes.rb
```diff
@@ -46,6 +46,10 @@
         as: :confirm_nomis_cancellation_visit
     end
 
+    resources :visits, only: [] do
+      resource :preview, only: :show
```
alan: Maybe 'email_preview' ? Not really obvious otherwise what 'preview' might be

spec/features/process_a_request_spec.rb
```diff
@@ -71,6 +71,18 @@
...
 
+      preview_window = window_opened_by {
+        click_link 'Preview Email'
+      }
+
+      within_window preview_window do
+        expect(page).to have_css('p', text: /Dear #{vst.visitor_full_name}/)
+      end
+
+      # switch window
```
alan: ✂️ commented out?

mloughran:
Email preview looks good. Wondering how hard it would be to make this available from the visit show page in case someone wants to view it retrospectively?

Also noticed a bug: it's possible to enter dates in the past for the date when the VO will renew, but the email only contains the day & month, not year. I assume this should be a validation error? Why are we asking for the year if it's not used? /cc @chrishanes @kyliemulholland

StupidCodeFactory:
@mloughran not very hard i think

kyliemulholland:
@mloughran It shouldn't be possible to enter dates in the past. We should add the year, as especially around Christmas this will be really important

app/models/visitor.rb
```diff
@@ -2,6 +2,13 @@ class Visitor < ActiveRecord::Base
   include Person
   extend FreshnessCalculations
 
+  scope :banned,   -> { where(banned: true) }
+  scope :unlisted, -> { where(not_on_list: true) }
+
+  def self.unlisted
```
alan: Can we delete this method? It is also a scope

app/views/prison_mailer/booked.html.erb
```diff
@@ -5,35 +5,35 @@
...
   <tr>
     <th valign="top" align="left">Prisoner number</th>
-    <td><%= @visit.prisoner_number %></td>
+    <td><%= @booking_response.prisoner_number %></td>
   </tr>
   <tr>
     <th valign="top" align="left">Approved visitors</th>
     <td>
-      <% @visit.visitors.allowed.each do |visitor| %>
+      <% @booking_response.visitors.allowed.each do |visitor| %>
```
alan: We still need to update this to use the in memory data from the booking response instead of loading the allowed visitors from the database. In a preview the allowed visitors from the database will always be all the visitors

PR #535, opened 2016-10-04
Improve inbox cancellations

alan:
- Fix the date to show the actual cancellation date
- Show the reference number of the booking when expanding the cancellation
- Show which visitors are banned or not on the list

### DIFF
--- a/app/models/visit.rb
+++ b/app/models/visit.rb
@@ -164,6 +164,10 @@ cancellations.id IS NULL OR cancellations.nomis_cancelled = :nomis_cancelled
     visit_state_changes.order('created_at desc').first
   end
 
+  def additional_visitors
+    visitors.reject { |v| v == principal_visitor }
+  end
+
 private
 
   def cancellation!(reason, nomis_cancelled:)
--- a/app/models/visit.rb
+++ b/app/models/visit.rb
@@ -165,7 +165,7 @@ cancellations.id IS NULL OR cancellations.nomis_cancelled = :nomis_cancelled
   end
 
   def additional_visitors
-    visitors.reject { |v| v == principal_visitor }
+    @additional_visitors ||= visitors.reject { |v| v == principal_visitor }
   end
 
 private
--- a/app/views/prison/dashboards/_request_details_row.html.erb
+++ b/app/views/prison/dashboards/_request_details_row.html.erb
@@ -1,9 +1,10 @@
 <tr class="hidden-row hide" aria-hidden="true" aria-expanded="false">
   <td colspan="9" class="request-details">
     <div class="request-row-details">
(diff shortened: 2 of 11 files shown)


app/views/prison/dashboards/_request_details_row.html.erb
```diff
@@ -1,24 +1,17 @@
 <tr class="hidden-row hide" aria-hidden="true" aria-expanded="false">
   <td colspan="9" class="request-details">
     <div class="request-row-details">
-      <% if visit.slot_granted? -%>
+      <% visit.slots.each_with_index do |slot, i| %>
         <dl>
-          <dt><%= t('.booked_date') %></dt>
-          <dd><%= format_visit_slot_date_for_staff(visit) %><br><%= format_visit_slot_times_for_staff(visit) %></dd>
+          <dt><%= t('.choice', number: ActiveSupport::Inflector.ordinalize(i+1)) %></dt>
```
StupidCodeFactory: probably can do the following: 
```ruby
(i+1).ordinalize
```

app/views/prison/dashboards/_request_details_row.html.erb
```diff
@@ -1,24 +1,17 @@
 <tr class="hidden-row hide" aria-hidden="true" aria-expanded="false">
   <td colspan="9" class="request-details">
     <div class="request-row-details">
-      <% if visit.slot_granted? -%>
+      <% visit.slots.each_with_index do |slot, i| %>
```
StupidCodeFactory: What about collection rendering? Then shoving the inside of the loop in a partial:
```ruby
<%= render visit.slots %>
```

app/models/visit.rb
```diff
@@ -164,6 +164,10 @@ def last_visit_state
     visit_state_changes.order('created_at desc').first
   end
 
+  def additional_visitors
+    visitors.reject { |v| v == principal_visitor }
```
StupidCodeFactory: i'd memoize this but it's fine as it is

app/views/prison/dashboards/_request_details_row.html.erb
alan: `ConcreteSlot` is not an ActiveModel so doesn't implement `#to_partial_path` for this to work. I'm tempted to leave it as it is as the partial won't be reused anywhere else. What do you think?

app/views/prison/dashboards/_request_details_row.html.erb
alan: I've updated it in the last commit

app/models/visit.rb
alan: Good spot, I've updated it

PR #543, opened 2016-10-06
Disable more emails

alan:
- Delayed prisons
- South West
- South Central
- North West


kyliemulholland:
Need to add Whitemoor, Wolds and Everthorpe

PR #546, opened 2016-10-10
multiple rejection reasons

StupidCodeFactory:
**CHANGES:**
This is the first step towards allowing multiple rejection reasons:
* mirror the ```reason``` fields with the ```reasons`` `
* backfill previous reason

### DIFF
--- a/db/migrate/20161010113526_migrate_reason_to_reasons.rb
+++ b/db/migrate/20161010113526_migrate_reason_to_reasons.rb
@@ -1,7 +1,0 @@
-class MigrateReasonToReasons < ActiveRecord::Migration
-  def change
-    Rejection::REASONS.each do |reason|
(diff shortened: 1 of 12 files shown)


db/migrate/20161010113526_migrate_reason_to_reasons.rb
```diff
@@ -0,0 +1,7 @@
+class MigrateReasonToReasons < ActiveRecord::Migration
+  def change
+    Rejection::REASONS.each do |reason|
```
alan: I think it would be safer to put this on a rake task because of the type of postgres locks that migrations use. Unless we deploy this out of hours when we are sure that staff aren't using the dashboards.

It would also be worth checking how big the batches are as updating large batches could cause load on the db to go up. If the batches can be very large then we could update it in batches of 10k with sleeps in between.

StupidCodeFactory:
let me extract that migration in a rake task, you eroded my confidence the migration will be fast ;)

PR #555, opened 2016-10-11
New visit show page

alan:
Needs email copy ~~and UI polish~~


app/controllers/prison/messages_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+    Visit.joins(prison: :estate).
+      where(estates: { id: current_estate }).
+      find(params[:visit_id])
+  end
+
+  def message_params
+    params.require(:message).permit(:body).
+      merge(user: current_user, visit: @visit)
+  end
+end
```
StupidCodeFactory: There is a tricky bit here, ```load_visit``` needs to be called before ```message_params```, one could easily be trapped here.
I believe the following would be clearer:

```ruby
class Prison::MessagesController < ApplicationController
  before_action :authorize_prison_request

  def create
    load_visit
    @message = Message.create_and_send_email(message_params)

    if @message.persisted?
      flash[:notice] = t('message_created', scope: [:prison, :flash])
      redirect_to :back
    else
      flash[:notice] = t('message_create_error', scope: [:prison, :flash])
      render 'prison/visits/show'
    end
  end

private

  def visit
    @visit ||= Visit.joins(prison: :estate).
      where(estates: { id: current_estate }).
      find(params[:visit_id])
  end
  alias :load_visit :visit

  def message_params
    params.require(:message).permit(:body).
      merge(user: current_user, visit: visit)
  end
end
```
I usually wouldn't bother with aliasing ```load_visit``` to ```visit``` but it's a bit more explicit.

app/controllers/prison/visits_controller.rb
```diff
@@ -72,4 +80,30 @@ def visit_page(visit)
       prison_deprecated_visit_path(visit)
     end
   end
+
+  def load_visit
```
StupidCodeFactory: Same pattern as above. Memoizing the scope an renaming/aliasing it to visit. It would also read a lot better line 51 and 8

StupidCodeFactory: Also all these methods are now in a booking response context:

https://github.com/ministryofjustice/prison-visits-2/blob/e8552b47a2e1e04a4c58e5bdd62947c8f6979521/app/controllers/concerns/booking_response_context.rb

app/controllers/prison/visits_controller.rb
```diff
@@ -57,8 +62,12 @@ def cancel
 private
 
   def part_of_trial?
-    estate = unscoped_visit.prison.estate
-    estate.name.in?(Rails.configuration.dashboard_trial)
+    estate_name = Estate.
```
StupidCodeFactory: Don't think this method belongs to the controller.

app/controllers/prison/visits_controller.rb
```diff
@@ -27,23 +27,28 @@ def update
...
 
   def show
-    @visit = scoped_visit
-    @message = @visit.messages.first
+    @visit = scoped_visit.
+             includes(
+               :visitors,
+               messages: :user,
+               visit_state_changes: :processed_by).
+             find(params[:id])
```
StupidCodeFactory: That's a lot of SQL/ActiveRecord for a controller. Perhaps a presenter layer would improve testability and readability of this controller.

app/models/message.rb
```diff
@@ -5,4 +5,11 @@ class Message < ActiveRecord::Base
 
   validates :body, presence: true
   validates :user_id, presence: true
+
+  def self.create_and_send_email(attrs)
+    message = new(attrs)
+
+    VisitorMailer.one_off_message(message).deliver_later if message.save
+    message
+  end
```
StupidCodeFactory: I'd create a service object to do just that. I'm not a big fan of coupling the database layer and email sending. But this is simple enough, i'm not too fussed if stays this way.

spec/services/visit_timeline_spec.rb
```diff
@@ -0,0 +1,57 @@
+require 'rails_helper'
+
+RSpec.describe VisitTimeline do
+  let(:instance) { described_class.new(visit) }
```
StupidCodeFactory: that's the subject of the test

spec/services/visit_timeline_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+  let(:visit) { FactoryGirl.create(:visit) }
+
+  describe '#events' do
+    subject(:events) { instance.events }
```
StupidCodeFactory: Why not just ```let```? also a collection of events isn't the subject being tested.

alan: In my view `subject` doesn't have to be exclusively the instance of the objects, it can also be the method being described. In this case what's under test is the collection of events. I somewhat follow http://betterspecs.org/#subject with the use of subject

app/controllers/prison/visits_controller.rb
alan: I fixed the bad merge, I missed the BookingResponseContext. Let's pair on the presenter idea

app/controllers/prison/visits_controller.rb
StupidCodeFactory: agreed

alan:
@chrishanes is this ready for final review?

chrishanes:
@alan I believe this is ready now for final review, although I think we should wait for Andreea to user test today and tomorrow in case we need to make some changes.

PR #561, opened 2016-10-12
Multiple rejection reasons form

StupidCodeFactory:


app/views/prison/visits/_issue_with_prisoner_section.html.erb
```diff
@@ -4,10 +4,14 @@
       <%= error_container(f, :selection) do %>
         <label class="block-label">
           <%=
-            f.radio_button(
-              :selection, :no_allowance,
-              class: 'js-Conditional',
-              data: { 'conditional-el': 'no_allowance_details' }
+            f.check_box(
```
alan: I think we need to preserve the checked boxes when the page gets re-rendered after a validation error, at the moment if a validation error happens any checkboxes that the user has selected will be lost. I haven't tried but I think that will be the case...

app/models/booking_response.rb
```diff
@@ -168,4 +167,26 @@ def not_allowed_visitor_ids
     @not_allowed_visitor_ids ||=
       unlisted_visitor_ids + banned_visitor_ids
   end
+
+  def no_banned_or_unlisted_visitors?
```
alan: I agree, using the validation callbacks to calculate additional rejection reasons is a bit strange and is a smell in this class. I think we can decouple the calculation of automated rejection reasons from the class into a different class perhaps.

app/models/rejection.rb
```diff
@@ -9,16 +11,20 @@ class Rejection < ActiveRecord::Base
...
     'duplicate_visit_request'
   ].freeze
 
   belongs_to :visit
 
-  validates :reason, inclusion: { in: REASONS }
   validate :validate_reasons
 
+  # TODO: Delete me when the column has dropped
+  def self.columns
```
alan: The review app script to populate dummy data is broken and is preventing the deploy of the review app for the PR

app/models/booking_response.rb
```diff
@@ -1,173 +1,133 @@
...
-  attribute :selection, String, default: 'none'
-  validates :selection,
-    inclusion: { in: SLOTS + Rejection::REASONS },
-    if: :at_least_one_valid_visitor?
-
-  attribute :reference_no, String
-  validates :reference_no, presence: true, if: :bookable?
-  attribute :closed_visit, Virtus::Attribute::Boolean
+  validate :validate_visit_is_processable
+  validate :visit_or_rejection_validity
```
alan: This is not quite working, I've tried selecting a slot and a rejection and the visit was processed

alan: This would be another validation

app/controllers/prison/email_previews_controller.rb
```diff
@@ -4,7 +4,11 @@ class Prison::EmailPreviewsController < ApplicationController
   include BookingResponseContext
 
   def show
-    render html: email_preview
+    if booking_response.valid?
+      render html: email_preview
+    else
+      render text: booking_response.errors.full_messages
```
alan: Ignore, I thought this was non JS, probably worth thinking if we can render a styled view

app/controllers/prison/visits_controller.rb
```diff
@@ -42,8 +46,7 @@ def show
                :visitors,
                messages: :user,
                visit_state_changes: :processed_by).
-             find(load_visit.id)
-    @nomis_checker = StaffNomisChecker.new(@visit)
+             find(load_visit.id).decorate
```
alan: Does the show page use decorators now?

app/services/booking_responder/reject.rb
```diff
@@ -1,44 +1,9 @@
 class BookingResponder
```
alan: When the visitors are banned / not on list the rejections reasons are: not enough adults, visitor banned and visitor not on list. We need to rethink what the rejection reasons should be, one to discuss with @kyliemulholland

alan: When rejecting all the visitors and selecting a slot, the slot granted persists in the DB even thought the visit is rejected

alan:
Found an inconsistency with master:
- user selects no VO allowance
- then selects one of the checkboxes that reveal the fields for the renewal date
- this branch: no VO allowance is deselected (in this branch) and select a date / in master a date is selected as booked
- Submit form

In this branch the visit is booked, in master a validation error occurs on the inner checkbox as no date was entered for the renewal.

I think that master validates this case because the user might have selected / deselected an option by mistake.

The behaviour in this branch while not the same might be ok... Thoughts @kyliemulholland @StupidCodeFactory ?

alan:
Decided that the renewal date shouldn't be validated if the no vo's option is deselected, which does happen when another rejection reason is selected

alan:
* When the renewal date is invalid, the value is lost when the page is rendered again
* User overrides nomis validation failure, page is rendered again after a validation failure (ie invalid date on vo renewal) - The nomis validation is checked again, expected to remember the user override.
* Error happens when selecting no vo alloance with an invalid date and also selecting a date - `StateMachines::InvalidTransition`

PR #564, opened 2016-10-13
Design iteration

chrishanes:
This will apply design amends based on designer feedback, mainly around the process page.


chrishanes:
Before
before

chrishanes:
After
after

mloughran:
Generally looks good to me.

Is it potentially confusing that "the prisoner date of birth... have been verified" is included in the warning?

chrishanes:
The verified content and warning isn't part of this PR, but I've seen at Brixton that the whole 'Verified' concept is confusing as they didn't know who or what has verified the info. Something to pick up with our content editor Shane.

kyliemulholland:
@chrishanes looks good to me. Yes agreed, let's look at content for the validation errors separately. See https://trello.com/c/1dwdLFmU/380-prisoner-validation-staff-3

PR #589, opened 2016-11-07
Records who cancels and withdraws a visit

alan:
Follows the same schema as booking / rejecting visits by using responders and
processors.

Once deployed the rake task will need deploying.


app/controllers/api/visits_controller.rb
```diff
@@ -49,20 +49,36 @@ def create
...
+        VisitorCancellationResponse.new(visit: load_visit)
+    end
+
+    def visitor_withdrawal_response
+      @_visitor_withdrawal_response ||=
+        VisitorWithdrawalResponse.new(visit: load_visit)
+    end
+
+    def load_visit
+      @_visit ||= Visit.find(params[:id])
```
StupidCodeFactory: having ```@_visit``` and ```@visit``` is redundant. We could get rid of ```@_visit```
Also ```@_xxxx``` is not consistent with the rest of the code base, instance variables are already private and whenever accessed outside a class they have to be accessed via a method.

config/environments/development.rb
```diff
@@ -19,6 +19,6 @@
   config.i18n.load_path =
     Dir[Rails.root.join('config', 'locales', '**', '*.yml').to_s]
 
-  config.nomis_api_host = ENV.fetch('NOMIS_API_HOST',
-    'http://172.22.16.2:8080/')
+  # config.nomis_api_host = ENV.fetch('NOMIS_API_HOST',
```
StupidCodeFactory: That's probably unintentionally commited

app/services/visitor_cancellation_response.rb
```diff
@@ -0,0 +1,30 @@
+class VisitorCancellationResponse
+  attr_reader :visit
```
alan: It is used by the processor

app/services/cancellation_response.rb
```diff
@@ -0,0 +1,28 @@
+class CancellationResponse
+  attr_reader :visit, :user, :reason
```
alan: It is used by the cancellation processor so it has to stay as it is

PR #592, opened 2016-11-08
Fix invalid renewal dates

StupidCodeFactory:


spec/support/helpers/booking_response_helper.rb
```diff
@@ -1,5 +1,5 @@
 module BookingResponseHelper
-  def reject_visit(visit, reasons = [Rejection::NO_ALLOWANCE])
+  def reject_visit(visit, reasons = [Rejection::SLOT_UNAVAILABLE])
```
alan: Why the default value for reasons? I wouldn't expect a default value to be necessary as the reasons should always be explicitly set

StupidCodeFactory: It's a fait point, although that's rather convenient. The factory itself used to have a default reason
https://github.com/ministryofjustice/prison-visits-2/commit/bc07392b99d1c6088bc3ad73558d9e3a2616828c#diff-23ebfcb2be1d390978dbe12b82767259L4

app/controllers/prison/visits_controller.rb
```diff
@@ -18,14 +18,15 @@ def process_visit
...
-    @visit = load_visit.decorate
+    @visit = load_visit
     @visit.assign_attributes(visit_params)
     @booking_response = BookingResponse.new(visit: @visit, user: current_user)
     if @booking_response.valid?
       BookingResponder.new(@booking_response, message).respond!
       flash[:notice] = t('process_thank_you', scope: [:prison, :flash])
       redirect_to visit_page(@visit)
     else
+      @visit = @visit.decorate
```
alan: Can you add a comment on the importance of doing this here? I still don't see how this fixes it...

spec/support/helpers/booking_response_helper.rb
alan: Yep, I thought this was the reject processor, makes sense now

app/controllers/prison/visits_controller.rb
StupidCodeFactory: Will do.
Typically you only decorate your objects once right before render views. The README  https://github.com/drapergem/draper#when-to-decorate-objects as much better explanation
What happened here is that decorating the ```@visit``` object would also automatically decorate the related rejection object (as I have set it up). RejectionDecorator got passed down to ```BookingResponder::Reject``` processor and tried to save allowance_renews_on as an AccessibleDate with casted in an unexpected way.
I think that is what actually may have been wrong with test i marked as pending but i need to verify this.

StupidCodeFactory:
Yes it is definitely worth, i'd like to get this out of the door though relatively quickly

kyliemulholland:
screen shot 2016-11-09 at 09 24 36

Not working as expected - it should be possible to reject the visit simply by ticking the checkbox. Not mandatory to enter a renewal date. @StupidCodeFactory @alan

kyliemulholland:
Also, it's currently possible to reject the request with only the Day and Month. This should create a validation error without the year.

kyliemulholland:
It's currently possible to enter a VO renewal date in the past. I was able to process the visit with VO renewal date 1/1/2016.

alan:
@kyliemulholland the date in the past has always been the case, do you want to fix this is part of this bug?

kyliemulholland:
It should not be possible to process a VO renewal date _without_ checking the checkbox. This is possible currently.

kyliemulholland:
@alan see above comment - Fine to push out the bug fix without this, and work on it afterwards.

kyliemulholland:
@StupidCodeFactory just reviewed, looks like it is currently not possible to process a rejection by only selecting the checkbox.

StupidCodeFactory:
@kyliemulholland yes just saw that as well

alan:
Entering a date without selecting 'no allowance' checkbox and another reason ie 'child protection measures the visit is rejected because of child protection. Think we want to force the date to be blank if the 'no vo allowance' is not checked. @kyliemulholland can you confirm?

kyliemulholland:
Reviewed and good to go!

app/models/rejection.rb
```diff
@@ -31,46 +31,36 @@ class Rejection < ActiveRecord::Base
...
-  def privileged_allowance_available=(value)
-    @privileged_allowance_available = truthy?(value)
-  end
+private
 
-  def allowance_will_renew=(value)
-    @allowance_will_renew = truthy?(value)
+  def validate_allowance_renews_on_date
+    if no_allowance? && !acceptable_allowance_renews_on_date?
+      errors.add(:allowance_renews_on, 'ius not a fate')
```
alan: 'ius not a fate' ?

alan: 'ius not a fate' ?

PR #597, opened 2016-11-09
Ensure that the timeline is ordered correctly

alan:
Fixes ordering of the timeline by sorting by creation date. This will also fix flakey failures that happen in CI from time to time.

Ideally we would have a sort key to enforce linearibility of the state changes, in practice creation dates should have enough precision in production.

### DIFF
--- a/app/services/visit_timeline.rb
+++ b/app/services/visit_timeline.rb
@@ -24,7 +24,7 @@ class VisitTimeline
   end
 
   def events
-    events = @visit.visit_state_changes.reverse.map.with_index { |state, i|
+    events = visit_state_changes.reverse.map.with_index { |state, i|
       build_event_from_state_change(state, last: i.zero?)
     }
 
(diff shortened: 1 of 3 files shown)


app/services/visit_timeline.rb
```diff
@@ -34,6 +34,10 @@ def events
 
 private
 
+  def visit_state_changes
+    @visit.visit_state_changes.sort { |a, b| a.created_at <=> b.created_at }
```
StupidCodeFactory: Any reason to sort this in memory instead of SQL?

app/services/visit_timeline.rb
```diff
@@ -24,7 +24,7 @@ def initialize(visit)
   end
 
   def events
-    events = @visit.visit_state_changes.reverse.map.with_index { |state, i|
+    events = visit_state_changes.reverse.map.with_index { |state, i|
```
StupidCodeFactory: you might want to move the ```reverse``` down to the ```#visit_state_changes``` to keep all things *sorting* together

alan: I think is a bit more explicit that we reverse the ordering to be able to spot the last element

app/services/visit_timeline.rb
alan: Takes advantage of eager loaded data if it is in memory, and given that it will be a small number of objects to sort there is little gain to doing in sql

PR #603, opened 2016-11-10
Prevent easily double processing a visit

alan:
It was possible to process a visit twice by opening the process page in 2 tabs
and processing them separately.

This change makes it a lot less likely to happen, but still possible. To avoid
it completely we need a state machine that enforces linearibility.

### DIFF
--- a/app/controllers/prison/visits_controller.rb
+++ b/app/controllers/prison/visits_controller.rb
@@ -4,15 +4,11 @@ class Prison::VisitsController < ApplicationController
   before_action :require_login_during_trial,
     only: %w[show nomis_cancelled process_visit update]
   before_action :cancellation_reason_set, only: :cancel
+  before_action :visit_is_processable, only: [:process_visit, :update]
 
   def process_visit
     @visit            = load_visit.decorate
     @booking_response = BookingResponse.new(visit: @visit)
-
-    unless @visit.processable?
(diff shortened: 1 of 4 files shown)


app/controllers/prison/visits_controller.rb
```diff
@@ -4,15 +4,11 @@ class Prison::VisitsController < ApplicationController
   before_action :require_login_during_trial,
     only: %w[show nomis_cancelled process_visit update]
   before_action :cancellation_reason_set, only: :cancel
+  before_action :visit_is_processable, only: [:process_visit, :update]
```
StupidCodeFactory: This could be done inside the ```process_visit``` and ```update``` actions as it requires the ```@visit``` to be loaded and we already do that in each action.

app/controllers/prison/visits_controller.rb
```diff
@@ -65,6 +61,14 @@ def cancel
 
 private
 
+  def visit_is_processable
+    visit = load_visit
```
StupidCodeFactory: it would be nice to avoid loading the visit twice per action (even if the second time it is memoized)

app/controllers/prison/visits_controller.rb
alan: I think having this as a before action is slightly clearer that both require the visit is processable as well as making the action methods more concise. We load the visit twice, but as you say they are memoised so I see no problem.

PR #607, opened 2016-11-15
Banned visitor until date

alan:


kyliemulholland:
I've tested this. Feedback:

- font size for 'Banned until eg 28 04 1996' looks too large, should be same size as surrounding text
- in the case where a banned date has been entered, but the 'visitor banned' checkbox has not bee ticked, there is no validation error. the preview email gives the following error ["Visit translation missing: en.activemodel.errors.models.booking_response.attributes.visit.visitors_invalid"]

alan:
@kyliemulholland if you try to process the visit, instead of using preview, the form will be rendered with the error that you are expecting.

The email preview shows the white page that you saw when there is any validation error, that isn't related to this PR. Although you found a missing translation so I've fixed that.

kyliemulholland:
@chrishanes final thing to resolve is around font sizing - could you take a look?

kyliemulholland:
@alan this is good to go now

PR #621, opened 2016-11-28
Fix Url too long for IE

StupidCodeFactory:
**CHANGES:**
In order to circumvent URL being to long for IE the visit process form is serialised and sent using ajax.

### DIFF
--- a/app/assets/javascripts/modules/moj.emailPreview.js
+++ b/app/assets/javascripts/modules/moj.emailPreview.js
@@ -17,7 +17,7 @@
 
       var ajaxOptions = {
         url:    event.target.href,
-        method: 'POST',
+        method: 'POST', // not all browsers support PUT and DELETE
         data:   $(this.el).serialize()
       };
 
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/modules/moj.emailPreview.js
```diff
@@ -13,7 +13,20 @@
     },
 
     handleClick: function(event) {
-      event.target.search = $(this.el).serialize();
+      event.preventDefault();
+
+      var ajaxOptions = {
+        url:    event.target.href,
+        method: 'POST',
```
alan: Should this be a PUT based on the route, or change route to a POST?

StupidCodeFactory: It actually is a 'PUT', the serialised  form has a hidden input '_method=put' so it is perfectly fine. And not all browsers (yes internet explorer, i'm looking at you!) supports does PUT and DELETE requests. Which is why rails always add the ```_method``` in forms.

StupidCodeFactory: @alan ^

alan: Ah, the hidden method explains it, worth adding a comment? Otherwise is not clear how this actually works

PR #623, opened 2016-11-28
Log api calls

StupidCodeFactory:
https://trello.com/c/L2s8TdDi/737-5-log-api-calls-made-from-pvb2

**CHANGE**:

I've imported the ```Instrumentation``` class from pvb-public.

### DIFF
--- a/spec/services/nomis/client_spec.rb
+++ b/spec/services/nomis/client_spec.rb
@@ -15,11 +15,13 @@ RSpec.describe Nomis::Client do
 
   subject { described_class.new(api_host, client_token, client_key) }
 
-  it 'sets the X-Request-Id header if a request_id is present', vcr: { cassette_name: 'client-request-id' } do
-    RequestStore.store[:request_id] = 'uuid'
-    subject.get(path, params)
-    expect(WebMock).to have_requested(:get, /\w/).
-      with(headers: { 'X-Request-Id' => 'uuid' })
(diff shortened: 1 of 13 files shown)


spec/services/nomis/client_spec.rb
```diff
@@ -15,19 +15,44 @@
...
+        to_raise(Excon::Errors::Timeout.new('Request Timeout'))
+    end
+
+    it 'raises an APIError if an unexpected exception is raised containing request information' do
+      expect {
+        subject.get(path, params)
+      }.to raise_error(Nomis::APIError, 'Exception Excon::Errors::Timeout calling GET /nomisapi/lookup/active_offender: Request Timeout')
+    end
+
+    it 'imcrements the api error count' do
```
alan: increments

alan:
Can you squash the commits as there is a lot of wip commits?

StupidCodeFactory:
Yes there are few challenges with API's api, it's rather difficult to hook into it. 
Also i've had a look at the ```excon``` gem it does not do connection pooling and keep alive out of the box. Implementing this could potentially reduce the amount of connection errors we have.

alan:
We are using persistent connections with exconn, which I think is what you mean by keep alive?Agree with the connection pool, this is something we have with the sendgrid api client, we could do something similar.  Without it we'll potentially have contention between puma threads when making API calls.

PR #624, opened 2016-11-30
Nomis integration metrics

StupidCodeFactory:
**CHANGES:** 
New metrics page

### DIFF
--- a/app/models/percentiles_by_calendar_date.rb
+++ b/app/models/percentiles_by_calendar_date.rb
@@ -1,0 +1,11 @@
+class PercentilesByCalendarDate < ActiveRecord::Base
+  include PercentileSerialisation
+
+  def readonly?
+    true
+  end
+
+  def self.refresh
+    Scenic.database.refresh_materialized_view(table_name, concurrently: false)
+  end
+end
--- a/app/views/metrics/index.html.erb
+++ b/app/views/metrics/index.html.erb
@@ -23,4 +23,3 @@
   </h3>
 <% end %>
 <%= render partial: 'all_prisons', locals: { dataset: @dataset } %>
-
--- a/db/views/rejection_percentage_by_days_v01.sql
+++ b/db/views/rejection_percentage_by_days_v01.sql
@@ -1,0 +1,71 @@
+WITH rejection_reasons AS (
+  SELECT visit_id, unnest(reasons) AS reason
+  FROM rejections
+  GROUP BY visit_id, reasons
+),
+
+rejected_visits_count_by_prison AS (
+  SELECT prisons.name            AS prison_name,
+         prisons.id              AS prison_id,
+         visits.created_at::date AS date,
+         COUNT(*)                AS rejected_count
+  FROM visits
+  INNER JOIN prisons ON prisons.id = visits.prison_id
+  WHERE visits.processing_state    = 'rejected'
+  GROUP BY prison_name, prisons.id, date
+),
+
+visit_count_by_prison AS (
+  SELECT prisons.name            AS prison_name,
+         prisons.id              AS prison_id,
+         COUNT(*)                AS total_count,
+         visits.created_at::date AS date
+  FROM visits
+  INNER JOIN prisons ON prisons.id = visits.prison_id
+  GROUP BY prison_name, prisons.id, date
+),
+rejected_visit_count_reason_date_and_prison AS (
+  SELECT reason,
+         prisons.name            AS prison_name,
+         prisons.id              AS prison_id,
+         COUNT(*)                AS rejected_count,
+         visits.created_at::date AS date
+  FROM visits
+  INNER JOIN prisons           ON prisons.id = visits.prison_id
+  INNER JOIN rejection_reasons ON visits.id  = rejection_reasons.visit_id
+  WHERE visits.processing_state = 'rejected'
+  GROUP BY prison_name, prisons.id, reason, date
+),
+visit_count_by_prison_and_date AS (
+  SELECT prisons.name            AS prison_name,
+         prisons.id              AS prison_id,
+         COUNT(*)                AS total_count,
+         visits.created_at::date AS date
+  FROM visits
+  INNER JOIN prisons ON prisons.id = visits.prison_id
+  GROUP BY prison_name, prisons.id, date
+)
+
+SELECT rejected.prison_name,
+       rejected.prison_id                                     AS prison_id,
+       'total'                                                AS reason,
+       ROUND((rejected_count::numeric / total_count * 100),2) AS percentage,
+       rejected.date                                          AS date
+FROM
+  rejected_visits_count_by_prison AS rejected,
+  visit_count_by_prison           AS total
+WHERE rejected.prison_id = total.prison_id AND rejected.date = total.date
+GROUP BY rejected.prison_name, rejected.prison_id, reason, percentage, rejected.date
+
+UNION
+
+SELECT rejected.prison_name,
+       rejected.prison_id,
+       reason,
+       ROUND((rejected_count::numeric / total_count * 100),2) AS percentage,
+       rejected.date                                          AS date
+FROM
+  rejected_visit_count_reason_date_and_prison AS rejected,
+  visit_count_by_prison_and_date              AS  total
+WHERE rejected.prison_name = total.prison_name AND rejected.date = total.date
+GROUP BY rejected.prison_name, rejected.prison_id, reason, percentage, rejected.date
--- a/db/views/visit_counts_by_prison_state_date_and_timely_v01.sql
+++ b/db/views/visit_counts_by_prison_state_date_and_timely_v01.sql
@@ -1,0 +1,33 @@
+WITH visits_timely AS (
+     SELECT
+          prisons.name AS prison_name,
+          prisons.id   AS prison_id,
+          v.processing_state,
+          v.created_at,
+          (
+            CASE
+            WHEN processing_state IN ('booked', 'rejected')
+              THEN (vsc.created_at - v.created_at)::interval < '3 days'::interval
+            ELSE
+              NULL
+            END
+          )::boolean AS timely
+    FROM visits v
+    LEFT OUTER JOIN (SELECT MAX(created_at) AS created_at, visit_id FROM visit_state_changes GROUP BY visit_id) vsc ON vsc.visit_id = v.id
+    LEFT OUTER JOIN prisons ON prisons.id = v.prison_id
+)
+
+SELECT
+    v.prison_name,
+    v.prison_id,
+    v.processing_state,
+    v.timely,
+    v.created_at::date AS date,
+    COUNT(*)
+FROM visits_timely v
+GROUP BY
+      v.prison_name,
+      v.prison_id,
+      v.processing_state,
+      v.timely,
+      date;
--- a/lib/tasks/populate.rake
+++ b/lib/tasks/populate.rake
@@ -49,7 +49,8 @@ namespace :pvb do
         slot_option_0:          slots.pop,
         slot_option_1:          slots.pop,
         slot_option_2:          slots.pop,
-        locale:                 'en'
+        locale:                 'en',
+        created_at: (1..20).to_a.sample.days.ago
       )
       rand(1..5).times.map do |i|
         visit.visitors.create!(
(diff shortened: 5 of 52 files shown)


spec/presenters/new_metrics_presenter_spec.rb
```diff
@@ -0,0 +1,287 @@
...
+        expect(median).to be_within(1).of(1)
+      end
+    end
+
+    describe '#percentiles_by_day_for' do
+      it 'returns the distributions of time to process a request for a given prison' do
+        nineteenth = subject.percentiles_by_day_for(prison).first.percentiles.first / 86_400
+        median     = subject.percentiles_by_day_for(prison).first.percentiles.last / 86_400
+        expect(nineteenth).to be_within(1).of(5)
+        expect(median).to be_within(1).of(1)
```
alan: Can we have more precision with these `be_within` ? The possible acceptable range is pretty big given the expected result

spec/presenters/new_metrics_presenter_spec.rb
```diff
@@ -0,0 +1,287 @@
...
+    RejectionPercentageByDay.refresh
+  end
+
+  describe 'Percentiles Statistics' do
+    let(:visits) { create_list(:visit, 10, prison: prison) }
+    before do
+      travel_to(visits.first.created_at + 2.days) do
+        visits[0..8].each do |visit|
+          visit.assign_attributes(slot_granted: visit.slots.first)
+          BookingResponder::Accept.new(BookingResponse.new(visit: visit)).process_request
```
alan: Can we use the `BookingResponseHelper` here?

spec/presenters/new_metrics_presenter_spec.rb
```diff
@@ -0,0 +1,287 @@
...
+        end
+      end
+
+      travel_to(visits.first.created_at + 6.days) do
+        visit = visits.last
+        visit.assign_attributes(slot_granted: visit.slots.first)
+        BookingResponder::Accept.new(
```
alan: Can we use the `BookingResponseHelper` here?

spec/controllers/metrics_controller_spec.rb
```diff
@@ -24,6 +24,12 @@
     it_behaves_like 'disallows untrusted ips'
   end
 
+  describe 'show' do
+    let!(:prison) { create :prison }
```
alan: Is the `!` necessary here?

lib/tasks/populate.rake
```diff
@@ -96,9 +121,9 @@ namespace :pvb do
 
     desc 'populate visits for review apps or dev environments'
     task visits: :environment do
-      %w[Sudbury Pentonville Usk].each do |prison_name|
+      %w[Sudbury Pentonville Usk Aylesbury].each do |prison_name|
         prison = Prison.find_by!(name: prison_name)
-        visits = 0.upto(14).map { |_i|
+        visits = 0.upto(140).map { |_i|
```
alan: Could we make the large number of visits optional? For the majority of review apps it wont be necessary.

app/views/metrics/index.html.erb
```diff
@@ -24,3 +24,16 @@
 <% end %>
 <%= render partial: 'all_prisons', locals: { dataset: @dataset } %>
 
+<%= render 'graph' %>
+
+<div  class="grid-row">
+  <% @prisons.in_groups_of(@prisons.size < 3 ? @prisons.size : (@prisons.size / 3), nil) do |prisons| %>
```
alan: Why the 3?

spec/models/percentiles_by_prison_and_calendar_date_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+RSpec.describe PercentilesByPrisonAndCalendarDate, type: :model do
+  let(:prison)                          { create :prison }
+  let(:processed_within_a_day_visits)   { create_list :visit, 4, prison: prison }
+  let(:processed_within_two_days_visit) { create :visit, prison: prison }
+
+  subject { described_class.first }
+
+  before do
+    travel_to(processed_within_a_day_visits.first.created_at + 6.hours) do
```
StupidCodeFactory: Good spot

app/models/percentiles_by_calendar_date.rb
```diff
@@ -0,0 +1,7 @@
+class PercentilesByCalendarDate < ActiveRecord::Base
+  include PercentileSerialisation
+
+  def self.refresh
+    Scenic.database.refresh_materialized_view(table_name, concurrently: false)
```
StupidCodeFactory: It requires a unique index for each table, which is easy enough to add. The only downside to not do it concurrently is it locks the table for ```select```, which in our case is not a problem at as the view will be refresh once a day at midnight. So i didn't bother with concurrent refresh.

app/views/metrics/index.html.erb
StupidCodeFactory: Prisons are displayed using 3 columns. That's why. 
I will tidy this up in a presenter, this ugly.

db/views/distribution_by_prison_and_calendar_dates_v02.sql
```diff
@@ -0,0 +1,13 @@
+SELECT prisons.name AS prison_name,
+       PERCENTILE_DISC(ARRAY[0.99, 0.95, 0.90, 0.75, 0.50, 0.25])
```
StupidCodeFactory: This is actually not used i will remove

lib/tasks/populate.rake
StupidCodeFactory: It's a good way to catch performance problem, i.e on the inbox page there are  two n+1 queries. I was able to catch it because i loaded a load of records. But i'm happy to revert it back to 14 or use an ENV var. 
We don't really care about the number of records on review apps, do we?

spec/models/metrics/rejection_percentage_spec.rb
```diff
@@ -0,0 +1,27 @@
+require "rails_helper"
+
+RSpec.describe Metrics::RejectionPercentage do
+  Rejection::REASONS.each do |reason|
+    it "##{reason} defaults to 0" do
+      expect(subject.public_send(reason)).to eq(0)
+    end
+  end
+
+  describe '#attributes' do
```
StupidCodeFactory: No need for more. Perhaps a better name would be clearer.
Instead of ```has all the attributes``` it would be ```has numeric attributes initialised to 0``` ?

spec/presenters/new_metrics_presenter_spec.rb
```diff
@@ -0,0 +1,287 @@
...
+          BookingResponse.new(visit: visit)
+        ).process_request
+      end
+
+      refresh_views
+    end
+
+    describe '#percentiles_by_day' do
+      it 'returns the distributions of time to process a request' do
+        nineteenth = subject.percentiles_by_day.first.percentiles.first / 86_400
```
StupidCodeFactory: the percentiles are stored a seconds in the database that just convert it back to a day value. 86_400 is commonly known as the number of seconds per day, i can make that clearer though, will stick this into a ```let(:seconds_per_day)```

spec/presenters/new_metrics_presenter_spec.rb
StupidCodeFactory: I'd have to do a lot of math to get these number be exact but i can try.

db/views/rejection_percentage_by_days_v01.sql
```diff
@@ -0,0 +1,69 @@
+WITH rejection_reasons AS (
```
StupidCodeFactory: So yes it does not a lot cleaner but I'm really not convinced it's helpful on such queries, In fact i did do that at first but when debugging the query I found my self jumping up and down the file and loosing the focus off the main query. 
I have very mixed feelings about this syntax specially since we are not reusing those bits anywhere which is the main reason behind them.
I'm a happy to change it but be aware of that.

StupidCodeFactory:
* views will be refreshed via a cron job (though we spoke about it apologies if we didn't)
* I did ran those in production (had to add the PSQL client to the container), also with only 17k visits i'm quite sure we won't run into performance problem for a while, that is also why we use materialized view which mitigates performance issues as they have to be precompiled.

db/views/rejection_percentage_by_days_v01.sql
alan: I'm not married to a particular style, but would like to keep the style consistent across files.

lib/tasks/populate.rake
alan: The reason it got my attention is that review apps have been flakey in Heroku, we seem to be hitting rate limiting on one of the addons, I think it was Mailgun.

If review apps can be deployed consistently then I also prefer to have a larger dataset in review apps..

I should have mentioned this is my previous message

spec/features/metrics_spec.rb
```diff
@@ -2,10 +2,30 @@
 require 'rails_helper'
 require_relative '../metrics/shared_examples_for_metrics'
 
-RSpec.feature 'Metrics', js: true do
+RSpec.xfeature 'Metrics', js: true do
```
alan: What's the plan for this long term?

spec/models/metrics/rejection_percentage_spec.rb
alan: How come is not needed, is it tested elsewhere?

StupidCodeFactory: I don't understand what you want to test here?

alan: What I'm thinking is that this model is for the view that calculates the rejection percentages, so would be good to cover the case of when there are rejections to ensure that the values are correct.

StupidCodeFactory: @alan that model doesn't calculate anything. It really just is a DAO. 
I could test that setting a value sets a value.
All the computation is tested in ```GraphMetricsPresenter``` specs
Or maybe i'm miss understanding what you're saying

alan: The computation is what I meant 👍

app/views/metrics/_graph.erb
```diff
@@ -0,0 +1,8 @@
+<h2 class="heading-medium">Statistics for integration outcomes</h2>
```
alan: Harcoded text, all others seem to be locale files

db/views/visit_counts_by_prison_state_date_and_timely_v01.sql
```diff
@@ -0,0 +1,33 @@
+WITH visits_timely AS (
```
alan: This format is the one we get to get rid no?

spec/presenters/graph_metrics_presenter_spec.rb
```diff
@@ -0,0 +1,283 @@
...
+        travel_to(Time.zone.today + 6.days) do
+          (overdue_visits + other_prison_overdue_visits).each_with_index do |visit, i|
+            process_visit(visit, i)
+          end
+        end
+
+        refresh_views
+      end
+
+      it "returns the formted overdue and timely visits by date for the given prison" do
```
alan: formatted

spec/presenters/graph_metrics_presenter_spec.rb
```diff
@@ -0,0 +1,283 @@
...
+  describe 'processing_state breakdown' do
+    describe '#visits_per_processing_state_for' do
+      let!(:requested_visits)       { create_list :visit, 2, prison: prison }
+      let!(:other_requested_visits) { create :visit, prison: other_prison }
+
+      before do
+        refresh_views
+      end
+
+      it 'returns timeseries for of visits per processing state only for the given prison' do
```
alan: timeserios ~~for~~ of visits

spec/presenters/graph_metrics_presenter_spec.rb
```diff
@@ -0,0 +1,283 @@
...
+        travel_to(Time.zone.today + 6.days) do
+          overdue_visits.each_with_index do |visit, i|
+            process_visit(visit, i)
+          end
+        end
+
+        refresh_views
+      end
+
+      it "returns the formted overdue and timely visits by date" do
```
alan: formatted

PR #625, opened 2016-11-30
Set NOMS API timeout to 2 seconds

alan:
The timeout needs to be stricter than the timeout in PVB public.


alan:
https://github.com/ministryofjustice/prison-visits-public/pull/91

PR #633, opened 2016-12-05
Set a connection pool for the Nomis API

alan:
The app runs in a threaded environment so we need to make sure that we have at
least a client available for each thread. If we have less clients than threads
then we could have threads blocked while waiting for the connection to be
released.

### DIFF
--- a/app/services/nomis/api.rb
+++ b/app/services/nomis/api.rb
@@ -4,33 +4,31 @@ module Nomis
   NotFound = Class.new(Error)
 
   class Api
-    class << self
-      def enabled?
-        Rails.configuration.nomis_api_host != nil
-      end
+    include Singleton
 
-      def instance
-        unless enabled?
-          fail DisabledError, 'Nomis API is disabled'
-        end
-
-        @instance ||= begin
-          host = Rails.configuration.nomis_api_host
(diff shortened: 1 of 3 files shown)


spec/services/nomis/api_spec.rb
```diff
@@ -3,6 +3,12 @@
 RSpec.describe Nomis::Api do
   subject { described_class.instance }
 
+  # Ensure that we have a new instance
```
StupidCodeFactory: does this cause issues with mocked and stubbed objects?

app/services/nomis/api.rb
```diff
@@ -4,33 +4,31 @@ module Nomis
...
+    def initialize
+      unless self.class.enabled?
+        fail DisabledError, 'Nomis API is disabled'
       end
-    end
 
-    def initialize(client)
-      @client = client
+      pool_size = Rails.configuration.connection_pool_size
+      @pool = ConnectionPool.new(size: pool_size, timeout: 1) do
```
StupidCodeFactory: Probably want the timeout  to be configurable as ENV var so it can easily be tweaked when needed (i.e a new release when the NOMIS API is slower than usual)

PR #645, opened 2016-12-08
Verify past prisoner details in Nomis

alan:
Reads in the Nomis API configuration and the number of days of most recent
prisoners to check and prints out json entry for each prisoner check saying if
it was found or had and error.

By reading the Nomis API configuration we should be able to test in production
before configuring the server.

### DIFF
--- a/lib/tasks/one_off.rake
+++ b/lib/tasks/one_off.rake
@@ -45,4 +45,86 @@ namespace :pvb do
       vs.update_column(:visitor_id, vs.visit.principal_visitor.id)
     end
   end
+
+  # rubocop:disable Metrics/MethodLength
+  # rubocop:disable Lint/AssignmentInCondition
+  # rubocop:disable Style/RescueModifier
+  def check_prisoner_details(client, number, dob)
+    RequestStore.store[:custom_log_items] = {}
+    log = { noms_id: number }
+
+    begin
+      response = client.get('/lookup/active_offender',
+        noms_id: number,
+        date_of_birth: dob)
+
+      log[:found] = !!response['found']
+    rescue Nomis::APIError => e
+      log[:error] = e.message
+    end
+
+    log[:timing] = RequestStore.store[:custom_log_items][:api]
+    log
+  end
+
+  def new_worker(pool, queue, thread_number)
+    Thread.new do
+      request_id = "check_prisoner_details_#{thread_number}"
+      RequestStore.store[:request_id] = request_id
+
+      begin
+        while job = queue.pop(true) rescue nil
+          number, dob = job
+          pool.with do |client|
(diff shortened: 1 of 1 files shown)


lib/tasks/one_off.rake
```diff
@@ -45,4 +45,86 @@ namespace :pvb do
...
+
+    nomis_host = cli.ask('Nomis host: ')
+    days_to_check = cli.ask('Number of days to check: ', Integer)
+
+    data = Prisoner.
+           where('created_at > ?', days_to_check.days.ago).
+           pluck(:number, :date_of_birth).
+           uniq
+
+    queue = Queue.new
```
StupidCodeFactory: 👍

lib/tasks/one_off.rake
```diff
@@ -45,4 +45,86 @@ namespace :pvb do
...
+    data.each do |pair| queue << pair end
+    STDOUT.puts "Checking #{queue.size} prisoners"
+    STDOUT.print "\n"
+
+    pool = ConnectionPool.new(size: 5, timeout: 60) do
+      Nomis::Client.new(nomis_host, nomis_token, nomis_key)
+    end
+
+    workers = 1.upto(10).map { |i| new_worker(pool, queue, i) }
```
StupidCodeFactory: could have a bit more than 10 threads maybe if this is a bit long? (need to be careful not to DDOS Nomis though)

lib/tasks/one_off.rake
```diff
@@ -45,4 +45,86 @@ namespace :pvb do
...
+    end
+
+    log[:timing] = RequestStore.store[:custom_log_items][:api]
+    log
+  end
+
+  def new_worker(pool, queue, thread_number)
+    Thread.new do
+      request_id = "check_prisoner_details_#{thread_number}"
+      RequestStore.store[:request_id] = request_id
```
StupidCodeFactory: not sure why this is useful?

alan: To find the requests easily in Kibana

StupidCodeFactory: how this being logger to kibana?

lib/tasks/one_off.rake
alan: Yeah, let's see. I'll test in dev when it gets deployed, we'll also want to tweak the connection pool

lib/tasks/one_off.rake
alan: The request_id gets sent as a header, and kibana parses it from the gateway logs

StupidCodeFactory: ah yes found it https://github.com/ministryofjustice/prison-visits-2/blob/master/app/services/nomis/client.rb#L48

PR #648, opened 2016-12-13
Add IRB history

StupidCodeFactory:
This will retain history between IRB sessions in containers


StupidCodeFactory:
reopening this PR

config/irbrc.example
```diff
@@ -0,0 +1,6 @@
+require 'irb/ext/save-history'
+
+#History configuration
```
StupidCodeFactory: reopen

PR #651, opened 2016-12-13
Multi prison

alan:
Only to be merged once users have been setup!


app/models/signon_identity.rb
```diff
@@ -34,21 +32,16 @@ def from_session_data(data)
...
       orgs = permissions.map { |p| p.fetch('organisation') }
 
-      return [] if orgs.empty?
-
-      if orgs.include?(DIGITAL_ORG)
-        Estate.all
-      else
-        Estate.where(sso_organisation_name: orgs)
-      end
+      mapper = EstateSSOMapper.new(orgs)
```
StupidCodeFactory: 👍

app/services/estate_sso_mapper.rb
```diff
@@ -0,0 +1,56 @@
...
+      grouped_estates = Hash.new { |h, k| h[k] = [] }
+      Estate.where.not(group: nil).each do |estate|
+        grouped_estates[estate.group] << estate.sso_organisation_name
+      end
+      grouped_estates.values.freeze
+      grouped_estates.freeze
+    end
+  end
+
+  def self.reset_grouped_estates
```
StupidCodeFactory: I'm not keen on mixing test and production code, this could easily be stubbed/mocked in tests

app/services/estate_visit_query.rb
```diff
@@ -35,7 +31,23 @@ def processed(limit:, prisoner_number: nil)
...
     Visit.
       ready_for_processing.
-      from_estate(@estate).
+      from_estates(@estates).
       count
   end
+
+private
+
+  def grouped_visits(visits)
```
StupidCodeFactory: :scream: could we make this readable, i'm thinking you can get this same result using a postgres window function? I know that the intent here is to reuse the the  visits in memory array but this is really hard to read.

app/views/prison/dashboards/_prison_switcher.html.erb
```diff
@@ -1,11 +1,19 @@
...
-      <%= label_tag "estate_id", class: "form-label" do -%>
-        Change prison
-      <% end %>
-      <%= select_tag 'estate_id', options_from_collection_for_select(sso_identity.accessible_estates, 'id', 'name', current_estate.id), class: 'form-control form-control-auto-width' %>
-      <%= submit_tag t('.switch_estate'), class: 'button-link' %>
+<div class="grid-row push-bottom">
+  <div class="column-full">
+    <% content_for :prison_switcher do %>
+      <% if sso_identity.accessible_estates.size > 1 -%>
+        <%= form_tag prison_switch_estates_path, enforce_utf8: false, class: 'form prison-switcher-form' do -%>
```
StupidCodeFactory: out of curiosity why is the enforce utf8 set to false?

app/controllers/prison/dashboards_controller.rb
```diff
@@ -31,44 +30,53 @@ def print_visits
...
+                     prisoner_number: prisoner_number)
   end
 
-  def switch_estate
-    estate = Estate.find(params[:estate_id])
-
-    if sso_identity.accessible_estate?(estate)
-      @_current_estate         = nil
-      session[:current_estate] = estate.id
+  def switch_estates
```
StupidCodeFactory: Could we have a ```SwitchEstatesController``` and not bundle a tone of unrelated stuff together?

StupidCodeFactory: possibly sharing common code via a module or a new class (which i'd prefer)

config/routes.rb
```diff
@@ -53,7 +53,7 @@
       get '/processed', action: :processed, as: 'processed_visits'
       get '/print_visits', action: :print_visits, as: 'print_visits'
       get '/search', action: :search, as: :search
-      post '/switch_estate', action: :switch_estate, as: 'switch_estate'
+      post '/switch_estates', action: :switch_estates, as: 'switch_estates'
```
StupidCodeFactory: ```SwitchEstatesController#create```

```ruby
resource :switch_estates, only: :create
```

spec/controllers/prison/feedbacks_controller_spec.rb
```diff
@@ -8,7 +8,9 @@
 
     before do
       login_user(
-        FactoryGirl.create(:user), FactoryGirl.create(:estate))
+        FactoryGirl.create(:user),
```
StupidCodeFactory: Maybe we can drop the ```FactoryGirl``` would be less cluttered?

config/routes.rb
StupidCodeFactory: Using ```resources``` even if it means a having a controller with only one action isn't a big deal and encourages separation of concerns. I'm weary to see fatter controller than what we have now.

app/views/prison/dashboards/_prison_switcher.html.erb
alan: Can't remember why we added it...

app/services/estate_sso_mapper.rb
alan: Yeah, I prefer to avoid generally. The issue here is that the order of the tests matter, so if this spec file runs before the feature specs, then some of those fail, so the best thing is to always cleanup after using the method

app/services/estate_visit_query.rb
alan: We've looked into it and it is not possible to simplify with window functions

app/controllers/prison/switch_estates_controller.rb
```diff
@@ -0,0 +1,26 @@
+class Prison::SwitchEstatesController < ApplicationController
+  before_action :authorize_prison_request
+  before_action :authenticate_user
+  before_action :verify_switch_estates
+
+  def create
+    estates = Estate.where(id: params[:estates_id])
```
StupidCodeFactory: typo? ```estate_ids```

app/views/prison/dashboards/search.html.erb
```diff
@@ -16,3 +16,8 @@
 
 <h3 class="heading-medium"><%= t('.processed') %></h3>
 <%= render 'processed' %>
+<script>
+  var pvb = {
```
StupidCodeFactory: @chrishanes how do you feel about having this as a data attributes on the body element (or the another relevant element) rather than poluting the gloabal JS namespace?

PR #656, opened 2016-12-20
Use excon instrumentation

StupidCodeFactory:


lib/pvb/excon/instrument/request.rb
```diff
@@ -0,0 +1,31 @@
+module PVB
+  module Excon
+    module Instrument
+      class Request
+        include Instrument
+
+        def process
+          Instrumentation.incr(:api_request_count)
```
StupidCodeFactory: **NOTE:** 

needs to add  counter to the custom log

spec/controllers/application_controller_spec.rb
```diff
@@ -33,7 +33,8 @@ def create
       WebMock.stub_request(:get, /\w/).
         to_raise(Excon::Errors::Timeout.new('Request Timeout'))
       post :create
-      expect(Instrumentation.custom_log_items[:api_request_count]).to eq(1)
+      # 2 because we now catch the retries
+      expect(Instrumentation.custom_log_items[:api_request_count]).to eq(2)
```
alan: Here I would expect 1 api request, and the log separately that there was 1 retry. Don't know if that is possible though.

That way we can differentiate between an http request trigerring 2 api calls, from another making 1 api call that gets retried.

StupidCodeFactory: i think the instrumentation payload contains a flag saying if it is a retry so this could actually be done

spec/controllers/application_controller_spec.rb
```diff
@@ -33,6 +33,7 @@ def create
       WebMock.stub_request(:get, /\w/).
         to_raise(Excon::Errors::Timeout.new('Request Timeout'))
       post :create
+      # 2 because we now catch the retries
```
alan: Needs updating

lib/pvb/excon/instrument/retry.rb
```diff
@@ -0,0 +1,13 @@
+module PVB
+  module Excon
+    module Instrument
+      class Retry < Request
+        def process
+          Instrumentation.incr(:api_retry_count)
```
alan: 👍

config/initializers/excon_instrumentation.rb
```diff
@@ -0,0 +1,6 @@
+require 'pvb/excon/instrument_factory'
+
+ActiveSupport::Notifications.subscribe(/excon/) do |event, start, finish, _id, payload|
+  instrument = PVB::Excon::InstrumentFactory.for(event, start.to_f, finish.to_f, payload)
```
alan: Can you add a comment roughly explaining how Excon instrumentation works? ie when a request happens it calls the generates a 'request', 'response' events, etc ? I think it will make it easier to maintain down the line having a bit of an intro.

StupidCodeFactory: 👍

StupidCodeFactory:
@alan this won't generate date from the StaffController as only the Nomis::Client Excon connection hooks has instrumentation enabled.
However we should generate custom event names to prevent potential future collisions if we decide to monitor other Excon calls

PR #661, opened 2017-01-05
Add instrumentation to API logs

alan:
We were recording other data that wasn't being added to the logs for API calls.

### DIFF
--- a/app/controllers/api/api_controller.rb
+++ b/app/controllers/api/api_controller.rb
@@ -7,8 +7,6 @@ module Api
     before_action :store_request_id
     before_action :enforce_json
 
-    after_action :log_api_counters
-
     rescue_from ActionController::ParameterMissing do |e|
       render_error 422, "Missing parameter: #{e.param}"
     end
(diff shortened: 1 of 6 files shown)


app/controllers/api/api_controller.rb
```diff
@@ -40,17 +42,28 @@ def render_error(status, message)
...
+    end
+
+    def api_request_count
+      RequestStore.store[:api_request_count]
+    end
+
+    def api_error_count
+      RequestStore.store[:api_error_count]
+    end
+
```
mloughran: I find this implementation a bit low level. Rather than handing each incrementer separately, what about adding a method which increments a counter, and handles appending to the log internally?

    Instrumentation.incr(:api_request_count)

app/controllers/api/api_controller.rb
```diff
@@ -7,6 +7,8 @@ class ApiController < ActionController::Base
     before_action :store_request_id
     before_action :enforce_json
 
+    after_action :log_api_counters
+
```
mloughran: This hook would be unnecessary.

app/controllers/api/api_controller.rb
alan: Good idea, I've pushed the change

PR #663, opened 2017-01-09
Disabled prisons

StupidCodeFactory:
**CHANGES:**

Reinstate disabled prison in the API

### DIFF
--- a/app/services/prison_seeder/seed_entry.rb
+++ b/app/services/prison_seeder/seed_entry.rb
@@ -5,7 +5,7 @@ class PrisonSeeder::SeedEntry
 
   KEYS = %i[
     address adult_age booking_window email_address enabled lead_days name
-    phone_no postcode slot_details translations weekend_processing
+    phone_no postcode slot_details translations weekend_processing closed private
   ]
 
   def initialize(hash)
@@ -46,6 +46,14 @@ private
 
   def name
     hash.fetch('name')
+  end
+
+  def closed
+    hash.fetch('closed')
+  end
+
+  def private
+    hash.fetch('private')
   end
 
   def phone_no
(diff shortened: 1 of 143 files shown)


app/services/prison_seeder/seed_entry.rb
```diff
@@ -48,6 +48,14 @@ def name
     hash.fetch('name')
   end
 
+  def closed
+    hash.fetch('closed', false)
```
alan: Don't need the default here since we already have a default in the db, plus a not null constraint.

app/services/prison_seeder/seed_entry.rb
```diff
@@ -48,6 +48,14 @@ def name
...
+  end
+
+  def private
+    hash.fetch('private', false)
```
alan: Same here

db/seeds/prisons/BLI-bristol.yml
```diff
@@ -8,6 +8,10 @@ postcode: BS7 8PS
 email_address: visitsbooking.westmidlands@noms.gsi.gov.uk
 phone_no: 0300 060 6510
 enabled: true
+private: false
+closed: false
+private: false
```
alan: Duplicates

db/seeds/prisons/KTI-kennet.yml
```diff
@@ -6,7 +6,9 @@ address: |-
 postcode: L31 1HX
 email_address: socialvisits.kennet@hmps.gsi.gov.uk
 phone_no: 0151 213 3179
-enabled: true
+enabled: false
```
alan: I guess disabling the prison here is intentional?

StupidCodeFactory: yes as holloway shoud be as well

PR #664, opened 2017-01-10
Fix processing time on visit show page

alan:


app/views/prison/visits/show.html.erb
```diff
@@ -14,7 +14,7 @@
         <div class="tag tag--requested tag--heading"><%= t('requested', scope: 'shared') %>: <br><%= @visit.created_at.to_date.to_s(:short_nomis) %> - <%= @visit.created_at.to_s(:time) %></div>
       </div>
       <div class="column-one-third">
-        <div class="tag tag--<%= @visit.processing_state %> tag--heading"><%= @visit.processing_state.capitalize %>: <br><%= @visit.updated_at.to_date.to_s(:short_nomis) %> - <%= @visit.created_at.to_s(:time) %></div>
+        <div class="tag tag--<%= @visit.processing_state %> tag--heading"><%= @visit.processing_state.capitalize %>: <br><%= @visit.updated_at.to_date.to_s(:short_nomis) %> - <%= @visit.updated_at.to_s(:time) %></div>
```
StupidCodeFactory: Would be more robust to use the ```visit_state_change.created_at``` field if for some reason the ```visit``` object is changed in any other way than this field will be off.

PR #665, opened 2017-01-11
Not run Nomis validations on non requested visits

alan:


app/services/staff_nomis_checker_factory.rb
```diff
@@ -0,0 +1,10 @@
+class StaffNomisCheckerFactory
+  def self.for(visit)
+    case visit.processing_state
+    when 'requested'
```
StupidCodeFactory: doesn't matter much but it's a bit more elegant and convenient to use constants for these types of things.

alan: Agreed, was surprised we didn't have them. Didn't want to introduce them as part of this PR because it would have introduced a lot of changes.

PR #680, opened 2017-01-26
Add 3 missing disabled prisons

kyliemulholland:


kyliemulholland:
@alan if you could take a look at why the checks are failing, when you get a chance

PR #683, opened 2017-01-30
Staff slot availability

alan:
Introduces staff slot availability similarly to prisoner availability in the staff UI.


app/models/slot_availability_validation.rb
```diff
@@ -0,0 +1,63 @@
...
+  rescue Nomis::APIError => e
+    @api_error = true
+    Rails.logger.warn "Error calling the NOMIS API: #{e.inspect}"
+    nil
+  end
+
+  def valid_requested_slots
+    @valid_slots ||= visit.slots.select { |slot|
+      slot_date = slot.to_date
+      slot_date >= Date.current && slot_date <= 60.days.from_now.to_date
```
StupidCodeFactory: I wonder if we could use the ```#between?(min, max)``` here

app/services/nomis/api_slot_correcter.rb
```diff
@@ -0,0 +1,35 @@
...
+    def slot
+      return @raw_slot if valid_times?
+
+      ConcreteSlot.parse_times(fix_time(@raw_slot.begin_at),
+        fix_time(@raw_slot.end_at))
+    end
+
+  private
+
+    def fix_time(time)
```
StupidCodeFactory: Maybe ```#normalise_times``` would be a better name as ```fix``` suggests what comes back from the API is broken. But I'm not too fussed about it.

app/services/nomis/slot_availability.rb
```diff
@@ -0,0 +1,14 @@
+module Nomis
+  class SlotAvailability
+    include Enumerable
```
StupidCodeFactory: 👍

spec/models/api_slot_availability_spec.rb
```diff
@@ -59,45 +59,47 @@
...
-        expect(Nomis::Api.instance).to receive(:fetch_bookable_slots).
-          with(
-            prison: prison,
-            start_date: Date.parse('2016-04-06'),
-            end_date: Date.parse('2016-04-28')
-          ).
-          and_return([ConcreteSlot.parse('2016-04-12T09:00/10:00')])
+      context 'and the prison slots feature is enabled' do
+        before do
+          allow_any_instance_of(described_class).
```
StupidCodeFactory: I'm not keen on using ```allow_any_instance_of```. Is there a specific reason ```allow``` does not work?

app/services/nomis/api_slot_correcter.rb
```diff
@@ -0,0 +1,35 @@
+module Nomis
+  class ApiSlotCorrecter
```
doctorpod: This name feels a bit 'bondagey'. How about 'TimeGranuator' or 'TimeSnapper' --- just a thought :)

spec/models/slot_availability_validation_spec.rb
```diff
@@ -0,0 +1,186 @@
...
+
+        expect(subject).to be_unknown_result
+      end
+    end
+
+    context 'and working correctly with valid slots' do
+      before do
+        allow(Nomis::Api).to receive(:enabled?).and_return(true)
+
+        expect_any_instance_of(Nomis::Api).
```
StupidCodeFactory: What's the problem with a normal ```expect```?

spec/decorators/concrete_slot_decorator_spec.rb
```diff
@@ -31,37 +33,75 @@
...
+          expect(html_fragment).to have_css('span.date-box__day',    text: 'Friday')
+          expect(html_fragment).to have_text('23 October 2015 14:00–15:30')
+        end
+      end
+
+      context 'when a prisoner is not available' do
+        let(:slot_errors) { ['prisoner_not_available'] }
+
+        it 'renders the chexbox' do
+          expect(html_fragment).to have_css('label.block-label.date-box.date-box--error')
```
doctorpod: Type -> checkbox

spec/models/slot_availability_validation_spec.rb
alan: The connection pool of nomis api clients means we don't know exactly what instance it will be, if I remember correctly

spec/models/slot_availability_validation_spec.rb
```diff
@@ -0,0 +1,186 @@
...
+
+          requested_slots.each do |slot|
+            expect(subject.slot_error(slot)).to be_nil
+          end
+        end
+
+        it { is_expected.to_not be_unknown_result }
+      end
+
+      context 'with dates in the past or to far in the future' do
```
doctorpod: Typo: to -> too

spec/models/api_slot_availability_spec.rb
alan: It doesn't work because the initialiser uses the method that we want to stub

spec/models/slot_availability_validation_spec.rb
alan: I was wrong, I'll take it out.

app/services/nomis/api_slot_correcter.rb
doctorpod: `SlotCorrector` sounds like something from a Bondage Starter Kit. Better names might include `TimeSnapper`, `TimeCorrector`, or `TimeGranulator`.

PR #685, opened 2017-02-01
Upgrade gems

doctorpod:
Trello card: https://trello.com/c/B2yeOZo7


Gemfile
```diff
@@ -34,21 +34,19 @@ gem 'draper'
 # version
 gem 'rake', '< 11.0'
```
StupidCodeFactory: Can we unpin rake?

PR #697, opened 2017-02-08
Automate selection of 'slot unavailable' reason

alan:
When all the slots are either expired (today or older) or not available based on
slot availability.

For prisons that don't have slot availability selected only the expiry check
applies.


alan:
Reviewed by @kyliemulholland already

PR #699, opened 2017-02-08
 Prisoner availability API

StupidCodeFactory:
**CHANGES:**

Create an API to provide a prisoner's availability


app/services/prisoner_slot_availability.rb
```diff
@@ -0,0 +1,69 @@
...
+  end
+
+  def slots
+    if enforce_all_available_slots?
+      return all_slots_available
+    end
+
+    results = Hash.new { |h, slot| h[slot] = [] }
+    prison_slots.each_with_object(results) do |slot, slots_with_availabilities|
+      slots_with_availabilities[slot.to_s]
```
alan: Why read the value but do nothing with it?

config/environments/test.rb
```diff
@@ -21,7 +21,7 @@
   # Always make testing calls to the NOMIS dev server, and hard-code
   # credentials to avoid the possibility of leaking sensitive credentials or
   # returned data into VCR test output files.
-  config.nomis_api_host = 'http://172.22.16.2:8080/'
+  config.nomis_api_host = 'http://151.237.239.116:4888'
```
alan: Why the change of ip?

app/services/prisoner_slot_availability.rb
StupidCodeFactory: That's creating a the key to the ```results``` hash

config/environments/test.rb
StupidCodeFactory: Because it's the new IP, the other one wasn't working

app/services/prisoner_slot_availability.rb
```diff
@@ -0,0 +1,73 @@
...
+    self.api_error     = false
+  end
+
+  def slots
+    if enforce_all_available_slots?
+      return all_slots_available
+    end
+
+    all_slots_available.each do |slot, unavailability_reasons|
+      unless offender_availabilities_dates.delete(slot.to_date)
```
alan: I think this might have a bug. If there are 2 slots for the same date, then only one of the slots will be marked as 'prisoner unavailable' ?

I'd prefer if this method didn't have side effects, so that it would calculate the overall slot availability (`all_slots_available `) each time and also not modify the `offender_availabilities` list.

app/services/prisoner_slot_availability.rb
```diff
@@ -0,0 +1,73 @@
...
+  end
+
+  def enforce_all_available_slots?
+    !nomis_public_prisoner_availability_enabled? ||
+      !offender.valid? ||
+      (load_offender_availabilities && api_error)
+  end
+
+  def nomis_public_prisoner_availability_enabled?
+    Rails.configuration.nomis_public_prisoner_availability_enabled
```
alan: We probably also want to check `Nomis::Api.enabled?` here

config/environments/test.rb
alan: Is that the dev environment?

spec/controllers/api/slots_controller_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+        prisoner_number: prisoner.number,
+        date_of_birth:   prisoner.date_of_birth,
+        start_date:      '2016-02-15',
+        end_date:        '2016-04-15'
+      }
+    }
+
+    let(:slots) {
+      [
+        { '2016-02-15T13:30/14:30' => true },
```
alan: I guess the calendar doesn't need the specific reason is not available?

spec/services/prisoner_slot_availability_spec.rb
```diff
@@ -0,0 +1,112 @@
...
+  let(:start_date)    { Date.parse('2017-02-10') }
+  let(:end_date)      { Date.parse('2017-04-11') }
+
+  subject do
+    described_class.new(prison, offender_id, date_of_birth, start_date..end_date)
+  end
+
+  def all_slots_available
+    {
+      ConcreteSlot.new(2017, 2, 13, 14, 0, 16, 10).to_s => [],
```
alan: Can we make the test more concise by removing some of the slots? I don't think we need so many slots to test this class

config/environments/test.rb
StupidCodeFactory: yes

app/services/prisoner_slot_availability.rb
```diff
@@ -0,0 +1,74 @@
+class PrisonerSlotAvailability
+  PRISONER_UNAVAILABLE = 'prisoner_unavailable'.freeze
```
alan: I'm pretty sure we have the same string defined as a constant somewhere else. We should think of a good way of sharing constants to avoid duplicates thorough the app. Just a general comment, not meant to be addressed in this PR.

app/services/prisoner_slot_availability.rb
```diff
@@ -0,0 +1,74 @@
...
+    self.prison        = prison
+    self.noms_id       = noms_id
+    self.date_of_birth = date_of_birth
+    self.start_date    = date_range.min
+    self.end_date      = date_range.max
+    self.api_error     = false
+  end
+
+  def slots
+    if enforce_all_available_slots?
```
alan: Inline the `if`?

app/services/prisoner_slot_availability.rb
```diff
@@ -0,0 +1,74 @@
...
+      return all_slots
+    end
+
+    results = all_slots.deep_dup.each { |slot, unavailability_reasons|
```
alan: 👍

alan:
@StupidCodeFactory not sure if you saw my previous comment about validating the dates, however after seeing the change in public, perhaps we can remove the dates?

PR #720, opened 2017-02-15
Add ability to reset date selection

chrishanes:
This is a JS solution only as a normal form reset without JS will not clear radio button selections

### DIFF
--- a/spec/javascripts/clear_form_radio_buttons_spec.js
+++ b/spec/javascripts/clear_form_radio_buttons_spec.js
@@ -1,0 +1,51 @@
+//= require jquery
+//= require jasmine-jquery
+
+var input = $('<div class="js-clearRadioButtons">' +
+  '<label class="block-label date-box selected" for="visit_slot_granted">' +
+  '<input checked="checked" type="radio" value="2017-01-24T14:00/16:00" name="visit[slot_granted]" id="visit_slot_granted">' +
+  '</label>' +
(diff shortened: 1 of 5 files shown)


spec/javascripts/clear_form_radio_buttons_spec.js
```diff
@@ -0,0 +1,52 @@
...
+
+    it('label should have `selected` class', function() {
+      expect($('#visit_slot_granted').parent('label')).toHaveClass('selected');
+    });
+  });
+
+  describe('should be unchecked', function() {
+    beforeEach(function() {
+      moj.Modules.clearRadioButtons.init();
+      spyOnEvent('.js-clearRadioButton', 'click');
```
StupidCodeFactory: I'm not sure why you would spy on the the click event if it get triggered the line after?

chrishanes: I forgot to add the test to test the click event

StupidCodeFactory: I see, i don't really understand the value of that test. To me it seems you're that triggering an event triggers the event. You would want to spy on an event if it is triggered by something else like from a calling function.

chrishanes: I was bored

StupidCodeFactory: I think it's safe to remove the spy and the extra test

PR #724, opened 2017-02-20
Generate random human friendly visit ids

alan:
Generate a random [Crockford::Base32](http://www.crockford.com/wrmg/base32.html) id when a visit is created and use the new id in the API.

It also adds a task to populate old visits.

This PR does not use the id to the emails as there could be a small window during deploy when sidekiq could send emails with a nil id.

### DIFF
--- a/lib/human_readable_id.rb
+++ b/lib/human_readable_id.rb
@@ -1,0 +1,21 @@
+class HumanReadableId
+  MAX_NUMBER = Base32::Crockford.decode('ZZZZZZZZ') # Largest 8 digit number
+
+  def self.update_unique_id(klass, primary_key, column)
+    ActiveRecord::Base.transaction(requires_new: true) do
+      generate_id do |candidate_id|
+        klass.
+          where(id: primary_key, column => nil).
+          update_all(column => candidate_id)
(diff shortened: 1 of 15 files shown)


lib/human_readable_id.rb
```diff
@@ -0,0 +1,26 @@
+class HumanReadableId
+  ID_LENGTH = 8
+  MAX_NUMBER = Base32::Crockford.decode('Z' * ID_LENGTH) # Largest 8 digit number
+
+  def self.update_unique_id(klass, primary_key, column)
+    ActiveRecord::Base.transaction(requires_new: true) do
```
StupidCodeFactory: Maybe add a comment about the ```requires_new```.
My assumption is that when this is ran within the booking request creator it's already within a transaction but we may run this in another context without a transaction and this would ensure there is a transaction?

alan: No, it doesn't require an outer transaction. It just ensures it is contained within a transaction. I don't think it is necessary to add a comment as it is explained well in the Rails docs.

StupidCodeFactory: yes that's what i meant

PR #726, opened 2017-02-22
Start using the visits human_id

alan:
Use the human_id on emails and the api

### DIFF
--- a/app/views/visitor_mailer/booked.html.erb
+++ b/app/views/visitor_mailer/booked.html.erb
@@ -103,5 +103,5 @@
 </p>
 
 <p>
-  <small><%= t(".visit_id") %> <%= @visit&.id %></small>
+  <small><%= t(".visit_id") %> <%= @visit.human_id %></small>
 </p>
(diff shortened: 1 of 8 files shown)


app/views/visitor_mailer/booked.html.erb
```diff
@@ -103,5 +103,5 @@
 </p>
 
 <p>
-  <small><%= t(".visit_id") %> <%= @visit&.id %></small>
+  <small><%= t(".visit_id") %> <%= @visit.human_id %></small>
```
StupidCodeFactory: I'm wondering wether it is worth keeping the ```&.```?

alan: I was wondering why it was there in the first place, all visits should have an `id` or `human_id`. If there is a good reason for I can put it back here as well as changing the rest of the emails

StupidCodeFactory: it's for the email preview i think as we rebuild a new visit object

StupidCodeFactory: @alan ^

alan: I've just tried and it works

PR #728, opened 2017-02-23
Prison availability in api for public

StupidCodeFactory:
**CHANGES:**

Adds prison availability to slots


app/services/prisoner_slot_availability.rb
```diff
@@ -11,17 +12,24 @@ def initialize(prison, noms_id, date_of_birth,
...
 
     results = all_slots.deep_dup.each { |slot, unavailability_reasons|
-      unless offender_availabilities_dates.include?(slot.to_date)
+      slot_date = slot.to_date
+      unless offender_availabilities_dates.include?(slot_date)
         unavailability_reasons << PRISONER_UNAVAILABLE
       end
-    }
 
+      unless bookable_prison_slots.include?(slot_date)
```
alan: For prison availability we want to check against the specific slot time instead of the slot date as the granularity is per slot.

app/services/prisoner_slot_availability.rb
```diff
@@ -44,6 +52,15 @@ def prison_slots
     ).to_a
   end
 
+  def bookable_prison_slots
+    @bookable_prison_slots ||=
+      begin
+        api_slot_availability = ApiSlotAvailability.new(
+          prison: prison, use_nomis_slots: true)
+        api_slot_availability.slots.map(&:to_date)
```
alan: This will need to go as we want the slot granularity that the API gives us

config/initializers/public_availabilty.rb
```diff
@@ -0,0 +1,2 @@
+Rails.configuration.public_prisons_with_slot_availability =
```
alan: I think this should be an environment variable as we want to have different lists per environment. This is how the staff equivalent is also done

app/services/nomis/api_slot.rb
```diff
@@ -11,7 +11,7 @@ class ApiSlot
     attribute :visitors_booked, Integer
     attribute :adults_booked, Integer
 
-    delegate :to_s, :to_date, to: :time
+    delegate :to_s, :to_date, :begin_at, to: :time
```
alan: I don't think we need to delegate `begin_at` any longer right?

app/services/prisoner_slot_availability.rb
```diff
@@ -74,4 +91,10 @@ def end_date
     # ensures the range does not go over the 28 days constraint
     @end_date ||= [@end_date, start_date + 28.days].min
   end
+
+  def in_trial?
```
alan: The naming of this method, and class I find very confusing.

We have the class named `PrisonerSlotAvailability` but with this change is no longer just does prisoner availability.

Then the name of this method is generic so made me thing it was related to prisoner availability to it does prison availability.

We have a class called `ApiSlotAvailability`, I think that is used on the deprecated endpoint, that could be a good name for this class?

app.json
```diff
@@ -32,6 +32,12 @@
     "NOMIS_PUBLIC_PRISONER_AVAILABILITY_ENABLED": {
       "required": true
     },
+    "STAFF_PRISONS_WITH_SLOT_AVAILABILITY": {
```
alan: These should be added to the READ<E

spec/services/slot_availability_spec.rb
```diff
@@ -0,0 +1,214 @@
...
+            "2017-02-21T14:00/16:10" => [],
+            "2017-02-27T14:00/16:10" => [],
+            "2017-02-28T09:00/10:00" => ['prison_unavailable'],
+            "2017-02-28T14:00/16:10" => []
+          )
+        end
+      end
+    end
+
+    describe 'with a prison not is trial' do
```
doctorpod: Did you mean 'with a prison not in trial'?

spec/services/slot_availability_spec.rb
```diff
@@ -0,0 +1,214 @@
...
+      describe 'and prisoner availability enabled' do
+        before do
+          allow(Nomis::Api.instance).
+            to receive(:offender_visiting_availability).
+            and_return(prisoner_availability)
+        end
+
+        it 'applies the prisoner availability' do
+          expect(subject.slots).to eq(
```
doctorpod: Even with prison not in trial?

PR #739, opened 2017-03-09
Staff prisoner unavailability api

alan:


config/locales/en/prison_views.yml
```diff
@@ -114,7 +114,10 @@ en:
         email_preview: Preview Email
         prisoner_available: Prisoner available
         slot_available: Space available
-        prisoner_not_available: Prisoner unavailable
+        prisoner_banned: Visits banned
+        out_of_vo: No visiting allowance
```
doctorpod: Shouldn't this be OUT_OF_VA rather than OUT_OF_VO?

alan: Don't think so, VO is Visiting Order. This concept is reference as `vo` in other parts of the code

app/models/concrete_slot.rb
```diff
@@ -56,4 +56,8 @@ def duration
   def <=>(other)
     to_s <=> other.to_s
   end
+
+  def overlaps?(other)
+    end_at > other.begin_at
+  end
```
alan: No, if s1 is 10:00 to 11:00 and other if 12:00 to 13:00 then `other.end_at > begin_at` is true but it does not overlap.

app/models/prisoner_availability_validation.rb
```diff
@@ -21,46 +27,43 @@ def unknown_result?
...
-        min = Date.tomorrow
-        max = 60.days.from_now.to_date
-        requested_dates.select { |date| date.between?(min, max) }
-      end
+  def valid_requested_slots
+    @valid_slots ||= requested_slots.select { |slot| valid_slot?(slot) }
+  end
+
+  def valid_slot?(slot)
+    slot.to_date.between?(Date.tomorrow, 60.days.from_now.to_date)
```
alan: Agree in principle, although I'm not sure if this is going to make things clearer.

The tomorrow constraint is because we don't want to consider dates of today, although the api accepts today, and the 60 days is an API requirement, we could put them as constants but then we would be adding indirection and we would lose clarity on what a valid slot is easily as the constants would be in different classes. I think in this case since these numbers are only used in this method it is actually not an issue. What do you think?

app/models/concrete_slot.rb
doctorpod: OK

app/models/prisoner_availability_validation.rb
doctorpod: OK - leave as is

config/locales/en/prison_views.yml
doctorpod: OK

doctorpod:
Clear for merge - APPROVED

PR #749, opened 2017-03-20
Apply NOMIS prisoner unavailability reasons to check boxes

doctorpod:


app/decorators/visit_decorator.rb
```diff
@@ -26,7 +29,9 @@ def slots
   end
 
   def rejection
-    @rejection ||= (object.rejection || object.build_rejection).decorate
+    @rejection ||= (object.rejection || object.build_rejection).decorate.tap do |rej|
+      rej.apply_nomis_reasons(self)
```
doctorpod: On the second point. `VisitDecorator.prisoner_details_incorrect?` is explicitly called from `RejectionDecorator` - since it doesn't make sense to move this method to `NomisChecker` probably best leave the delegations as they are and continue to pass `self`.

app/services/staff_nomis_checker.rb
```diff
@@ -72,8 +73,27 @@ def slots_unavailable?
     end
   end
 
+  def no_allowance?
```
doctorpod: Doesn't seem a good idea as they refer a lot to the slot errors stored locally - if they were moved to `RejectionDecorator` they would have to reach across the object model and break the Law of Dementer.

alan: The issue is that this logic is specific to the form in the view, it's not a responsibility at this layer.

The logic of returning false if slots are unavailable, or checking any of the slots errors to decide true or false is specific to how we want the form to work and is not a reusable anywhere else so it makes more sense to be in the decorator.

Maybe the solution is to have methods in the checker like this:

`prisoner_banned?(slot)` and then move the rest of the logic to the decorator.

What do you think?

PR #756, opened 2017-03-22
Custom puma config for heroku

alan:
Trying to lower memory consumption

### DIFF
--- a/config/puma_heroku.rb
+++ b/config/puma_heroku.rb
@@ -1,0 +1,17 @@
+# These environment variables are Heroku specific so are not referenced in the
+# Readme
+workers Integer(ENV['WEB_CONCURRENCY'] || 1)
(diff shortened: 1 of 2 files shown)


config/puma_heroku.rb
```diff
@@ -0,0 +1,15 @@
+workers Integer(ENV['WEB_CONCURRENCY'] || 2)
+threads_count = Integer(ENV['RAILS_MAX_THREADS'] || 5)
```
doctorpod: Are these env vars in the readme?

config/puma_heroku.rb
```diff
@@ -0,0 +1,15 @@
...
+threads threads_count, threads_count
+
+preload_app!
+
+rackup      DefaultRackup
+port        ENV['PORT']     || 3000
+environment ENV['RACK_ENV'] || 'development'
```
doctorpod: ...and these?

PR #766, opened 2017-03-27
Fixes percentiles BST bug

alan:
The view used to calculate the percentiles wasn't taking into account the
daylight savings so visits created just after midnight.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170302081707) do
+ActiveRecord::Schema.define(version: 20170327084830) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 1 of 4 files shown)


db/schema.rb
```diff
@@ -371,22 +371,6 @@
     GROUP BY v.prison_name, v.prison_id, v.processing_state, v.timely, (v.created_at)::date;
   SQL
 
-  create_view :percentiles_by_prison_and_calendar_dates, materialized: true,  sql_definition: <<-SQL
-      SELECT prisons.name AS prison_name,
-      prisons.id AS prison_id,
-      (v.created_at)::date AS date,
```
alan: Time was being cast into a date without taking into account the timezone

db/schema.rb
```diff
@@ -512,4 +496,20 @@
     GROUP BY prisons.name;
   SQL
 
+  create_view :percentiles_by_prison_and_calendar_dates, materialized: true,  sql_definition: <<-SQL
+      SELECT prisons.name AS prison_name,
+      prisons.id AS prison_id,
+      (timezone('Europe/London'::text, (v.created_at)::timestamp with time zone))::date AS date,
```
alan: Take into account the timezone before casting

PR #770, opened 2017-04-04
Timebox multiple requests

doctorpod:


spec/services/timebox_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+  let(:result) do
+    subject.run(-> { :fellback }) do
+      sleep 0.2
+      :finished
+    end
+  end
+
+  context '#run' do
+    context 'big timebox' do
+      let(:seconds) { 10 }
```
alan: Are there other ways we could test this without sleeping 10 seconds?

doctorpod: We don't actually wait 10 seconds - just the 0.2 seconds in the main block (line 8) - the 10 seconds is the param we put in when we instantiate Timebox to ensure we'll never trigger a fallback.

alan: 👍

app/services/timebox.rb
```diff
@@ -0,0 +1,21 @@
+# An object that runs a provided block within the time limit or else returns
+# the result of the fallback block. run may be called multiple times; the
+# original time limit being applied to all runs collectively.
+class Timebox
+  def initialize(time_limit)
+    @deadline = Time.now.to_i + time_limit
+  end
+
+  def run(fallback_block)
+    time_remaining = @deadline - Time.now.to_i
```
mloughran: Using `to_i` on the time will lead to very inconsistent timeouts being applied in practice, depending on the fractional component of the time when the deadline and current time are recorded.

app/controllers/api/slots_controller.rb
```diff
@@ -1,10 +1,16 @@
...
     def index
       prison = Prison.enabled.find(params.require(:prison_id))
-      @slots = SlotAvailability.new(
-        prison, prisoner_number, date_of_birth, start_date..end_date
-      ).slots
+      slot_availability = SlotAvailability.new(
+        prison, prisoner_number, date_of_birth, start_date..end_date)
+      timebox = Timebox.new(TIMEBOX_LIMIT)
+
+      @slots = timebox.run(-> { slot_availability.all_slots }) {
```
alan: Would be good to add a controller test to check the fallback response

PR #773, opened 2017-04-06
Fetch contact list from Nomis

alan:


app/decorators/visit_decorator.rb
```diff
@@ -7,6 +7,9 @@ class VisitDecorator < Draper::Decorator
     :prisoner_availability_unknown?,
     :slot_availability_unknown?,
     :slots_unavailable?,
+    :contact_list_unknown?,
```
alan: @chrishanes these are the methods you can use from the view with the `@visit` object

app/services/nomis/contact.rb
```diff
@@ -0,0 +1,32 @@
...
+    attribute :restrictions, Array[ContactRestriction]
+
+    def approved?
+      approved_visitor
+    end
+
+    def active?
+      active
+    end
+
```
alan: @chrishanes methods on each contact that you can use to populate the banned form fields

config/application.rb
```diff
@@ -108,5 +108,8 @@ class Application < Rails::Application
 
     config.public_prisons_with_slot_availability =
       ENV['PUBLIC_PRISONS_WITH_SLOT_AVAILABILITY']&.split(',')&.map(&:strip) || []
+
+    config.staff_prisons_with_nomis_contact_list =
+      ENV['STAFF_PRISONS_WITH_NOMIS_CONTACT_LIST']&.split(',')&.map(&:strip) || []
```
StupidCodeFactory: There seem to be an emerging pattern here, a encapsulating this behaviour in a united tested class might not have been a bad idea.

app/services/prisoner_contact_list.rb
```diff
@@ -0,0 +1,29 @@
+class PrisonerContactList
+  delegate :approved, to: :contact_list
```
StupidCodeFactory: that's a bit convoluted. I was looking for that method missed this lines a few times.

PR #774, opened 2017-04-06
Roll out Timebox to other API calls that access NOMIS

doctorpod:


app/controllers/api/available_slots_controller.rb
```diff
@@ -3,14 +3,15 @@ class AvailableSlotsController < ApiController
...
 
-      slot_availability = ApiSlotAvailability.new(
-        prison: prison,
-        use_nomis_slots: false
-      )
-      slot_availability.restrict_by_prisoner(
-        prisoner_number: params.fetch(:prisoner_number),
-        prisoner_dob: Date.parse(params.fetch(:prisoner_dob))
-      )
+      slot_availability = ApiSlotAvailability.new(prison: prison,
```
alan: Is this controller still being used? I think it might be dead code...

app/services/timebox.rb
```diff
@@ -1,13 +1,13 @@
...
+# optional fallback block. run may be called multiple times; the original time
+# limit being applied to all runs collectively.
 # Time limit may be an integer or a float.
 class Timebox
   def initialize(time_limit_seconds, start_time = Time.now.to_f)
     @deadline = start_time + time_limit_seconds
   end
 
-  def run(fallback_block)
+  def run(fallback_block = nil)
```
alan: Where's the `nil` option being used?

app/controllers/api/available_slots_controller.rb
```diff
@@ -3,14 +3,15 @@ class AvailableSlotsController < ApiController
...
+                                                  use_nomis_slots: false)
+
+      Timebox.new(TIMEBOX_LIMIT).run do
```
alan: There is no fallback here, so what is this doing if it times out?? If the API is still being used then it should return the prison hardcoded availability

doctorpod: If you look at the block being passed in, it calls `restrict_by_prisoner`, a command style method (rather than a query) - no return result, it just populates `slots` in the `slot_availability` instance. If this times out we don't need to return anything - `slots` just doesn't get populated and retains the vanilla info that it acquired during `initialize`.

app/services/timebox.rb
doctorpod: The nil option allows it to be called with no fallback block - as from `AvailableSlotsController` - having the fallback block optional was suggested in Martyn's original prototype if you remember - I didn't think we'd need it until now.

app/controllers/api/available_slots_controller.rb
doctorpod: `config/routes` line 66

PR #776, opened 2017-04-10
Add Secure Headers + some script tags hash

StupidCodeFactory:
**CHANGES:**
Adds Secure Headers to pvb2

**NOTES:**
Modernizr need to be updated  for the PR to be complete


app/assets/javascripts/modules/moj.AsyncGA.js
```diff
@@ -0,0 +1,23 @@
+/*global ga */
+(function() {
+  'use strict';
+
+  moj.Modules.AsyncGA = {
+    el: '.js-AsyncGA',
```
StupidCodeFactory: @doctorpod this JS is to be able to fire those analytics dynamically using data-attributes
and replace the dynamic content inside the script tags:
https://github.com/ministryofjustice/prison-visits-2/pull/776/files#diff-9599427925097c3c66f26ac1e0de5cadL110

PR #780, opened 2017-04-13
Add search by visit human ID

doctorpod:


app/services/estate_visit_query.rb
```diff
@@ -52,4 +69,14 @@ def grouped_visits(visits)
...
       end
   end
+
+  def by_prisoner_number(number, visits)
+    number = Prisoner.normalise_number(number)
+    visits.joins(:prisoner).where(prisoners: { number: number }).to_a
+  end
+
+  def by_human_id(id, visits)
+    id = Prisoner.normalise_number(id)
```
alan: using the `normalise_number` method for the human_id makes sense given the implementation, however it looks out of place in this class. I think it should probably be implemented as a helper method used for searching.

app/services/estate_visit_query.rb
```diff
@@ -14,18 +14,35 @@ def visits_to_print_by_slot(date)
...
+  end
+
+  def requested(query: nil)
+    visits = Visit.preload(:prisoner, :visitors).
+             with_processing_state(:requested).
+             from_estates(@estates).
+             order('created_at asc')
+
+    return visits.to_a unless query
+    by_prisoner_number(query, visits) + by_human_id(query, visits)
```
alan: I'd combine the 2 queries using a SQL `OR`, that way we'll avoid making an extra db call.

app/services/estate_visit_query.rb
```diff
@@ -14,18 +14,35 @@ def visits_to_print_by_slot(date)
...
-    visits.to_a
+  def cancelled(query: nil)
+    visits = Visit.preload(:prisoner, :visitors, :cancellation).
+             joins(:cancellation).
+             from_estates(@estates).
+             where(cancellations: { nomis_cancelled: false }).
+             order('created_at asc')
+
+    return visits.to_a unless query
+    by_prisoner_number(query, visits) + by_human_id(query, visits)
```
alan: Same here, we can use an `OR` query.

app/services/estate_visit_query.rb
```diff
@@ -14,18 +14,35 @@ def visits_to_print_by_slot(date)
...
              processed.
              from_estates(@estates).
              order('visits.updated_at desc').limit(limit)
 
-    if prisoner_number.present?
-      number = Prisoner.normalise_number(prisoner_number)
-      visits = visits.joins(:prisoner).where(prisoners: { number: number })
-    end
+    return visits.to_a unless query
+    by_prisoner_number(query, visits) + by_human_id(query, visits)
```
alan: See comment below about using an `OR` query.

app/views/prison/visits/show.html.erb
```diff
@@ -13,8 +13,11 @@
         <div class="tag tag--<%= @visit.processing_state %> tag--heading"><%= @visit.processing_state.capitalize %>: <br><%= @visit.processed_at.to_date.to_s(:short_nomis) %> - <%= @visit.processed_at.to_s(:time) %></div>
       </div>
       <div class="column-one-third">
+        <% if @visit.human_id -%>
```
alan: All visits have human ids in production so the check can go. You might need to generate human ids for your local database

PR #785, opened 2017-04-18
Try to avoid Race conditions in Nomis::Api

StupidCodeFactory:
Store the @pool instance at class level rather than the instance level


alan:
Closing this as we've found out that the issue is related to the recent timebox

PR #794, opened 2017-04-24
add location service

StupidCodeFactory:


app/views/prison/visits/_prisoner_details.html.erb
```diff
@@ -50,5 +50,11 @@
   <div class="column-one-quarter">
     <span class="text-secondary"><%= t('.prison_name') %></span>
     <div class="bold-small"><%= @visit.prison_name %></div>
+    <% case @visit.location_confirmed? %>
```
StupidCodeFactory: @chrishanes that's where you should hook into

app/services/nomis/api.rb
```diff
@@ -24,6 +24,19 @@ def initialize
       end
     end
 
+    def lookup_offender_location(offender_id:)
+      response = @pool.with do |client|
+        client.get("/offenders/#{offender_id}/location")
+      end
+    rescue Nomis::APIError => e
+      Raven.capture_exception(e, fingerprint: %w[nomis api_error])
+      PVB::Instrumentation.append_to_log(lookup_offender_location: false)
```
alan: How about renaming the log key to `lookup_offender_location_error`?

app/services/nomis/api.rb
```diff
@@ -24,6 +24,19 @@ def initialize
...
+      Nomis::Establishment.new
+    else
+      PVB::Instrumentation.append_to_log(lookup_offender_location: true)
+      Nomis::Establishment.new(response.fetch('establishment') { {} })
```
alan: How come the fetch is needed? Does the api not return a response with an `establishment` key when is successful?

app/views/prison/visits/_prisoner_details.html.erb
alan: I think `location_confirmed` will need to be delegated in the visit decorator for this to work

app/services/nomis/api.rb
```diff
@@ -24,6 +24,19 @@ def initialize
       end
     end
 
+    def lookup_offender_location(offender_id:)
+      response = @pool.with do |client|
+        client.get("/offenders/#{offender_id}/location")
+      end
+    rescue Nomis::APIError => e
+      Raven.capture_exception(e, fingerprint: %w[nomis api_error])
```
alan: How about renaming this log key to `lookup_offender_location_error`?

app/services/nomis/api.rb
alan: Does the api not return a response with an `establishment` key when is successful? I think the fetch might be unnecessary.

app/views/prison/visits/_prisoner_details.html.erb
alan: I think this will need to be delegated in the visit decorator to work

app/models/prisoner_validation.rb
```diff
@@ -1,6 +1,7 @@
 class PrisonerValidation
-  UNKNOWN            = 'unknown'.freeze
-  PRISONER_NOT_EXIST = 'prisoner_does_not_exist'.freeze
+  UNKNOWN                   = 'unknown'.freeze
+  PRISONER_NOT_EXIST        = 'prisoner_does_not_exist'.freeze
+  PRISONER_LOCATION_INVALID = 'prisoner_location_not_valid'.freeze
```
alan: The validation for location needs the `unknown` state. For example the prisoner could be found and then the location api could timeout. All of the other checks follow the pattern of having 3 states: valid, invalid and unknown

app/models/prisoner_validation.rb
```diff
@@ -9,18 +10,35 @@ def initialize(offender)
     self.offender = offender
   end
 
+  # rubocop:disable Metrics/MethodLength
   def active_offender_exists
-    unless Nomis::Api.enabled? && offender.api_call_successful?
```
alan: Are we checking for the api being enabled somewhere else?

app/services/api_prisoner_checker.rb
```diff
@@ -21,7 +14,28 @@ def error
...
+
+  def fetch_offender
+    Nomis::Api.instance.lookup_active_offender(
+      noms_id:       noms_id,
+      date_of_birth: date_of_birth
+    )
+  end
+
+  def null_offender
+    @null_offender ||= Nomis::NullOffender.new(api_call_successful: false)
```
alan: No need to memoize since this will get called at most once.

Unrelated, but I wonder if the null offender could be a singleton object.

app/services/api_prisoner_checker.rb
```diff
@@ -1,14 +1,7 @@
 class ApiPrisonerChecker
   def initialize(noms_id:, date_of_birth:)
-    @offender = if Nomis::Api.enabled? &&
```
alan: Same here, is the api being disabled check gone?

app/services/api_prisoner_checker.rb
```diff
@@ -1,14 +1,7 @@
 class ApiPrisonerChecker
```
alan: Are the changes in this class mostly a refactor? It's confusing since this class is unrelated for the location service work.

app/services/slot_availability.rb
```diff
@@ -29,7 +29,7 @@ def all_slots
 
 private
 
-  attr_reader :prison, :noms_id, :date_of_birth, :start_date, :end_date
```
alan: How come these objects can't be value objects anymore? I don't see any need for them to be mutable.

app/services/slot_availability.rb
```diff
@@ -76,8 +75,7 @@ def offender_availability_dates
   end
 
   def nomis_public_prisoner_availability_enabled?
-    Nomis::Api.enabled? &&
```
alan: The api check is gone here also

app/services/staff_nomis_checker.rb
```diff
@@ -6,8 +6,7 @@ class StaffNomisChecker
   NOT_LIVE = 'not_live'.freeze
 
   def initialize(visit)
-    @visit = visit
-    @nomis_api_enabled = Nomis::Api.enabled?
```
alan: This class also removes the check for the nomis api

app/views/prison/visits/_prisoner_details.html.erb
```diff
@@ -50,5 +50,10 @@
   <div class="column-one-quarter">
     <span class="text-secondary"><%= t('.prison_name') %></span>
     <div class="bold-small"><%= @visit.prison_name %></div>
-  </div>
+    <% if @visit.location_verified? %>
```
alan: It's missing the third state `unknown`

spec/controllers/api/validations_controller_spec.rb
```diff
@@ -12,16 +12,16 @@
   end
 
   describe 'visitors' do
-    let!(:prison) { FactoryGirl.create(:prison) }
+    let!(:prison)   { create(:prison) }
```
alan: Why are these changes in this PR? How is it related to the location service?

app/services/slot_availability.rb
StupidCodeFactory: It's not changed, this change just prevent(ish) accessing the instance variable directly and encourages using accessor methods (initialize method).
I don't mind reverting it, in my opinion it's safer to mutate instance variables to through accessor methods.

app/services/api_prisoner_checker.rb
StupidCodeFactory: if need to be able to have a different id

app/services/slot_availability.rb
alan: Ah, yes, I missed it was private. Ignore my comment.

app/services/api_prisoner_checker.rb
StupidCodeFactory: this does not really hurt and would be the same as accessing the instance variable directly which we do all over the place

app/models/prisoner_validation.rb
```diff
@@ -1,14 +1,23 @@
 class PrisonerValidation
-  UNKNOWN            = 'unknown'.freeze
-  PRISONER_NOT_EXIST = 'prisoner_does_not_exist'.freeze
+  VALID                     = 'ok'.freeze
```
alan: This is not used I think

app/models/prisoner_validation.rb
```diff
@@ -17,10 +26,30 @@ def active_offender_exists
 
     unless offender.valid?
       errors.add :base, PRISONER_NOT_EXIST
+      return
+    end
+
+    unless prisoner_location.api_call_successful?
```
alan: I think here we need to check for the offender being valid before calling the prisoner_location

app/services/staff_nomis_checker.rb
```diff
@@ -14,10 +18,10 @@ def prisoner_existance_status
     return NOT_LIVE unless prisoner_check_enabled?
 
     case prisoner_existance_error
-    when nil
-      VALID
-    when UNKNOWN
-      UNKNOWN
+    when nil              then VALID
```
alan: I find this confusing; it mixes the prisoner existence with location and depends on the first sorted error message.

The location is almost the same as the prisoner availability in terms of how it interacts with prisoner existence:

- If prisoner existence is invalid then location is unknown: The user could have typoed the noms_id but still be valid (and existence be invalid because the date of birth does not match). If we lookup the location then it will be the location of a completely different prisoner.

- If the prisoner existence is unknown then the location is unknown. Similar to above but the existence api times out, responds with a non 200 code.

- Prisoner existence is valid, then location could be valid, unknown or invalid

app/views/prison/visits/_prisoner_details.html.erb
```diff
@@ -7,21 +7,23 @@
...
+<% when StaffNomisChecker::UNKNOWN, StaffNomisChecker::LOCATION_UNKNOWN %>
   <div class="notice push-top">
     <i class="icon icon-important">
       <span class="visuallyhidden"><%= t('warning', scope: :shared) %></span>
     </i>
     <strong class="bold-small"><%= t(".#{@visit.prisoner_existance_error}_text") %></strong>
   </div>
 <% end %>
+
+
```
alan: ✂️ ?

PR #803, opened 2017-05-03
Update to latest omniauth-oauth2

alan:


lib/mojsso.rb
```diff
@@ -27,6 +27,12 @@ class Mojsso < OmniAuth::Strategies::OAuth2
       def raw_info
         @raw_info ||= access_token.get('/api/user_details').parsed
       end
+
+      # As suggested here: https://github.com/intridea/omniauth-oauth2/commit/26152673224aca5c3e918bcc83075dbb0659717f#commitcomment-19809835
+      # Other link about the issue: https://github.com/intridea/omniauth-oauth2/issues/81
```
StupidCodeFactory: That is not very descriptive. Event after reading the two links I still find it hard to understand the issue in pvb2

StupidCodeFactory: @alan ^

alan: Fair enough, I'll open a PR explaining what breaks without this and a bit about the issue.

PR #813, opened 2017-05-10
Fix GA analytics pageview

chrishanes:


app/assets/javascripts/modules/moj.AsyncGA.js
```diff
@@ -13,9 +13,9 @@
       window.ga('create', gaTrackingId, 'service.gov.uk');
 
       if (hitTypePage) {
-        window.ga('send', 'page_view', location.pathname + '#' + hitTypePage);
+        window.ga('send', 'pageview', location.pathname + '#' + hitTypePage);
```
StupidCodeFactory: ?? huh that's odd I remember fixing that a little while ago. Could we add test around this before it breaks again?

chrishanes: I think you fixed on Public

StupidCodeFactory: Ah :) that would make sense yeah

chrishanes: Can you approve now? Not sure we need a test (can't see this file being looked at again)?

PR #814, opened 2017-05-11
Store contact nomis id on visitor

alan:
- Pre-requisit for booking the visit to nomis. We'll be needing the visitors nomis
id when using the API to book visits.

- Preserves the selected contact after a validation errors.

- It also introduces a new feature spec with the contact list enabled, the plan is
to extend this test to test the booking to nomis.

### DIFF
--- a/spec/features/process_a_request_accept_spec.rb
+++ b/spec/features/process_a_request_accept_spec.rb
@@ -6,11 +6,78 @@ RSpec.feature 'Processing a request - Acceptance', js: true do
 
   include_context 'process request setup'
 
-  around do |ex|
-    travel_to(Date.new(2016, 12, 1)) { ex.run }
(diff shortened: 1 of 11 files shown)


spec/features/process_a_request_accept_spec.rb
```diff
@@ -6,26 +6,81 @@
 
   include_context 'process request setup'
 
-  around do |ex|
-    travel_to(Date.new(2016, 12, 1)) { ex.run }
+  context 'with visitor contact list', vcr: { cassette_name: 'process_happy_path_with_contact_list' } do
+    # Leicester has contact list enabled in 'test'
```
StupidCodeFactory: I would have preferred to keep the test setup in the test file and avoid:
```ruby
config.staff_prisons_with_nomis_contact_list = ['Leicester']
```
under config/envornments/test.rb

PR #822, opened 2017-05-16
Rename BookingResponse to StaffResponse

alan:
The naming has been always been confusing in this part of the code. StaffResponse response is clearer, at least to me, because it indicates that is what staff have responded, which is not the same as the final booking response.

The reason to do this now is that I'd like to use BookingResponse for the book to nomis feature.


app/controllers/prison/visits_controller.rb
```diff
@@ -1,24 +1,24 @@
...
+  include StaffResponseContext
   before_action :authorize_prison_request
   before_action :authenticate_user
   before_action :cancellation_reason_set, only: :cancel
   before_action :visit_is_processable, only: [:process_visit, :update]
 
   def process_visit
-    @visit            = load_visit.decorate
-    @booking_response = BookingResponse.new(visit: @visit)
+    @visit = load_visit.decorate
```
StupidCodeFactory: :(

PR #823, opened 2017-05-16
Display a message when a visit will be rejected

chrishanes:


app/views/prison/visits/_visitor_details.html.erb
```diff
@@ -5,14 +5,14 @@
 
 <% if Nomis::Feature.contact_list_enabled?(@visit.prison_name) && !@visit.contact_list_unknown? %>
```
chrishanes: This has been changed in https://github.com/ministryofjustice/prison-visits-2/pull/864 which needs to be merged before release

app/views/prison/visits/_issue_with_prisoner_section.html.erb
```diff
@@ -64,13 +64,13 @@
...
           <%= t('.prisoner_released') %>
         </label>
       <% end %>
       <%= error_container(rf, :reasons) do %>
         <label class="block-label">
-          <%= rf.object.checkbox_for(:prisoner_banned, class: 'js-Conditional') %>
+          <%= rf.object.checkbox_for(:prisoner_banned, class: 'js-Rejection', 'data-rejection-el': 'rejection-message') %>
           <%= t('.prisoner_banned') %>
         </label>
       <% end %>
```
doctorpod: There seems to be a repeating pattern here with only the field names changing. Could we have a loop go through an array of field names or use a partial here?

chrishanes: That sounds like a good idea, but beyond my knowledge.

chrishanes: @doctorpod Can we accept the current partial and make a card in Trello inbox for refactor or should we do this now? If the later can you help please?

doctorpod: Lets do it later - I'll make a card

kyliehavelock: @doctorpod can you point me to the card?

PR #826, opened 2017-05-17
Update StaffResponse/BookingResponder interaction

alan:
This change will let us recover from errors when visits get booked to Nomis.
There is no change in behaviour, just a refactor.

The previous interaction relied that StaffReponse would run the validations, and
it it was valid then BookingResponder would execute the action and always succeed.
This will no longer be the case when booking to nomis since the api might fail
on the execution path.

The main change is to introduce a BookingResponse object that represents the
result of executing the action (accepting, rejecting, etc). This object will be
used to abort the transaction, not send emails and render errors back to the
user when booking to nomis.

### DIFF
--- a/app/services/booking_responder.rb
+++ b/app/services/booking_responder.rb
@@ -1,16 +1,20 @@
 class BookingResponder
   delegate :visit, to: :staff_response
 
-  def initialize(staff_response, message = nil)
-    @staff_response = staff_response
-    @message = message
+  def initialize(visit, user: nil, message: nil)
+    self.staff_response = StaffResponse.new(visit: visit, user: user)
+    self.message = message
   end
 
   def respond!
-    return unless visit.requested?
-    processor.process_request(message)
+    unless staff_response.valid?
+      return BookingResponse.new(success: false)
+    end
 
-    send_notifications
+    booking_response = processor.process_request(message)
+
+    send_notifications if booking_response.success?
+    booking_response
   end
 
   def visitor_mailer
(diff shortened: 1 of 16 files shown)


app/services/booking_response.rb
```diff
@@ -0,0 +1,9 @@
+class BookingResponse
+  def initialize(success:)
+    @success = success
+  end
+
+  def success?
```
StupidCodeFactory: Couldn't we just return true or false?

app/services/booking_responder.rb
```diff
@@ -1,16 +1,20 @@
 class BookingResponder
   delegate :visit, to: :staff_response
 
-  def initialize(staff_response, message = nil)
-    @staff_response = staff_response
-    @message = message
+  def initialize(visit:, user: nil, message: nil)
```
alan: They work well here because both user and message are optional, so you get cases where you pass only the user, or only the message, none, or both.

app/services/booking_response.rb
alan: This class is just a stub for now, will get updated properly with the book to nomis. Do you prefer to hardcode it to return true for now?

app/services/booking_responder.rb
StupidCodeFactory: what about this then?
```ruby
BookingResponder.new(visit, user: user, message: message)
```
To avoid passing `nil` for the `user:` if you want to pass a message

app/services/booking_response.rb
StupidCodeFactory: That' s cool for now

app/services/booking_responder.rb
alan: You don't have to pass `nil` for the user if you want to pass just a message:

```ruby
BookingResponder.new(visit: visit, message: message)
```

PR #828, opened 2017-05-17
Split the feature checking from nomis checker

alan:


app/services/nomis/feature.rb
```diff
@@ -0,0 +1,28 @@
...
+        Rails.configuration.nomis_staff_prisoner_check_enabled
+    end
+
+    def self.prisoner_availability_enabled?
+      Nomis::Api.enabled? &&
+        Rails.configuration.nomis_staff_prisoner_availability_enabled
+    end
+
+    def self.slot_availability_enabled?(visit)
+      Nomis::Api.enabled? &&
```
StupidCodeFactory: Would be preferable to pass just prison_name string rather then having to know about the visit class.

app/services/nomis/feature.rb
```diff
@@ -0,0 +1,28 @@
+module Nomis
+  class Feature
```
StupidCodeFactory: Could this just be a module and be injected where needed (service mock spec helper, some views?) rather than having to type `Nomis::Feature` all the time?

alan: I prefer to be explicit of where business domain classes and objects are being called, worth the typing!

app/views/prison/visits/_visitor_details.html.erb
```diff
@@ -24,7 +24,7 @@
     <% end %>
   </ul>
   <div class="form-group">
-    <% if !@visit.contact_list_unknown? %>
+    <% if Nomis::Feature.contact_list_enabled?(@visit.prison_name) && !@visit.contact_list_unknown? %>
```
doctorpod: There is an argument that `Nomis::Feature.contact_list_enabled?(@visit.prison_name) && !@visit.contact_list_unknown?` could be encapsulated into a helper as it's used at least 3 times in this view.

spec/decorators/concrete_slot_decorator_spec.rb
```diff
@@ -92,46 +92,54 @@
...
 
-          it 'renders the checkbox neither verified or errors' do
-            expect(html_fragment).to have_css('label.block-label.date-box.date-box.disabled')
-            expect(html_fragment).to have_css('span.date-box__number', text: '1')
-            expect(html_fragment).to have_css('span.date-box__day',    text: date.strftime('%A'))
-            expect(html_fragment).to have_text("#{slot.to_date.strftime('%e %B %Y')} 14:00–15:30")
-            expect(html_fragment).to have_css("input[disabled='disabled']")
-            expect(html_fragment).not_to have_css('span.tag--error')
-            expect(html_fragment).not_to have_css('span.tag--verified')
+          context 'wiht a date in the past' do
```
doctorpod: wiht -> with

spec/services/staff_nomis_checker_spec.rb
```diff
@@ -102,15 +104,15 @@
...
+      context 'with the feature disabled' do
         before do
           switch_off(:nomis_staff_prisoner_availability_enabled)
         end
 
         it { is_expected.not_to be_prisoner_availability_unknown }
       end
 
-      context 'with NOMIS_STAFF_PRISONER_AVAILABILITY_ENABLED is enabled' do
+      context 'with the featured is enabled' do
```
doctorpod: remove 'is'

PR #829, opened 2017-05-18
[WIP] - Bunde update all the tings

StupidCodeFactory:


StupidCodeFactory:
@chrishanes over to you before I start fixing the tests (they should fix themselves when the radio boxes are fixed)

PR #860, opened 2017-05-22
Empty Contact List

StupidCodeFactory:


spec/services/nomis/api_spec.rb
```diff
@@ -201,13 +201,20 @@
     end
 
     subject { super().fetch_contact_list(params) }
-
+    before do
+      Nomis::Contact.class_eval do
```
alan: This mutates the class for other tests

app/decorators/visit_decorator.rb
```diff
@@ -1,6 +1,7 @@
 class VisitDecorator < Draper::Decorator
   delegate_all
   decorates_association :rejection
+  NO_VISITORS_IN_NOMIS = 'No record of this visitor in NOMIS'.freeze
```
alan: This text should not be hardcoded in english, we've talked to translate staff one day.

PR #861, opened 2017-05-22
Enable one-click booking to NOMIS

alan:


spec/decorators/visit_decorator_spec.rb
```diff
@@ -35,4 +35,76 @@
...
+
+    before do
+    end
+
+    describe 'when the book to nomis feature is disabled' do
+      before do
+        switch_off :nomis_staff_book_to_nomis_enabled
+      end
+
+      it { expect(subject.book_to_nomis_opt_in?).to eq(false) }
```
StupidCodeFactory: 1) Although this is correct the canonical way to check a boolean in RSpec is:

```ruby
it { expect(subject.book_to_nomis_opt_in?).to be false }
```
2) Also when testing predicate methods, RSpec provide a very nice and readable syntactic sugar API:

```ruby
it { is_expected.not_to be_book_to_nomis_opt_in }
```

spec/services/nomis/api_spec.rb
```diff
@@ -185,4 +185,41 @@
...
+      }
+    end
+
+    let(:offender_id) { 1_057_307 }
+
+    subject { super().book_visit(offender_id: offender_id, params: params) }
+
+    context 'happy path', vcr: { cassette_name: 'book_visit_happy_path' } do
+      it 'returns the visit_id' do
+        expect(subject.visit_id).to eq(5_467)
```
StupidCodeFactory: RSpec provides a nice helper to do this more elegantly:
```ruby
expect(subject.book_visit(offender_id: offender_id, params: params)).to have_attributes(visit_id: 5_467)
```

spec/services/nomis/api_spec.rb
```diff
@@ -185,4 +185,41 @@
...
+      end
+
+      it 'instruments the outcome of the call' do
+        subject
+        expect(PVB::Instrumentation.custom_log_items[:book_to_nomis_success]).to eq(true)
```
StupidCodeFactory: I think it is clearer as you directly see what method call is being tested with which arguments and also remove this odd call to subject on the first line of the test:

```ruby
expect {
  subject.book_visit(offender_id: offender_id, params: params)
}.to change { PVB::Instrumentation.custom_log_items[:book_to_nomis_success] }.from(nil).to(true)

File: spec/services/nomis/feature_spec.rb [hunk 4]
```diff
@@ -107,4 +107,32 @@
...
     end
   end
+
+  describe '.book_to_nomis_enabled?' do
+    context 'when the book to nomis flag is disabled' do
+      before do
+        switch_off :nomis_staff_book_to_nomis_enabled
+      end
+
+      it { expect(described_class.book_to_nomis_enabled?(anything)).to eq(false) }
```
StupidCodeFactory:
```ruby
 is_expected.not_to be_book_to_nomis_enabled anything
```

alan: I find the use of predicate methods with variables not very readable when they take parameters. I think is clearer like it is now.

spec/decorators/visit_decorator_spec.rb
alan: The use of predicate methods has the downside that you lose the context of the exact method call. I think the tests end up being more maintainable by using less RSpec sugar methods, my preference would even to use mini_test!

app/views/prison/visits/process_visit.html.erb
```diff
@@ -11,6 +11,7 @@
     <%= hidden_field_tag 'visit[visitor_ids][]', visitor.id  %>
   <% end %>
   <%= f.hidden_field :prison_id  %>
+  <%# f.hidden_field_tag 'prisoner[nomis_offender_id]', @visit.nomis_offender_id %>
```
StupidCodeFactory: Could this be removed?

chrishanes: @alan mentioned we needed the offender_id as a hidden field, but left this commented out.

.rspec
```diff
@@ -1,3 +1,2 @@
 --color
 --require spec_helper
---backtrace
```
StupidCodeFactory: Should we add ```.rspec``` to the ignore file? I defo need the backtrace when doing any rspec testing

alan: Yes, we could do that.

In theory you should be able to add it to your global .rspec file - https://relishapp.com/rspec/rspec-core/v/2-0/docs/configuration/read-command-line-configuration-options-from-files

Out of curiosity, how come you find it useful all the time? I've found it useful very few times

app/services/booking_response.rb
```diff
@@ -1,9 +1,44 @@
...
+  def self.succesful
+    new(nil)
+  end
+
+  def self.process_required
+    new(PROCESS_REQUIRED_ERROR)
+  end
+
+  def self.nomis_validation_error
+    new(NOMIS_VALIDATION_ERROR)
```
StupidCodeFactory: these to `NOMIS_VALIDATION_ERROR` and `NOMIS_API_ERROR` should really be in the Nomis API service class instead of being bundled with other unrelated error messages

app/services/nomis/booker.rb
```diff
@@ -0,0 +1,56 @@
+module Nomis
+  class Booker
+    ALREADY_PROCESSED = 'Duplicate post'.freeze
+
+    def initialize(visit)
+      self.visit = visit
```
StupidCodeFactory: I'm this class doesn't really need to know about the the ActiveRecord Visit model, it would be much nice to pass in the `booking_params` and have the caller know about the know about the visit rather then have both know about the internals of the `Visit` ActiveRecord object

app/views/layouts/application.html.erb
```diff
@@ -64,9 +64,9 @@
     <% end %>
 
     <% if alert.present? %>
-        <p class="notification">
-            <%= alert %>
-        </p>
+      <div class="error-summary" role="group" aria-labelledby="error-summary-heading" tabindex="-1">
```
StupidCodeFactory: How does this relate to booking to nomis? Can you refactor this in a separate PR. Also I can see any unit test for this change.

app/views/prison/visits/process_visit.html.erb
```diff
@@ -68,10 +72,25 @@
         <strong class="bold-small"><%= t('.rejection_warning') %></strong>
       </div>
     </div>
-    <%= f.submit t('.submit'), class: 'button button-primary' %>
-    <div class="push-top">
-      <%= link_to t('.email_preview'), prison_visit_email_preview_path(@visit), class: 'js-LinkPreview', target: '_blank' %>
-    </div>
+    <% if Nomis::Feature.book_to_nomis_enabled?(@visit.prison_name) %>
```
StupidCodeFactory: We should stop treating our views as second class citizens and allow them to become such a mess.
This entire block should be in a decorator and be unit tested

spec/controllers/prison/visits_controller_spec.rb
```diff
@@ -71,6 +71,42 @@
 
         it { is_expected.to redirect_to(prison_inbox_path) }
       end
+
+      context 'when there is an api error when booking to nomis' do
```
StupidCodeFactory: there need to be an outer scope called "When booking to nomis" as it seems to be repeated in the next context block, that would share the setup of `let(:stadd_response)` and `let(:booking_response)`

spec/decorators/visit_decorator_spec.rb
```diff
@@ -11,6 +11,33 @@
...
+
+    context 'when the lookup is enabled' do
+      let(:checker) { double(StaffNomisChecker) }
+      let(:offender) { double(Nomis::Offender, id: 1234) }
+
+      before do
+        switch_on(:nomis_staff_prisoner_check_enabled)
+        mock_service_with(StaffNomisChecker, checker)
+        expect(checker).to receive(:offender).and_return(offender)
+      end
```
StupidCodeFactory: I  think you can have  a cleaner test setup with only:

```ruby
let(:offender) { Nomis::Offender.new(id:1234)
before do
  switch_on(:nomis_staff_prisoner_check_enabled)
  mock_nomis_with(:lookup_active_offender, offender)
end
```
and remove the `StaffNomisChecker` double and the `expect` in the  `before` block

spec/models/visit_spec.rb
```diff
@@ -343,4 +343,40 @@
       end
     end
   end
+
+  describe '#allowed_additional_visitors' do
+    let(:visitor1) { FactoryGirl.build_stubbed(:visitor) }
```
StupidCodeFactory: no need to add the `FactoryGirl` just use the short syntax, it looks a lot less cluttered and we don't have any other conflicting stubbing library.

spec/services/booking_responder/booking_request_processor_spec.rb
```diff
@@ -13,17 +13,53 @@
 
   subject { described_class.new(staff_response) }
 
-  it '#process_request' do
-    expect {
-      subject.process_request message do
-        visit.accept!
+  describe '#process_request' do
+    context 'happy path' do
```
StupidCodeFactory: happy path? Surely is a better way to describe the process. I have not idea what the "happy path" entails, same for the "sad path"

spec/services/booking_response_spec.rb
```diff
@@ -0,0 +1,34 @@
+require 'rails_helper'
+
+RSpec.describe BookingResponse do
+  describe 'succesful' do
```
StupidCodeFactory: you're missing ".xxxxx" in front of all the methods descriptions

spec/services/nomis/api_spec.rb
```diff
@@ -210,4 +210,48 @@
       expect(subject.first.id).to eq(first_contact.id)
     end
   end
+
+  describe 'book_visit' do
```
StupidCodeFactory: `#book_visit`

spec/services/nomis/api_spec.rb
```diff
@@ -210,4 +210,48 @@
...
+        override_restrictions: false,
+        client_unique_ref: 'visit_id_1234'
+      }
+    end
+
+    let(:offender_id) { 1_057_307 }
+
+    subject { super().book_visit(offender_id: offender_id, params: params) }
+
+    context 'happy path', vcr: { cassette_name: 'book_visit_happy_path' } do
```
StupidCodeFactory:
```ruby
context 'With a context description of "sad path"' do
  it { expect(maintainer).to be_a_sad_panda }
end
```

spec/services/nomis/booker_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+  end
+
+  subject { described_class.new(visit) }
+
+  describe '#book' do
+    let(:booking_response) { subject.book }
+
+    describe 'api call to Nomis' do
+      it 'with the correct parameters' do
+        expect(Nomis::Api.instance).
```
StupidCodeFactory: why not extend the `mock_nomis_with` helper  method to accept an extra argument or block to perform argument call check?

spec/services/nomis/feature_spec.rb
```diff
@@ -107,4 +107,32 @@
...
+    end
+
+    context 'when the book to nomis flag is enabled and the visit prison is not on the list' do
```
StupidCodeFactory: there need to be and outer context with the following description:

```ruby
describe 'the visit prison is not on the list' do
  context 'and when the book to nomis flag is enabled' 
  context 'and when the slot availability flag is enabled'
end
```

spec/services/nomis/feature_spec.rb
```diff
@@ -107,4 +107,32 @@
...
+        switch_feature_flag_with(:staff_prisons_with_book_to_nomis, [])
+      end
+
+      it { expect(described_class.book_to_nomis_enabled?(prison_name)).to eq(false) }
+    end
+
+    context 'when the slot availability flag is enabled and the visit prison is not on the list' do
+      before do
+        switch_on :nomis_staff_book_to_nomis_enabled
+        switch_feature_flag_with(:staff_prisons_with_book_to_nomis, [prison_name])
```
StupidCodeFactory: I thought the prison wasn't on the list? Is this a bug?

app/controllers/concerns/staff_response_context.rb
```diff
@@ -47,8 +48,9 @@ def visit_params
         :nomis_id,
         :banned,
         :not_on_list,
-        banned_until: %i[day month year]
-      ]
+        banned_until: [:day, :month, :year]
```
alan: Interesting, I think this must have been a rubocop autocorrection as I didn't change it. I'll revert and re-run rubocop to double check

app/controllers/prison/visits_controller.rb
```diff
@@ -9,18 +9,26 @@ def process_visit
     @visit = load_visit.decorate
   end
 
+  # rubocop:disable Metrics/MethodLength
```
alan: Yes, it is ugly, suggestions welcome. This was something I wanted to get ideas on.

app/controllers/prison/visits_controller.rb
```diff
@@ -9,18 +9,26 @@ def process_visit
...
   def update
     booking_response = booking_responder.respond!
+
     if booking_response.success?
-      flash[:notice] = t('process_thank_you', scope: %i[prison flash])
+      flash[:notice] = t('process_thank_you', scope: [:prison, :flash])
```
alan: @StupidCodeFactory here it has also changed from `%i` to explicit array. I'm pretty sure is rubocop

app/decorators/visit_decorator.rb
```diff
@@ -45,6 +45,23 @@ def processed_at
     @processed_at ||= last_visit_state_change&.created_at || object.created_at
   end
 
+  def contact_list_working?
```
alan: No, this was wip as it dependend of yesterday's discussion. Coming soon with final logic

app/models/visit.rb
```diff
@@ -1,6 +1,10 @@
 class Visit < ActiveRecord::Base
   extend FreshnessCalculations
   include PrincipalVisitor
+
+  attr_writer :book_to_nomis_opt_out
```
alan: We need to get the opt out logic to the booking responder acceptor class, the part that decides to make the api call or not, once is in that part of the code it only has access to the visit. Have you got any suggestions on how to improve this? I agree that is not ideal.

app/models/visit.rb
```diff
@@ -1,6 +1,10 @@
...
+  attr_accessor :offender_id
```
alan: Good spot, I'll take it out

app/models/visit.rb
```diff
@@ -84,6 +89,15 @@ class Visit < ActiveRecord::Base
     end
   end
 
+  def book_to_nomis_opt_out
+    # TODO: Changes in Rails 5 to `ActiveRecord::Type::Boolean.new.cast(string)`
+    ActiveRecord::Type::Boolean.new.type_cast_from_database(@book_to_nomis_opt_out)
```
alan: Yep, setter makes more sense.

Lets take a look together on other ways this could be done as you seem to have other ideas.

app/services/booking_responder/accept.rb
```diff
@@ -5,8 +5,28 @@ def process_request(message = nil)
         visit.rejection = nil
         visit.accept!
 
-        BookingResponse.new(success: true)
+        if visit.book_to_nomis_opt_out?
```
alan: As discussed in person, the problem is not the POST failing, it is succeeding and then the pvb having an error. The idea with this approach is to delay as much as possible the API call to avoid  other errors leaving PVB and Nomis inconsistent.

app/services/booking_response.rb
```diff
@@ -1,9 +1,44 @@
 class BookingResponse
-  def initialize(success:)
-    @success = success
+  ALREADY_PROCESSED_ERROR = 'already_processed'.freeze
+  PROCESS_REQUIRED_ERROR = 'process_required'.freeze
```
alan: I'll add a description on the intent of this class.

This class represents the result of the visit processing. Before the different states included the two above and were implicit. Now the response that staff can get includes extra states related to the result of the Nomis call (validation error / api error)

app/services/booking_response.rb
```diff
@@ -1,9 +1,44 @@
...
+  NOMIS_VALIDATION_ERROR = 'nomis_validation_error'.freeze
+  NOMIS_API_ERROR = 'nomis_api_error'.freeze
+
+  attr_reader :error
+
+  def self.succesful
+    new(nil)
```
alan: Yes, this isn't clear I'll refactor it.

Regarding the pattern I think it's better wrapping the constants in class methods since it avoids getting unexpected values passed in and the actual values are an implementation detail

app/services/nomis/api.rb
```diff
@@ -104,6 +104,20 @@ def fetch_contact_list(offender_id:)
       Nomis::ContactList.new(response)
     end
 
+    def book_visit(offender_id:, params:)
+      response = @pool.with { |client|
+        client.post(
+          "offenders/#{offender_id}/visits/booking", params, idempotent: false
```
alan: Actually, I need to change this. Now the call is idempotent.

app/services/nomis/booker.rb
```diff
@@ -0,0 +1,56 @@
+module Nomis
+  class Booker
```
alan: 👍 for the repository pattern

app/services/nomis/booker.rb
alan: Yeah, I'll pass the arguments necessary for the call instead

app/services/nomis/booker.rb
```diff
@@ -0,0 +1,56 @@
...
+    def call_api
+      self.booking = Nomis::Api.
+        instance.
+        book_visit(offender_id: offender_id, params: booking_params)
+    rescue Nomis::APIError
+      self.api_error = true
+    end
+
+    def build_booking_response
+      return BookingResponse.nomis_api_error if api_error
```
alan: This is related to the fact that we want to delay the API call as much as possible, so this class needs to return a `booking_response`

app/services/nomis/booking.rb
```diff
@@ -0,0 +1,17 @@
...
+  class Booking
+    include NonPersistedModel
+
+    attribute :visit_id, Integer
+    attribute :error_message, String
+
+    def self.build(response)
+      if response.key?('error')
+
+        new(error_message: response['error']['message'])
```
alan: `ActiveModel::Error` is for validations so not a fan of adding outside the validation lifecycle. We use `.build` somewhere else to translate the api call response

app/services/nomis/client.rb
```diff
@@ -27,9 +27,9 @@ def get(route, params = {})
       request(:get, route, params, idempotent: true)
     end
 
-  # def post(route, params)
-  #   request(:post, route, params)
-  # end
+    def post(route, params, idempotent:)
```
alan: In our case is idempotent

app/services/nomis/client.rb
```diff
@@ -80,17 +80,28 @@ def request(method, route, params, idempotent:)
     # rubocop:enable Metrics/MethodLength
     # rubocop:enable Metrics/AbcSize
 
+    def http_method_expects(method)
+      if method == :get
+        [200]
+      else
+        # TODO: Add 409 when idempotency key is ready
```
alan: Reminder: Delete todos.

app/views/layouts/application.html.erb
alan: @chrishanes do you want to give it a go separating this into a different PR?

app/views/prison/visits/_book_to_nomis_opt_in.html.erb
```diff
@@ -0,0 +1,27 @@
...
+    </label>
+  <% else %>
+    <% if book_to_nomis_config.already_booked_in_nomis? %>
+      <%= content_tag :p, t(".already_booked_in_nomis") %>
+    <% else %>
+      <%= content_tag :p, t(".book_to_nomis_not_possible") %>
+    <% end %>
+  <% end %>
+
+  <div id="rejection-message-opposite">
```
alan: @chrishanes this does work as we expect.

Currently if you visit is set to be saved to Nomis when you select a time and then match the visitors these fields appear. Also using the checkbox does not trigger any changes.

The way I thought it was going to work was:
  - when the visit is approved, and the visit is set to be saved in nomis then staff wont see these fields.

  - when the visit is approved, and staff opt out from saving to nomis, or the visit can't be saved to nomis then staff will see these 2 fields

  - prisons that don't have book to nomis feature enabled will see no changes

config/locales/en/prison_views.yml
```diff
@@ -25,9 +26,28 @@ en:
...
+          This visit wasn't booked.
+        </h1>
+        <p>Please either reject or accept the booking</p>
       at_least_one_estate: You must select at least one estate
       invalid_estate_selection: You don't access to these estates
+      nomis_validation_error_html: |
+        <h1 class="heading-medium error-summary-heading" id="error-summary-heading">
+          This visit wasn't booked.
+        </h1>
+        <p>Choose a rejection reason to reject the visit or <a href="#nomis-opt-out">book this visit anyway</a>.</p>
```
alan: @kayesoutham @kyliehavelock I'm wondering if we should tweak this validation error message. If the prisoner has a restriction, such as a Closed restriction, staff won't really know what validation failed. I think it would be helpful to explain that this validation error can happen when the offender has restrictions and to check in Nomis. What do you think?

alan:
@StupidCodeFactory can you review again please?

app/services/booking_responder.rb
```diff
@@ -57,4 +58,17 @@ def bookable?
...
       visit.slot_granted?
   end
+
+  def persist_to_nomis=(val)
+    # TODO: Changes in Rails 5 to `ActiveRecord::Type::Boolean.new.cast(string)`
+    @persist_to_nomis = ActiveRecord::Type::Boolean.new.type_cast_from_database(val)
+  end
+
+  def persist_to_nomis?
+    @persist_to_nomis
```
StupidCodeFactory: this can return nil `!!@persist_to_nomis`

app/views/prison/visits/_book_to_nomis_opt_in.html.erb
```diff
@@ -0,0 +1,29 @@
+<div class="panel" id="nomis-opt-out">
+  <% if book_to_nomis_config.possible_to_book? %>
+    <%= content_tag :p, t(".book_to_nomis_opted_in_by_default") %>
```
StupidCodeFactory:
```erb
<p>
  <%=  t(".book_to_nomis_opted_in_by_default") %>
</p>
```

app/views/prison/visits/_book_to_nomis_opt_in.html.erb
```diff
@@ -0,0 +1,29 @@
...
+    <%= hidden_field_tag :book_to_nomis_opted_in, true %>
+    <fieldset>
+      <div class="multiple-choice push-bottom">
+        <%= check_box_tag :book_to_nomis_opted_in, false, !book_to_nomis_config.opted_in?, class: 'js-BookToNomis', 'data-el': 'booktonomis-message', id: 'book_to_nomis_opt_in' %>
+        <label for="book_to_nomis_opt_in"><%= t('.dont_book_to_nomis') %></label>
+      </div>
+    </fieldset>
+  <% else %>
+    <% if book_to_nomis_config.already_booked_in_nomis? %>
+      <%= content_tag :p, t(".already_booked_in_nomis") %>
```
StupidCodeFactory: same as above

app/models/staff_response.rb
```diff
@@ -22,15 +22,16 @@ class StaffResponse
...
+      except: [
+        :created_at,
+        :updated_at,
+        :slot_granted,
+        :slot_option_0,
+        :slot_option_1,
+        :slot_option_2,
+        :nomis_id,
+        :human_id
+      ],
```
alan: Don't think so, rubocop is passing

PR #864, opened 2017-05-22
[WIP] - Check if the contact list is unknown

chrishanes:
I think there was a typo which resulted in the API down message being displayed when it was up (and visa versa).


StupidCodeFactory:
This feels like it would need a test. 
Since it's really hard to test would need to be able to control the API returning a 502 in the middle of the test. We are probably better off wrapping this presentational code in a Decorator.
Views really should not have conditionals in the first place.

PR #869, opened 2017-05-24
Validate lead visitor is on the contact list

StupidCodeFactory:
**CHANGES:**
This now validates that the lead visitor on the list and not banned otherwise it rejects the booking


config/locales/en/errors.yml
```diff
@@ -50,3 +50,4 @@ en:
   staff_response:
     errors:
       must_reject_or_accept_visit: "A visit must either be accepted by choosing a date or at least one rejection reason"
+      lead_visitor_not_on_contact_list: The lead visitor is not on the contact list
```
StupidCodeFactory: @kyliehavelock @kayesoutham what's the content for this validation error?

kyliehavelock: @StupidCodeFactory we should treat this error in the same way as any other rejection error. We don't need to have a specific error message - the visit will simply be rejected when the staff click 'Process'
Is that right do you think @chrishanes ?

kayesoutham: @StupidCodeFactory @kyliehavelock @chrishanes Think this should be covered by the generic rejection message we're going to put in: "This visit will be rejected and an email will be sent to the lead visitor"

StupidCodeFactory: @kyliehavelock so I'll reject the visit as "not_on_the_list" reason

kyliehavelock:
@alan yes that's correct, the validation should not be linked to the contact list API. The behaviour in this PR is correct: https://github.com/ministryofjustice/prison-visits-2/pull/877

cc @chrishanes

StupidCodeFactory:
3rd implementation:

This now rejects the visits if the lead visitor is either banned or not on the list. Meaning the visit is rejected if **any** of the two checkboxes is checked for the lead visitor .

Does this sound correct?

/cc @chrishanes @kyliehavelock @alan 


🙏  this is now correct

config/locales/en/errors.yml
chrishanes: @StupidCodeFactory Will this be a specific rejection reason in the rejection_decorator and a rejection email sent? We don't seem to have a lead visitor rejection email.

StupidCodeFactory: I actually need to remove this as it is not used

app/models/rejection.rb
```diff
@@ -37,6 +37,9 @@ class Rejection < ActiveRecord::Base
   validates :reasons, presence: true
   validate :validate_allowance_renews_on_date
 
+  before_save do
```
alan: It would be good to get this behaviour for unsaved objects, we could do it by overloading the `reasons` method to parse the db data as a set.

Something like this:

```ruby
def reasons
  Set.new(self[:reasons])
end
```

What do you think?

spec/factories/visitors.rb
```diff
@@ -10,6 +10,8 @@
       FFaker::Name.last_name
     end
 
+    sequence(:nomis_id)
```
alan: Is this needed here now?

app/models/rejection.rb
StupidCodeFactory: Yes I did think about that, I preferred cleaning up the ```reasons``` array before saving as I had very little idea how that would impact the aggregations calculations for the stats and for what it is worth didn't want to spend ages figuring out what are the implication when i could just simply user `Array#uniq!`, seemed more pragmatic.

StupidCodeFactory: What's your take on this  @doctorpod ?

alan: How do you see it impacting the aggregation calculations? Those are done in sql so it would be unchanged if I remember correctly.

I personally see more downsides by not keeping the object consistent during the whole lifecycle, however if it has a big knock on effect introducing this change lets keep it as it is.

StupidCodeFactory: Saving duplicates would impact stats as we aggregate per rejection reasons.
That would at least impact this CTE https://github.com/ministryofjustice/prison-visits-2/blob/master/db/views/rejection_percentage_by_days_v01.sql#L27

kyliehavelock:
@StupidCodeFactory we will need some test data to review this properly.

StupidCodeFactory:
@kyliehavelock I've created this visit if you want to playa around I've created this public test app 
 (https://pvb-public-pr-213.herokuapp.com/en/request) https://pvb-staff-pr-869.herokuapp.com/en/prison/visits/c83a91b2-3ce8-40b2-9f7b-a7f58d60eef8

app/models/rejection.rb
alan: Ah, I think I haven't explained myself well! Not saying that we should save duplicates, just keep the reasons uniq as it is modified

app/models/staff_response.rb
```diff
@@ -5,6 +5,8 @@ class StaffResponse
   attr_accessor :visit, :user
 
   before_validation :check_slot_available
+  before_validation :check_lead_visitor_not_banned
+  before_validation :check_lead_visitor_on_list
 
   validate :validate_visit_is_processable
   validate :visit_or_rejection_validity
```
alan: @StupidCodeFactory I just came across this method for my work and thought it might be related to this PR since it checks that there is at least one adult. Is it still relevant or do we need to check for valid lead visitor instead?

I'm not 100% sure right now, but I think it might need changing since at least one valid visitor is the old logic

StupidCodeFactory: I'd rather change this in a separate PR as part of the extra validation/scenarios needed to book to nomis

alan: Is not related to book to nomis as we split into what the card for this PR. Behaviour seems right, is just the logic on that method might be possible to simplify it since we don't check for at least one adult with this PR.

kyliehavelock:
@StupidCodeFactory looks good to me.

app/models/staff_response.rb
```diff
@@ -83,9 +84,15 @@ def visitors_attributes
...
+  def lead_visitor_attributes
+    attrs = visit.lead_visitor.attributes.slice('id', 'not_on_list', 'banned', 'type')
+    attrs['banned_until'] = visit.lead_visitor.banned_until.to_s
+    attrs
+  end
+
   def visit_or_rejection_validity
-    case [visit.slot_granted?, rejection.valid?, at_least_one_valid_visitor?]
-    when [true, true, true], [false, false, true]
+    case [visit.slot_granted?, rejection.valid?]
```
alan: 👍

config/locales/en/errors.yml
```diff
@@ -12,6 +12,7 @@ en:
             visit:
               already_processed: can't be processed
             base:
+              already_processed: "can't be processed"
```
alan: Looks like a duplicate on the live above

StupidCodeFactory:
Im closing this and will restart in another branch

PR #870, opened 2017-05-24
Ignore active flag when fetching contact list

doctorpod:


app/services/nomis/contact_list.rb
```diff
@@ -7,8 +7,9 @@ class ContactList
 
     attribute :contacts, Array[Contact]
     attribute :api_call_successful, Boolean, default: true
+
     def approved
-      select { |contact| contact.approved? && contact.active? }
```
StupidCodeFactory: you probably want to remove the `Nomis::Contact#active?` method

PR #878, opened 2017-05-31
Revert "Upgrade ruby 2.4.1"

alan:
Reverts ministryofjustice/prison-visits-2#873

The docker build is failing.


lukaszraczylo:
There's no need for revert - after pushing 2.4.1 to the public docker hub everything works just fine.

PR #879, opened 2017-05-31
Fix container build using Ruby 2.4.1

alan:


lukaszraczylo:
No need for this one either ( unless you really want it ), as the 2.4.1 is present in the docker hub now.

PR #889, opened 2017-06-05
Remove staff guide page links

chrishanes:
Remove the need for separate repo for staff content hosted on S3. Move downloads and telephone script pages to PVB2


app/controllers/concerns/prison_restriction_controller.rb
```diff
@@ -0,0 +1,8 @@
+module PrisonRestrictionController
```
alan: Should this be named without the `Controller` suffix since it is a concern?

kyliehavelock:
@chrishanes reviewed and approved.

PR #895, opened 2017-06-06
WIP - refactor to introduce lead visitor

StupidCodeFactory:


db/migrate/20170606112343_add_type_column_to_visitors.rb
```diff
@@ -0,0 +1,7 @@
+class AddTypeColumnToVisitors < ActiveRecord::Migration
+  def change
+    add_column :visitors, :type, :string
+    Visitor.where(sort_index: 0).update_all(type: 'LeadVisitor')
```
alan: This is dangerous for 2 reasons:
- ActiveRecord migrations use an `AccessExclusive` locks which is the most aggressive lock in PG since that is the lock needed to do DDL changes. This means that the table will be locked indefinitely (as there is no timeout set on the lock) for reads and writes until this transaction completes which will very likely cause some downtime if run during business hours. It is possible to set a timeout for the lock, so if it is exceeded the transaction is rollbacked to avoid locking out read and writes for long times. This will also require for the update to be speedy. The safer thing to do is to run the data update in a console / rake task which will use a normal transaction lock that doesn't lock the table.

- The second issue is that by loading the Visitor model in the migration is that the columns are cached so if there are any other migrations that run afterwards that modify the columns then the cached columns become out of date and errors happen that are really difficult to track down.

PR #896, opened 2017-06-06
Enforce contact list use

alan:
TODO - Finalise copy of error message

### DIFF
--- a/app/models/staff_response.rb
+++ b/app/models/staff_response.rb
@@ -3,11 +3,13 @@ class StaffResponse
   include ActiveModel::Validations::Callbacks
   ADULT_AGE = 18
   attr_accessor :visit, :user
+  attr_writer :validate_visitors_nomis_ready
 
   before_validation :check_slot_available
 
   validate :validate_visit_is_processable
   validate :visit_or_rejection_validity
+  validate :visitors_selection
 
   after_validation :check_for_banned_visitors
   after_validation :check_for_unlisted_visitors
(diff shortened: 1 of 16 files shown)


app/views/prison/visits/_visitor_contact.html.erb
```diff
@@ -20,9 +20,15 @@
         </br/>
         <%= @visit.contact_phone_no %></p>
       <% end %>
+
+      <% vf.object.errors.full_messages.each do |error_msg| %>
```
alan: @chrishanes this is the bit that needs styling

app/models/visitor.rb
```diff
@@ -2,9 +2,12 @@ class Visitor < ActiveRecord::Base
   include Person
   extend FreshnessCalculations
 
+  attr_accessor :validate_contact_list_matching
```
doctorpod: Shouldn't this be `attr_writer` because we have a reader method defined below?

alan: This will go. Based on a discussion with @StupidCodeFactory I'm going to try to get rid of this flag altogether from the model.

app/models/staff_response.rb
```diff
@@ -138,4 +151,16 @@ def clear_allowance_renews_on_date
...
   end
+
+  def visitors_selection
+    return unless validate_visitors_nomis_ready?
+
+    invalid_visitors = visit.visitors.select { |visitor|
+      visitor.nomis_id.present? == visitor.not_on_list?
+    }
+
+    invalid_visitors.each do |v| v.errors.add :base, :unprocessed_contact_list end
```
doctorpod: Is rubocop preventing you using curlies rather than `do...end`?

alan: Yep, this was an autocorrect. Strange that it allows do / end in a single line

kyliehavelock:
Reviewed and approved @alan - fine to deploy for those 10 prisons who are already using the contact list

PR #897, opened 2017-06-06
Move search label within search box

chrishanes:
To be inline with GOVUK style


kyliehavelock:
@chrishanes reviewed and approved

PR #906, opened 2017-06-12
Add chart from digital makeup spreadsheet

chrishanes:


config/routes.rb
```diff
@@ -25,6 +25,7 @@
       get 'metrics/:prison_id/summary',
         action: :summary,
         as: :prison_metrics_summary
+      get 'metrics/digital_takeup', action: :digital_takeup
```
StupidCodeFactory: Do we need to restrict access based on IP?

chrishanes: Not sure, a question for @kyliehavelock I think

kyliehavelock: The chart will be published on staff, so access will be restricted by IP anyway

PR #917, opened 2017-06-15
Add api call to book a visit in Nomis

alan:
Adds the api call with optional support of idempotency key

### DIFF
--- a/app/services/nomis/client.rb
+++ b/app/services/nomis/client.rb
@@ -9,6 +9,7 @@ module Nomis
   class Client
     TIMEOUT = 2 # seconds
     EXCON_INSTRUMENT_NAME = 'nomis_api'.freeze
+    JSON_MIME_TYPE = 'application/json'.freeze
 
     def initialize(host, client_token, client_key)
       @host = host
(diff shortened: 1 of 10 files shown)


app/services/nomis/client.rb
```diff
@@ -80,17 +81,27 @@ def request(method, route, params, idempotent:)
...
-        # else
-        #   {
-        #     body: params.to_json,
-        #     headers: { 'Content-Type' => 'application/json' }
-        #   }
+      else
+        {
+          body: params.to_json,
+          headers: { 'Content-Type' => JSON_MIME_TYPE }
+        }
```
doctorpod: Could you flip this and say: `if method == :post`?

app/services/nomis/client.rb
```diff
@@ -80,17 +81,27 @@ def request(method, route, params, idempotent:)
...
+        [200, 400, 409]
+      end
+    end
+
     # Returns excon options which put params in either the query string or body.
     def params_options(method, params)
       return {} if params.empty?
 
-      if %i[get delete].include?(method)
+      if method == :post
```
StupidCodeFactory: What about `put` and `patch` methods,  I would expect them to follow the post rules. I know we don't have put requests but that change seems unncessary

alan: When we start using those we can reconsider this code, no point worrying about it now

PR #922, opened 2017-06-15
Allow visits to be saved to Nomis in processor

alan:
Updates the BookingRequestProcessor to save visits to Nomis if configured to do
so.

The visit is created inside the database transaction to try to minimise the
chances of errors leaving Nomis and PVB inconsistent.

For example if the api is called was to be called outside the transaction then
we could end up with the visit persisted to nomis and the transaction failing,
or if it was called afterwards then pvb could end up with the visit booked but
not in nomis.

### DIFF
--- a/app/services/booking_response.rb
+++ b/app/services/booking_response.rb
@@ -1,9 +1,45 @@
 class BookingResponse
-  def initialize(success:)
-    @success = success
+  ALREADY_PROCESSED_ERROR = 'already_processed'.freeze
+  PROCESS_REQUIRED_ERROR = 'process_required'.freeze
+  NOMIS_VALIDATION_ERROR = 'nomis_validation_error'.freeze
+  NOMIS_API_ERROR = 'nomis_api_error'.freeze
+  SUCCESS = 'success'.freeze
+
+  attr_reader :message
+
+  def self.successful
+    new(SUCCESS)
+  end
+
+  def self.process_required
+    new(PROCESS_REQUIRED_ERROR)
+  end
+
+  def self.nomis_validation_error
+    new(NOMIS_VALIDATION_ERROR)
+  end
+
+  def self.nomis_api_error
+    new(NOMIS_API_ERROR)
+  end
+
+  def self.already_processed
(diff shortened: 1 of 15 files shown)


app/services/create_nomis_visit.rb
```diff
@@ -0,0 +1,54 @@
+class CreateNomisVisit
+  ALREADY_PROCESSED = 'Duplicate post'.freeze
+
+  def initialize(visit)
```
alan: @StupidCodeFactory I know you suggested passing the parameters instead of the visit. I think it makes sense to keep it with the visit as it keeps the logic of how to persist a PVB visit to the API here, and not spreading the logic to the `Accept` class.

What do you think @doctorpod ?

app/services/booking_response.rb
```diff
@@ -1,9 +1,45 @@
...
-    @success = success
+  ALREADY_PROCESSED_ERROR = 'already_processed'.freeze
+  PROCESS_REQUIRED_ERROR = 'process_required'.freeze
+  NOMIS_VALIDATION_ERROR = 'nomis_validation_error'.freeze
+  NOMIS_API_ERROR = 'nomis_api_error'.freeze
+  SUCCESS = 'success'.freeze
+
+  attr_reader :message
+
+  def self.succesful
```
doctorpod: Spelling: `successful`

app/services/create_nomis_visit.rb
doctorpod: Probably pass in the visit due to the number of params required otherwise

spec/services/booking_response_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+RSpec.describe BookingResponse do
+  describe '.succesful' do
+    subject { described_class.succesful }
+
+    it { is_expected.to be_success }
+    it { expect(subject.message).to eq(described_class::SUCCESS) }
+  end
+
+  describe '.already_processed' do
```
doctorpod: Is this description correct?

PR #924, opened 2017-06-15
Lead visitor validation

StupidCodeFactory:
**CHANGES:**
Validate if the lead visitor is on the list, not banned and more than 18

### DIFF
--- a/app/models/staff_response.rb
+++ b/app/models/staff_response.rb
@@ -117,15 +117,11 @@ privileged_allowance_expires_on])
   end
 
   def check_for_banned_visitors
-    if all_visitor_banned?
-      rejection.reasons << Rejection::BANNED
-    end
+    rejection.reasons << Rejection::BANNED if all_visitor_banned?
   end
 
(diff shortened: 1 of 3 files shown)


app/models/staff_response.rb
```diff
@@ -120,14 +117,12 @@ def check_slot_available
   end
 
   def check_for_banned_visitors
-    return if at_least_one_valid_visitor?
     if all_visitor_banned?
       rejection.reasons << Rejection::BANNED
     end
```
doctorpod: Probably room to do an inline `if`

app/models/staff_response.rb
```diff
@@ -120,14 +117,12 @@ def check_slot_available
...
   end
 
   def check_for_unlisted_visitors
-    return if at_least_one_valid_visitor?
     if all_visitor_not_on_list?
       rejection.reasons << Rejection::NOT_ON_THE_LIST
     end
```
doctorpod: As above

PR #929, opened 2017-06-16
Record processing time

StupidCodeFactory:
**CHANGES:**
Record time to process a visit. When a visit is processed it should send a timing event to the google analytics measurements API.


app/controllers/prison/visits_controller.rb
```diff
@@ -4,19 +4,22 @@ class Prison::VisitsController < ApplicationController
   before_action :authenticate_user
   before_action :cancellation_reason_set, only: :cancel
   before_action :visit_is_processable, only: %i[process_visit update]
+  after_action  :track_visit_process, only: :update
```
alan: We call the method inside `#update` when the booking response is a success so I think we can remove this after action.

StupidCodeFactory: @alan I've messed up this branch 💣  Will reopen if the latest version.

.rspec
```diff
@@ -1,2 +1,3 @@
 --color
+--backtrace
```
alan: ✂️ 
Previous discussion here https://github.com/ministryofjustice/prison-visits-2/pull/861#discussion_r119630544

I think another option is to added to `git exclude`, so that you can override the generic project `.rspec`, although I think your preference for any project is to have this enabled so using the global rspec setting might be the best for you.

spec/controllers/untrusted_examples.rb
```diff
@@ -1,9 +1,6 @@
 RSpec.shared_examples 'disallows untrusted ips' do
   context 'an untrusted ip' do
-    before do
-      allow_any_instance_of(ActionDispatch::Request).to receive(:remote_ip).
-        and_return('192.168.1.0')
-    end
+    before do request.headers['REMOTE_ADDR'] = '192.168.1.0' end
```
alan: This looks like a rubocop autocorrection, prefer to stick to the general convention of multi-line  for `do/end` blocks

alan: Can't see the link of this change for this PR, is this a general improvement?

app/services/ga_tracker.rb
```diff
@@ -0,0 +1,74 @@
...
+    }
+  end
+
+  def send_event
+    return unless value
+    client.post(
+      path: ENDPOINT.path,
+      headers: {
+        'Content-Type' => 'application/x-www-form-urlencoded'
+
```
alan: ✂️

spec/services/booking_responder_spec.rb
```diff
@@ -65,7 +65,7 @@
     end
   end
 
-  describe 'with a not requested visit' do
+  describe 'with a withdrawn visit' do
```
alan: 👍

PR #935, opened 2017-06-20
Refactor visit decorator to take a nomis checker

doctorpod:
The decorator now takes the staff nomis checker as an argument so that the same
nomis checker instance can be used by other objects if necessary.

### DIFF
--- a/app/controllers/concerns/staff_response_context.rb
+++ b/app/controllers/concerns/staff_response_context.rb
@@ -3,8 +3,17 @@ module StaffResponseContext
 
 private
 
-  def load_visit
-    @visit ||= scoped_visit
+  def memoised_visit
+    @_visit ||= scoped_visit
+  end
+
+  def decorate_visit(visit)
+    @_decorated_visit ||=
+      visit.decorate(context: { staff_nomis_checker: staff_nomis_checker })
+  end
+
+  def staff_nomis_checker
+    @staff_nomis_checker ||= StaffNomisChecker.new(memoised_visit)
   end
 
   def message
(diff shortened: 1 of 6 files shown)


app/controllers/concerns/staff_response_context.rb
```diff
@@ -7,6 +7,17 @@ def load_visit
     @visit ||= scoped_visit
   end
 
+  def decorate_visit(visit)
+    @decorated_visit ||=
+      VisitDecorator.decorate(
```
StupidCodeFactory: This should work:
```ruby
visit.decorate(context: { staff_nomis_checker: staff_nomis_checker })
```

app/controllers/concerns/staff_response_context.rb
```diff
@@ -7,6 +7,17 @@ def load_visit
...
+        visit,
+        context: { staff_nomis_checker: staff_nomis_checker })
+  end
+
+  def staff_nomis_checker
+    @staff_nomis_checker ||= StaffNomisChecker.new(load_visit)
```
StupidCodeFactory: Is there any reason why we load the visit twice?

alan:
@StupidCodeFactory can you take another look? We've tried to make it explicit that the visit is not loaded twice by renaming a `load_visit` to `memoised_visit`

app/controllers/concerns/staff_response_context.rb
```diff
@@ -3,8 +3,17 @@ module StaffResponseContext
...
 private
 
-  def load_visit
-    @visit ||= scoped_visit
+  def memoised_visit
+    @_visit ||= scoped_visit
+  end
+
+  def decorate_visit(visit)
+    @_decorated_visit ||=
```
StupidCodeFactory: looks like you can avoid passing around `memoised_visit` and just do as follow:

```ruby
def decorated_visit
  @_decorated_visit ||= memoised_visit.decorate(context: { staff_nomis_checker: staff_nomis_checker })
end

alan: No, because is not always the memoised visit. We use a visit with eager loaded associations in `prision/visits#show` action

PR #942, opened 2017-06-26
[STAGE-1] - Prevent prisons API to return ALI

StupidCodeFactory:
**CHANGES:**
* prevent `ALI` prison to be returned for no new visit booking requests are made to `ALI`
* rename Parkhust in to just `Isle of Wight` but keep sso_organisation_name the same. This will effectively have merged Isle of Wight prisons for visitors.

**NOTES:**
This is the first step of a 4 stages deploy process

### DIFF
--- a/app/controllers/api/prisons_controller.rb
+++ b/app/controllers/api/prisons_controller.rb
@@ -1,7 +1,10 @@
 module Api
   class PrisonsController < ApiController
     def index
-      @prisons = Prison.order(name: :asc).all
+      # temporarily filter out ALI so no new visit
+      # requests are sent to ALI
+      albany = Estate.find_by!(nomis_id: 'ALI')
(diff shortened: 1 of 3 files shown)


app/controllers/api/prisons_controller.rb
```diff
@@ -1,7 +1,10 @@
 module Api
   class PrisonsController < ApiController
     def index
-      @prisons = Prison.order(name: :asc).all
+      # temporarily filter out ALI so no new visit
+      # requests are sent to ALI
+      albany = Estate.find_by!(nomis_id: 'ALI')
+      @prisons = Prison.where.not(estate_id: albany.id).order(name: :asc).all
```
alan: It is possible to keep it to a single query:

```ruby
@prisons = Prison.joins(:estate).where.not(estates: { nomis_id: 'ALI' }).order(name: :asc).all
```

PR #943, opened 2017-06-26
Migrate Visits and FeedbackSubmission from ALI to IWI (step 2 and 4)

StupidCodeFactory:
**CHANGE:** 
This migrate all historical data from ALI prison to IWI prison (Visit and SubmissionFeedback)

**NOTES:**
This is the second step of a a 4 stage deployment


lib/tasks/one_off.rake
```diff
@@ -1,4 +1,20 @@
...
+      find_in_batches do |feebback_submissions|
+
+      FeedbackSubmission.where(id: feebback_submissions.map(&:id)).
+        update_all(prison_id: parkhurst.id)
+    end
+
+    Visit.where(prison_id: albany.id).find_in_batches do |visits|
+      Visit.where(id: visits.map(&:id)).update_all(prison_id: parkhurst.id)
+    end
+  end
```
alan: Be careful when you modify the set you are iterating over, what can happen is that only half the records get updated. I'm not sure if this will happen with this query.

Easy to run multiple times if necessary. The alternative is to avoid modifying the set you are looping by doing an update all with a limit and keep looking until there are no more records to update.

Something like:

```ruby
while FeedbackSubmission.where(prison_id: albany.id).any? do
  FeedbackSubmission.
    where(prison_id: albany.id).
    limit(1000).
    update_all(prison_id: parkhurst.id)
end
```

alan: You misunderstood me, I wasn't talking about validations.

It's about ending with a partial update after running the task. The code is looping through the records using a condition, and at the same time updating the database that relate to the batch query condition.

When doing this in the past I've seen cases where only half the set of records to be updated got updated because batches get skipped as the dataset gets mutated.

PR #948, opened 2017-06-27
Email slot availability

alan:
**CHANGES:**

Automated slot availability reporting
We receive an email everyday at 4am. 
Here is the corresponding PR adding a cronjob to automate that process 

@alan is that above correct?

### DIFF
--- a/app/views/admin_mailer/slot_availability.html.erb
+++ b/app/views/admin_mailer/slot_availability.html.erb
@@ -1,0 +1,10 @@
+<p>Prisons slot availability - <%= Date.today %></p>
+
+<% @availability.each do |prison_name, data| %>
+  <p>
+    <%= prison_name %><br>
+    <% data.each do |key, count| %>
+      <%= key %>: <%= count %><br>
+    <% end %>
+  </p>
+<% end %>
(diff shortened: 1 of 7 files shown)


lib/tasks/one_off.rake
```diff
@@ -230,13 +232,19 @@ namespace :pvb do
       visit.slots.all? { |s| s.to_date > Time.zone.today }
     }
 
-    non_expired.each do |visit| queue << visit end
+    SlotAvailabilityCounter.visits_checked = non_expired.size
+
+    non_expired.each do |visit|
+      queue << visit
+    end
```
doctorpod: In-line `{}` block?

app/views/admin_mailer/slot_availability.html.erb
```diff
@@ -0,0 +1,10 @@
+<p>Prisons slot availability - <%= Date.today %></p>
+
+<% @availability.each do |prison_name, data| %>
+  <p>
+    <%= prison_name %><br>
+    <% data.each do |key, count| %>
```
StupidCodeFactory: This would have been semantically more correct to use a `<dl/>`

lib/tasks/one_off.rake
alan: I think rubocop doesn't like and autocorrects to inline `do/end`

alan:
@StupidCodeFactory yep, updated description is correct.

I generally don't like to link issues or prs from open source repos to private repos. What's your view on it?

StupidCodeFactory:
yes i removed it 🔥

PR #952, opened 2017-06-28
Fix slot availability script

alan:
ActiveRecord `select` method seems to have changed behaviour which caused the
script to return wrong information.

The method was being used to select the estate nomis_id, however instead of
selected the nomis_id it now returns `0` which causes the API to return an
error.

### DIFF
--- a/lib/tasks/one_off.rake
+++ b/lib/tasks/one_off.rake
@@ -113,7 +113,7 @@ namespace :pvb do
 
     begin
       response = client.get(
-        "/prison/#{visit.nomis_id}/slots",
+        "/prison/#{visit.prison.nomis_id}/slots",
         start_date: current_slots.min.to_date - 1.day, # API bug workaround
         end_date: current_slots.max.to_date)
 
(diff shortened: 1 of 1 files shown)


lib/tasks/one_off.rake
```diff
@@ -223,8 +223,8 @@ namespace :pvb do
     queue = Queue.new
     task = ->(client, visit) { check_slot_availability(client, visit) }
 
-    visits = Visit.select('visits.*', 'estates.nomis_id nomis_id').
```
StupidCodeFactory: You were missing and AS nomis_id

alan: I believe the `AS` is optional, or it was until recently. I suspect a change of behaviour in ActiveRecord

alan: It is optional, and I'm more convinced of a regression in ActiveRecord:

```ruby
visit = Visit.select('visits.*', 'prisons.name prison_name', 'estates.nomis_id nomis_id').joins(prison: :estate).first

visit.prison_name # Aylesbury
visit.nomis_id # 0
```

PR #954, opened 2017-06-29
Better test coverage of contact list failure

doctorpod:
New scenario of feature test to cover when the call to the API fails

### DIFF
--- a/spec/features/process_a_request_accept_spec.rb
+++ b/spec/features/process_a_request_accept_spec.rb
@@ -216,4 +216,49 @@ RSpec.feature 'Processing a request - Acceptance', js: true do
       end
     end
   end
+
+  context 'contact list fails', vcr: {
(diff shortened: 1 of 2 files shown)


spec/features/process_a_request_accept_spec.rb
```diff
@@ -76,6 +76,29 @@
...
+
+    scenario 'accepting a booking but contact list fails', vcr: {
+      cassette_name: 'process_contact_list_fails',
+      allow_playback_repeats: true
+    } do
+      visit prison_visit_process_path(vst, locale: 'en')
+      click_button 'Process'
+
+      # Renders the form again
+      expect(page).to have_text('Visit details')
```
StupidCodeFactory: Always prefer have_css with a fairly specialised selector to improve speed and avoid capybara parsing all the text node of the page until it find the actual text

PR #966, opened 2017-07-06
Refactor acceptance feature specs

KPobeeNorris:
The feature specs for accepting a booking were becoming harder to maintain
due to the number of tests for different versions of the feature flags.

Now we have two files; one that has tests related to the contact list,
and the other where the contact list is disabled.


joseph-silverman:
Hey Kath 👍🏻😃😎💯🙋🏻‍♂️🖐🤗🎁🎈

Joseph Silverman

PR #967, opened 2017-07-06
Upgrade to Rails 5.1

doctorpod:
This is a work in progress...


app/decorators/rejection_decorator.rb
```diff
@@ -12,11 +12,9 @@ def allowance_renews_on
     @allowance_renews_on ||=
       begin
         if object.allowance_renews_on
-          AccessibleDate.new(
-            date_to_accessible_date(object.allowance_renews_on)
-          )
+          AccessibleDate.new(date_to_accessible_date(object.allowance_renews_on))
         else
-          AccessibleDate.new
+          AccessibleDate.from_multi_parameters(allowance_renews_on_before_type_cast)
```
alan: How come here we pass arguments while before we passed none?

app/models/cancellation.rb
```diff
@@ -17,5 +17,5 @@ class Cancellation < ActiveRecord::Base
 
   belongs_to :visit
 
-  validates :reason, inclusion: { in: REASONS }
+  validates :reason, inclusion: { in: REASONS }, allow_blank: true
```
alan: I don't think this is right, a cancellation should always have a `reason`

app/models/staff_response.rb
```diff
@@ -52,7 +52,7 @@ def validate_visitors_nomis_ready=(val)
     # TODO: Changes in Rails 5 to `ActiveRecord::Type::Boolean.new.cast(string)`
```
alan: :scissors: comment

app/services/book_to_nomis_config.rb
```diff
@@ -32,7 +32,7 @@ def already_booked_in_nomis?
 
   def opted_in=(val)
     # TODO: Changes in Rails 5 to `ActiveRecord::Type::Boolean.new.cast(string)`
```
alan: :scissors: comment

db/migrate/20170608082243_add_fields_to_feedback_submission.rb
```diff
@@ -1,4 +1,4 @@
-class AddFieldsToFeedbackSubmission < ActiveRecord::Migration
+class AddFieldsToFeedbackSubmission < ActiveRecord::Migration[4.2]
```
alan: We could think about consolidating the migrations so we don't have to keep maintaining very old migration files. To start a new db we would use the schema.

Gemfile
```diff
@@ -77,4 +75,6 @@ group :test do
   gem 'vcr'
   gem 'webmock'
   gem 'shoulda-matchers'
+  gem 'database_cleaner'
```
alan: Do we need `database_cleaner` after all?

We are not using it in `rails_helper` and we have the rspec setting `use_transactional_fixtures = true`, rspec docs say to have that setting to true when not using database cleaner. https://relishapp.com/rspec/rspec-rails/docs/transactions

app/decorators/rejection_decorator.rb
StupidCodeFactory: to retain the values submitted in the form

Gemfile
StupidCodeFactory: That's me forgetting to re-add the rails_helper bits :/. But we defo use not transactional test when using the feature tests. Although there are "SystemTest" that removes the need for database cleaner it does not seem available to RSpec unfortunately at the moment.

app/models/cancellation.rb
StupidCodeFactory: that's a good point, @doctorpod did that bit. No idea what's going on there

app/decorators/rejection_decorator.rb
alan: This is then broken also in master right?

StupidCodeFactory: nope works fine in master.

StupidCodeFactory: it's because before `Rejection#allowance_renews_on=` would retain the assigned value wether it was valid or not but with the newer active record the attribute/instance variable gets wiped out by the newer `ActiveRecord`

db/migrate/20170608082243_add_fields_to_feedback_submission.rb
StupidCodeFactory: Yes I thought about that, we should add something in the tech backlog

PR #975, opened 2017-07-13
Add unsupported content for users of IE8 or less

chrishanes:


app/views/layouts/application.html.erb
```diff
@@ -12,7 +12,9 @@
     <%= stylesheet_link_tag('application-ie8', media: 'all') %>
   <![endif]-->
 <% end %>
-<%= content_tag :div, nil, class: "js-AsyncGA", data: ga_tracking_data %>
+<!--[if gte IE 9]><!-->
```
StupidCodeFactory: I'm not sure what these are doing?

chrishanes: We're detecting for IE9 or above (plus other non-IE browsers) to render the content tag as this is what breaks IE8.

PR #979, opened 2017-07-18
Refactor test config env

KPobeeNorris:
Prevent the feature flags in the test environment to be configured
by the local .env settings.  This should mean that the tests should
pass regardless of where they run.

### DIFF
--- a/spec/services/staff_nomis_checker_spec.rb
+++ b/spec/services/staff_nomis_checker_spec.rb
@@ -33,7 +33,7 @@ RSpec.describe StaffNomisChecker do
 
   describe 'When the API is enabled' do
     describe '#prisoner_existance_status' do
-      describe 'api is configured and the check is disabled for staff' do
+      describe 'api is enabled and the check is disabled for staff' do
         before do
           switch_off(:nomis_staff_prisoner_check_enabled)
         end
@@ -43,6 +43,7 @@ RSpec.describe StaffNomisChecker do
 
       describe 'when the nomis api is live' do
         before do
+          switch_on :nomis_staff_prisoner_check_enabled
           mock_nomis_with(:lookup_active_offender, offender)
         end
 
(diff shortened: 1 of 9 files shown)


spec/services/staff_nomis_checker_spec.rb
```diff
@@ -224,21 +237,14 @@
         let(:offender) { Nomis::NullOffender.new }
 
         before do
+          switch_on(:nomis_staff_prisoner_availability_enabled)
           mock_nomis_with(:lookup_active_offender, offender)
         end
 
         context 'with NOMIS_STAFF_SLOT_AVAILABILITY_ENABLED switched OFF' do
-          before do
```
alan: Given the context I would not delete this before block

spec/services/staff_nomis_checker_spec.rb
```diff
@@ -224,21 +237,14 @@
...
-            switch_off :nomis_staff_slot_availability_enabled
-          end
-
           it { expect(subject.errors_for(slot)).to be_empty }
 
           describe "and STAFF_PRISONS_WITH_SLOT_AVAILABILITY switched OFF for the visit's prison" do
-            before do
```
alan: Same here

spec/services/staff_nomis_checker_spec.rb
```diff
@@ -99,6 +107,10 @@
         describe "and the API is unavailable" do
           let(:offender) { Nomis::NullOffender.new(api_call_successful: false) }
 
+          before do
```
alan: What do you think about having a context for when the api is enabled? That way we can avoid having the repetition of having to enable the feature flag

spec/services/staff_nomis_checker_spec.rb
```diff
@@ -33,7 +33,7 @@
 
   describe 'When the API is enabled' do
     describe '#prisoner_existance_status' do
-      describe 'api is configured and the check is disabled for staff' do
+      describe 'api is switch_on :nred and the check is disabled for staff' do
```
alan: typo here

PR #988, opened 2017-07-21
Closed visits restrictions

alan:
Shows when a requested slot has a closed restriction. When book to nomis is enabled, it forces the user to book the visit manually.


spec/features/process_a_request_accept_with_contact_list_spec.rb
```diff
@@ -94,6 +96,27 @@
...
+        select 'IRMA ITSU - 03/04/1975', from: 'Match to contact list'
+      end
+
+      expect(page).to have_css('.panel', text: "This is a closed visit. Book this visit into NOMIS, then enter the reference number")
+
+      click_button 'Process'
+      expect(page).to have_css('.notification', text: 'Thank you for processing the visit')
+
+      vst.reload
+      expect(vst).to be_booked
```
StupidCodeFactory: I know this is done in other places, but an integration test should not access a database record directly, `vst.reload` strongly hints a code smell. I'd rather have the test navigate to the processed visits page and checked that the label for this visit is `booked`

We could implement this as high level matcher and use it everywhere else.
This is just a comment to (re)open this discussion and does not need to be implemented in this PR.

spec/javascripts/moj.closedRestriction_spec.js
```diff
@@ -0,0 +1,78 @@
...
+    loadFixtures('process_form.html');
+    moj.Modules.MatchVisitors.init();
+    moj.Modules.Rejection.init();
+    moj.Modules.closedRestriction.init();
+    $('.issue-with-prisoner').hide();
+    $('.prisoner-details').hide();
+    $('.other-details').hide();
+    $('#visitors-fixture').hide();
+
+    // $('#visit_slot_granted_2017-05-13t09151115').click().trigger('change');
```
StupidCodeFactory: that could be cleaned up

spec/services/book_to_nomis_config_spec.rb
```diff
@@ -80,6 +80,14 @@
     end
   end
 
+  shared_context 'offender restrictions working' do
+    before do
+      allow(Nomis::Feature).
+        to receive(:offender_restrictions_enabled?).and_return(true)
```
StupidCodeFactory: don't we have a helper for this `#switch_on(feature_flag)`?

spec/services/nomis/api_spec.rb
```diff
@@ -169,6 +169,38 @@
...
+          type: { code: 'BAN', desc: 'Banned' },
+          effective_date: Date.parse('2017-03-09'),
+          expiry_date: Date.parse('2017-03-13')
+        )
+      end
+
+      let(:first_restriction) { subject.first }
+
+      it 'parses the response' do
+        expect(first_restriction.type).to eq(expected_restriction.type)
```
StupidCodeFactory: you could use `expect(first_restriction).to have_attributes({...})` only have one expect and also a much nicer error message and more checks on the object tested

spec/services/nomis/restriction_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+    described_class.new(type: type,
+                        effective_date: effective_date,
+                        expiry_date: expiry_date)
+  end
+
+  describe '#banned?' do
+    context 'with a banned restriction type' do
+      let(:type) { { code: "BAN", desc: "Banned" } }
+
+      it { expect(subject).to be_banned }
```
StupidCodeFactory: `it { is_expected.to be_banned }`
Same thing for all the bellow tests

spec/services/nomis/restriction_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+      it { expect(subject.effective?(date)).to eq(false) }
+    end
+
+    context 'when the date is after the effective date' do
+      let(:effective_date) { date - 1.week }
+
+      context 'and before the expiry date' do
+        let(:expiry_date) { date + 1.day }
+
+        it { expect(subject.effective?(date)).to eq(true) }
```
StupidCodeFactory:
```ruby
it { is_expected.to be_effective(date) }
```
or with this the suggested renamed method:
```ruby
it { is_expected.to be_effective_at(date) }
```

app/services/nomis/restriction.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    def banned?
+      type[:code] == BANNED_CODE
+    end
+
+    def closed?
+      type[:code] == CLOSED_CODE
+    end
+
+    def effective?(date)
```
StupidCodeFactory: what about  `effective_at?`, it reads better

spec/services/staff_nomis_checker_spec.rb
```diff
@@ -175,7 +179,18 @@
     describe '#errors_for' do
       let(:slot) { visit.slots.first }
 
+      # TODO: Cleanup after global test config for feature flags has
+      # been removed from cong/test.rb
```
StupidCodeFactory: I would not not know what to here. Can you be a bit more explicit?
also I believe you meant to write `config/environments/test.rb` ?

app/assets/javascripts/modules/moj.BookToNomis.js
```diff
@@ -24,10 +22,10 @@
     render: function () {
       var $el = $('#'+this.$conditionalEl);
```
chrishanes: I hadn't noticed the $ (I probably copied and pasted from another file), I will update the variable name

spec/features/process_a_request_accept_with_contact_list_spec.rb
alan: 👍 on relying less on database access to do the checks. I think this is a leftover from when the app didn't have a staff UI other than the processing page.

There are some things that we check that we can only know from the database such as the noms visit id.

spec/services/book_to_nomis_config_spec.rb
alan: Yes, but that method controls the environment variables configuration which is lower level.

`Nomis::Feature` decides if all dependencies for a given feature are met, which I think it makes it nicer to stub in this case as this test doesn't need to care about all the dependencies that make a feature enabled.

For example for offender restrictions the offender lookup also needs to be enabled.

app/services/prisoner_restriction_list.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def unknown_result?
+    !offender_restrictions.api_call_successful?
+  end
+
+  def on_slot(slot)
+    offender_restrictions.
+      select(&:closed?).
+      select { |restriction| restriction.effective?(slot.to_date) }.
+      map(&:name)
```
alan: Not sure what you mean, is doing a single `map`?

PR #994, opened 2017-07-31
Visitor ban rejection email issue

KPobeeNorris:
Issue where if the lead visitor is banned then the rejection email sent
to them included the incorrect copy i.e. the copy where someone is not
on the list.  Implement amendment so email only includes details about
the visitor ban.

### DIFF
--- a/app/models/staff_response.rb
+++ b/app/models/staff_response.rb
@@ -103,8 +103,10 @@ privileged_allowance_expires_on])
   end
 
   def check_principal_visitor
-    if principal_visitor.banned? || principal_visitor.not_on_list? ||
-           principal_visitor.age < ADULT_AGE
+    if principal_visitor.banned?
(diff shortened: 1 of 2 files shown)


app/models/staff_response.rb
```diff
@@ -103,8 +103,9 @@ def rejection
   end
 
   def check_principal_visitor
-    if principal_visitor.banned? || principal_visitor.not_on_list? ||
-           principal_visitor.age < ADULT_AGE
+    if principal_visitor.banned?
+      rejection.reasons << Rejection::BANNED
+    elsif principal_visitor.not_on_list?
       rejection.reasons << Rejection::NOT_ON_THE_LIST
       visit.slot_granted = nil
```
alan: We always need to nil the `slot_granted` so that the visit is rejected

PR #1064, opened 2017-09-12
only push per branch

StupidCodeFactory:


alan:
Closing since we stopped using Wercker for now

PR #1077, opened 2017-09-21
Rejection reason overrides

KPobeeNorris:
**CHANGES:**

General refactor or the rejection reasons checkboxes
Show a warning message when overwriting a rejection from nomis


app/decorators/rejection/reason_decorator.rb
```diff
@@ -0,0 +1,17 @@
+class Rejection::ReasonDecorator < Draper::Decorator
+  delegate_all
+  def checkbox_for(reason, html_options = {})
+    html_options[:id] = nil unless html_options.key?(:id)
```
alan: There is quite a lot going in this method, how about splitting some of the logic into private methods.

One could be to setup the default values on `html_options` (making sure it has keys for `id`, `data` and `class` and then another method that returns the `class` value ie html_class

app/views/prison/visits/_rejection_reason_checkbox.erb
```diff
@@ -0,0 +1,4 @@
+<%= rf.object.checkbox_for(id, id: id, data: {
```
alan: I'd be tempted to rename this partial to `_rejection_reason` since it is not only a checkbox. Also, should the extension be `html.erb` ?

PR #1082, opened 2017-09-21
Rearrange cancellation reasons

chrishanes:
To keep inline with processing page and group the cancellation reasons together


app/views/prison/visits/_cancel_visit.html.erb
```diff
@@ -0,0 +1,67 @@
+<h3 class="heading-medium">Issue with visit booking</h3>
```
alan: Move to locale file

chrishanes:
@alan My local integration test didn't show any errors

PR #1090, opened 2017-09-28
Update description regarding hack in sign-on client

StupidCodeFactory:
**CHANGES:**

Explain better why this hack is needed

### DIFF
--- a/lib/mojsso.rb
+++ b/lib/mojsso.rb
@@ -27,24 +27,6 @@ module OmniAuth
       def raw_info
         @raw_info ||= access_token.get('/api/user_details').parsed
       end
-
-      # Without this login with sso breaks.
-      #
-      # Fix as suggested here: https://github.com/intridea/omniauth-oauth2/commit/26152673224aca5c3e918bcc83075dbb0659717f#commitcomment-19809835
-      # Other link about the issue: https://github.com/intridea/omniauth-oauth2/issues/81
-      #
-      # omniauth-oauth2 after version 1.3.1 it changed the way that the callback
-      # url gets generated. This new version doesn't match the redirect uri as set in
-      # SSO so login breaks (I think it doesn't strip the previous query string,
-      # although I would need to double check to be sure)
-      #
-      # The issue seems quite common among multiple SSO providers like Google,
-      # Facebook, Dropbox, etc
-      #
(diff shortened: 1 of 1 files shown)


alan:
@StupidCodeFactory I can't login in the Heroku deploy, can you give it a go?

StupidCodeFactory:
Getting this failure:

Authentication failure! invalid_credentials: OAuth2::Error, invalid_grant: The provided authorization grant is invalid, expired, revoked, does not match the redirection URI used in the authorization request, or was issued to another client.

alan:
Should we close then?

StupidCodeFactory:
@alan so the previous comment is actually completely wrong about what problem is. Maybe at the time it was but it is wrong now.
I'll try to re-rephrase the comments before

StupidCodeFactory:
So by stripping the `state` params from the url when doorkeeper redirects the user to this URL,  the client app can not perform the check against the `state` params which is what is used to determine if a request is legit. 

There are two things to consider here:
* At first `omniauth-oauth2` was overriding the callback_url stripping off the very useful `state` params used to verify a request against CSRF attacks. Then to fix this (see https://github.com/intridea/omniauth-oauth2/issues/28) overriding the `callback_url` was removed allowing query params to be passed in the URL (probably generated in the `omniauth` gem).
* On the other side in the signon app Doorkeeper has a bug so that when it matches redirect URIs it does not stripping off the extra dynamically generated query params, which results in a 401 `invalid_grant`. 

The result of these two thing is that we need to keep the faulty/unsafe `callback_url` in the client for pvb to keep working with signon.

We've had complaints about 401 happening randomly in the signon app (not so random after all) and updating/patching doorkeep to make it work properly, right now it's faulty and causes errors for other services.

I can make a demo on Monday to of the problem.

alan:
OK, let me see if I understand. There are 2 issues:
- one is when there is no SSO session login breaks due to the redirect_uri which is what the old comment says
- the other one is when there is a SSO session when the client tries to set up a session a CSRF error can happen

StupidCodeFactory:
Yeah I think that is more or less accurate. 

1. We need to fix signon app with the PR I've pointed out
2. Remove previous (i.e overriding the callback_url) as it bypasses the CSRF check. As of now it is possible to hijack the session.

PR #1095, opened 2017-10-03
Multiple cancellation reasons

alan:
Starts storing the cancellation reason as an array of reasons in the database and adds a task to migrate old cancellations to the new format.

Next steps are to update emails and public to make use of multiple cancellation reasons and then update the form so that users can specify multiple cancellations reasons.

### DIFF
--- a/db/migrate/20170929084924_add_reasons_to_cancellations.rb
+++ b/db/migrate/20170929084924_add_reasons_to_cancellations.rb
@@ -1,0 +1,5 @@
+class AddReasonsToCancellations < ActiveRecord::Migration[5.1]
+  def change
+    add_column :cancellations, :reasons, :string, array: true, default: []
+  end
+end
--- a/spec/models/cancellation_spec.rb
+++ b/spec/models/cancellation_spec.rb
@@ -1,6 +1,8 @@
 require 'rails_helper'
 
 RSpec.describe Cancellation, model: true do
+  subject { FactoryGirl.build(:cancellation) }
+
   describe 'validation' do
     it 'enforces no more than one per visit' do
       cancellation = FactoryGirl.create(:cancellation)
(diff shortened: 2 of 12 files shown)


spec/models/cancellation_spec.rb
```diff
@@ -1,6 +1,8 @@
 require 'rails_helper'
 
 RSpec.describe Cancellation, model: true do
+  subject { FactoryGirl.build(:cancellation) }
```
StupidCodeFactory: `FactoryGirl` isn't needed

StupidCodeFactory:
👍  LGTM

Will we release this on its own or should we wait for the other part to be done?

db/migrate/20170929084924_add_reasons_to_cancellations.rb
```diff
@@ -0,0 +1,5 @@
+class AddReasonsToCancellations < ActiveRecord::Migration[5.1]
+  def change
```
StupidCodeFactory: @alan Are we sure this is going to complete within 750ms?

StupidCodeFactory: Should not matter much actually if this fails, but I'd increase the default timeout to make sure we don't hit the timeout as we've experienced before there is no guaranty the transaction will be rolled back properly. So I'd increase the timeout to be on the safe side.

alan: It's different, is a timeout to acquire the lock and a timeout per statement inside the transaction.

Rails migrations transactions use an access exclusive lock, so it will lock subsequent reads and writes to that table. If for whatever reason the lock waits on another transaction to finish before starting this timeout will kick in and ensure that reads and writes that come from http requests aren't locked for a long time and cause a downtime.

The transaction contents won't even run, and even if it does it would be safe as it does not disable DDL transactions. These transactions typically run in 1 or 2 ms so it should be fine and we typically don't have long running transactions that could prevent the lock to be acquired in the first place.

spec/models/cancellation_spec.rb
alan: Yep, I know, it's there for consistency.

alan:
Yes, this can be released independently, I'll merge, deploy and run the task.

db/migrate/20170929084924_add_reasons_to_cancellations.rb
StupidCodeFactory: Ok sounds perfectly fine then I was worried the timeout was for the overall DDL transaction

PR #1101, opened 2017-10-03
Cancellation reasons api and email

alan:
Makes cancellation emails compatible with multiple reasons and adds the reason to the api.

It also refactors the single cancellation view to always show the intro.

### DIFF
--- a/app/views/visitor_mailer/_multiple_cancellation_reasons.html.erb
+++ b/app/views/visitor_mailer/_multiple_cancellation_reasons.html.erb
@@ -1,0 +1,12 @@
+<p>
+  <%= t('intro',
+       date: format_slot_for_public(@visit.slot_granted),
+       prison: @visit.prison.name,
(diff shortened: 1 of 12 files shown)


app/views/visitor_mailer/_multiple_cancellation_reasons.html.erb
```diff
@@ -0,0 +1,12 @@
...
+  <%= t('intro',
+       date: format_slot_for_public(@visit.slot_granted),
+       prison: @visit.prison.name,
+       prisoner: @visit.prisoner_full_name,
+       scope: %i(visitor_mailer cancelled))
+  %>
+</p>
+<p><%= t('cant_visit', scope: %i(visitor_mailer cancelled)) %>:</p>
+<ul class="list list-bullet">
+  <%= render @cancellation.formatted_reasons %>
```
StupidCodeFactory: 👍

PR #1108, opened 2017-10-05
Allow staff to check multiple cancellation reasons

alan:
Changes radio input to checkboxes to let staff select multiple
cancellation reasons.

It also adds a few missing translation for cancellation reasons.

After this PR is deployed the 'reason' column can be dropped.

### DIFF
--- a/app/views/visitor_mailer/_multiple_cancellation_reasons.html.erb
+++ b/app/views/visitor_mailer/_multiple_cancellation_reasons.html.erb
@@ -10,3 +10,4 @@
 <ul class="list list-bullet">
   <%= render @cancellation.formatted_reasons %>
 </ul>
+<p><%= t('arrange_html', scope: %i(visitor_mailer cancelled)) %></p>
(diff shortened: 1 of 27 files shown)


app/views/visitor_mailer/_multiple_cancellation_reasons.html.erb
```diff
@@ -10,3 +10,4 @@
 <ul class="list list-bullet">
   <%= render @cancellation.formatted_reasons %>
 </ul>
+<p><%= t('arrange_html', scope: %i(visitor_mailer cancelled)) %></p>
```
StupidCodeFactory: might be wrong but is this not missing a `.` in front of `arrange_html`?

chrishanes: I don't think you use a . when adding a scope? Not sure, was just copying what was already there.

StupidCodeFactory: ok, it's probably fine then. What could possibly go wrong? :trollface:

PR #1124, opened 2017-10-09
Track bookable visits rejections

alan:
When a visit is bookable according to the checks done using the different Nomis API's and the visit is rejected an Google Analytics event is sent.


app/services/ga_tracker.rb
```diff
@@ -16,13 +16,13 @@ def set_visit_processing_time_cookie
...
 
-  def send_event
-    return unless value
-    client.post(
-      path:    ENDPOINT.path,
-      headers: { 'Content-Type' => 'application/x-www-form-urlencoded' },
-      body:    URI.encode_www_form(payload_data)
-    )
+  def send_unexpected_rejection_event
+    send_data(rejection_event_payload) if visit_rejected_unexpectedly?
```
StupidCodeFactory: I'd rather have the caller of this method do the check and not call the method at all

alan: We were in 2 minds about where the logic for this should be.

The controller didn't seem like a great place to have it - testing that there would have been a pain which is a sign that shouldn't be there, and didn't seem that the logic for `visit_rejected_unexpectedly?` belonged there.

The previous method to send timing data already made a check before sending the event so it made sense to keep the logic for this new event following the same pattern.

If you have other alternatives happy to look into it

alan: Yes, the other class passed my mind but it seemed a bit too much for what it was going to do. Let's leave it like this then.

PR #1142, opened 2017-10-12
Cancel a visit booked to NOMIS 

StupidCodeFactory:
**CHANGES:**
Cancel visit to NOMIS when book to NOMIS is enabled for a prison


app/views/prison/cancellations/new.erb
```diff
@@ -0,0 +1 @@
+<%= render 'prison/visits/processed' %>
```
alan: Is this used? Looks related to the rendering of cancellation errors, however there is no action or route to `cancellations#new`

app/controllers/prison/cancellations_controller.rb
```diff
@@ -0,0 +1,41 @@
+class Prison::CancellationsController < ApplicationController
+  include StaffResponseContext
```
alan: Seems a bit overkill to include the `StaffResponseContext` since it is very specific to logic to `requested` visits. You are only using the `memoised_visit` method right?

app/views/prison/visits/_cancel_visit.html.erb
```diff
@@ -1,24 +1,22 @@
-<%= form_tag cancel_prison_visit_path(@visit, locale: 'en'), class: 'form' do -%>
+<%= form_for [:prison, @visit, (@visit.cancellation || @visit.build_cancellation).decorate], class: 'form' do |f| -%>
```
alan: Can the logic for the cancellation object go somewhere else? Perhaps the visit decorator?
 Doesn't look like the right place to have this logic...

app/views/prison/visits/_cancel_visit.html.erb
```diff
@@ -1,24 +1,22 @@
...
 
   <div class="grid-row">
     <fieldset>
-      <div class="column-one-half">
-        <% %i[prisoner_vos prisoner_released child_protection_issues slot_unavailable visitor_banned].each do |reason| %>
-          <div class="multiple-choice">
-            <%= check_box_tag 'cancellation_reasons[]', reason, false, id: reason %>
-            <%= label_tag reason, t(".#{reason}") %>
+      <%= error_container(f, :reasons, class: 'form-group form-group-error') do %>
+        <% f.object.staff_cancellation_reasons.in_groups_of(5, false).each do |reasons| %>
```
alan: 👍

app/views/prison/visits/_cancel_visit.html.erb
```diff
@@ -1,24 +1,22 @@
...
-      <div class="column-one-half">
-        <% %i[prisoner_moved prisoner_non_association prisoner_cancelled booked_in_error capacity_issues].each do |reason| %>
-          <div class="multiple-choice">
-            <%= check_box_tag 'cancellation_reasons[]', reason, false, id: reason %>
-            <%= label_tag reason, t(".#{reason}") %>
-          </div>
-        <% end %>
-      </div>
+      <% end %>
+
```
alan: We'll need to tell the user if the visit is going to be cancelled in Nomis when they cancel it on PVB. Otherwise they wont know when to change their behaviour when the feature is enabled and get unexpected behaviour.

app/controllers/prison/cancellations_controller.rb
StupidCodeFactory: I'm using the `#memoized_visit` (relies on the `#scoped_visit`) and the `#decorate_visit` method. 
I'm quite happy extract some of those methods into their own classes or module, since they are closely related?

app/controllers/prison/cancellations_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+  def create
+    if cancellation_response.valid?
+      cancellation_response.cancel!
+      flash[:notice] = t('visit_cancelled', scope: %i[prison flash])
+      redirect_to prison_visit_path(memoised_visit)
+    else
+      flash.now[:alert] = cancellation_response.error_message
+      @visit = decorate_visit(memoised_visit)
+      @message = Message.new
+      render :new
```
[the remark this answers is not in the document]
StupidCodeFactory: this renders the `app/views/prison/cancellations/new.erb`, no new to have a route when the view exists.

app/services/booking_responder/cancel.rb
```diff
@@ -3,12 +3,24 @@ class Cancel < BookingRequestProcessor
...
       super do
         visit.cancel!
-        Cancellation.create!(visit: visit,
-                             reasons: staff_response.reasons,
-                             nomis_cancelled: true)
-
-        BookingResponse.successful
+        visit.cancellation.save!
+        if cancel_nomis_visit?
+          nomis_visit_cancellation.execute(comment: message)
```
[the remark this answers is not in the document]
StupidCodeFactory: So there is the ability to pass via NOMIS API a message/description to clarify why this visit was cancelled. But that would need some design work, iteration of the feature, so I figured I'd provide a way to pass the message a long. 
It's something we can discuss with @kyliehavelock when she's back. But I agree that if we are never going implement this then It'd be better to get rid of that.

app/services/nomis/client.rb
```diff
@@ -87,15 +91,15 @@ def http_method_expects(method)
       else
         # TODO: Change 400 to 422 when api is updated.
         # Currently it returns 400 for validation errors.
-        [200, 400, 409]
+        [200, 400, 404, 409]
```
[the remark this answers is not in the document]
StupidCodeFactory: That's the error code we get when the visit has already been cancelled? Although the API documentation specifies it should return a 400. I'll file a bug in the NOMIS API board.

Have a look at this cassette:
spec/fixtures/vcr_cassettes/cancelled_visit.yml

app/views/prison/cancellations/new.erb
StupidCodeFactory: that is indeed used, it is what is rendered when validation errors occurs, pretty standard rails pattern:
if `#create` actions has validation errors render the `#new` with validation errors displayed.

app/views/prison/visits/_cancel_visit.html.erb
StupidCodeFactory: good point. I'll still what we have for the book/reject to nomis

config/initializers/customise_error.rb.rb
```diff
@@ -0,0 +1,3 @@
+ActionView::Base.field_error_proc = proc do |html_tag, _instance_tag|
```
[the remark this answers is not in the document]
StupidCodeFactory: yes, classic trick to prevent rails from wrapping input fields with some extra error divs

spec/services/nomis/api_spec.rb
```diff
@@ -312,4 +312,63 @@
...
+        }.from(nil).to(false)
+      end
+    end
+
+    context 'with an invalid cancellation code' do
+      let(:visit_id) { 'POOBAR' }
+
+      it 'records the error message', vcr: { cassette_name: :cancel_invalid_cancellation_code } do
+        expect(subject.cancel_visit(offender_id, visit_id, params: {})).
+          to have_attributes(error_message: 'Invalid or missing visit_id')
```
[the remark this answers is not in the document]
StupidCodeFactory: this returns a 400, that's an expected code in the client, here is the corresponding VCR https://github.com/ministryofjustice/prison-visits-2/pull/1142/files#diff-67f9af546e2b4ea1244cb00863b15b2cR20

http://ministryofjustice.github.io/nomis-api/#visits-cancel-a-visit-patch

app/services/nomis/cancellation.rb
```diff
@@ -0,0 +1,11 @@
+module Nomis
+  class Cancellation
+    include NonPersistedModel
+    attribute :message, String
```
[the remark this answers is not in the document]
StupidCodeFactory: created a card for it

app/controllers/prison/cancellations_controller.rb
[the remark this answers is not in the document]
alan: 🤦‍♂️ I knew I must have been missing something obvious. 😪 brain

app/services/nomis/client.rb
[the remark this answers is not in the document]
alan: I have looked at it, I don't think the test setup is/was correct. The error message is:

`'{"error":{"message":"Visit not found"}}'`

config/initializers/customise_error.rb.rb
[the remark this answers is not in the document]
alan: I'd prefer if we could avoid having to change a global Rails setting for this feature. I've come across difficult to debug bugs because of the use of this method.

Given that we haven't had to rely on this before can't we do the same for this feature?

spec/services/nomis/api_spec.rb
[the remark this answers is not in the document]
alan: The cassette is returning a `404` visit not found so it needs to be fixed and the `404` removed from the expected codes.

app/controllers/prison/cancellations_controller.rb
alan: 👍

spec/services/nomis/api_spec.rb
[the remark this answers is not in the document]
StupidCodeFactory: Are you looking at the right cassette? Or am I missing something?

screen shot 2017-10-31 at 11 05 42

app/services/nomis/client.rb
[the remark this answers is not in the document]
StupidCodeFactory: I think you're confused between an *invalid cancellation code* and a *visit already cancelled*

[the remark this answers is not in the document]
StupidCodeFactory: so until the NOMIS API does not return a 400 for a visit already cancelled we can't release this feature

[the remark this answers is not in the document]
alan: Yes, correct. We can't expect a 404 code when the visit is already cancelled.

spec/services/nomis/api_spec.rb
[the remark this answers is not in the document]
alan: Probably wrong comment before, I find the setup confusing.

Re-reading the test context I do no expect to see an error message for the visit id, it should be an error message for the invalid cancellation reason.

Is the assumption in this test that `POOBAR` is a valid visit id? 

I think before I understood the test as testing the case of sending an invalid visit id given the ID value and the error message. 

If this is testing the case of an invalid visit id then the API should return a 404 and the code raise an API error. If it is testing an invalid rejection reason then the error message on the cassette does match the error.

spec/fixtures/vcr_cassettes/visit_not_found.yml
```diff
@@ -0,0 +1,42 @@
+---
```
alan: What do you think about renaming this cassette to `cancel_visit_not_found`?

app/services/nomis/client.rb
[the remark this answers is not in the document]
alan: Looked at the cassette, and is not very clear the setup.

Is the assumption that the same api endpoint was called before? 

Since there is only one call and the API returns a 404 with a specific error message for the visit id it makes it look like the visit id is invalid.

For me the intent of the cassette would be clearer if the cassette had 2 requests, the first showing the visit getting cancelled and then the second request failing. What do you think?

config/initializers/customise_error.rb.rb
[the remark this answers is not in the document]
StupidCodeFactory: The problem would arise elsewhere, there aren't any other places where we match validation errors to a specific field.
Ultimately we could use https://github.com/ministryofjustice/govuk_elements_form_builder to ensure that GDS styling is enforced. But without this fix displaying a validation error on the matching input field would remain broken.

app/views/prison/visits/_cancel_visit.html.erb
```diff
@@ -1,26 +1,29 @@
...
     </fieldset>
   </div>
+  <div class="panel panel-border-narrow">
+    <p>
+      <%= t(".cancel_to_nomis_warning")%>
+    </p>
+  </div>
```
StupidCodeFactory: @chrishanes could whisper in some css to fix the alignment and possibly make this message more user friendly?

screen shot 2017-10-31 at 16 18 18

config/locales/en/prison_views.yml
```diff
@@ -229,6 +228,7 @@ en:
         issue_with_date: Issue with the date
         visitor_banned: Visitor is banned
         issue_with_visitor: Issue with a visitor
+        cancel_to_nomis_warning: This booking will automatically be cancelled in NOMIS
```
StupidCodeFactory: @kayesoutham could you provide me with some adequate content to replace this with? 🙏  🍰  🎃

chrishanes: @StupidCodeFactory Should we be giving the user the chance to skip this as we do when processing the visit? A checkbox to "Don't automatically cancel this visit to NOMIS"?

app/views/prison/visits/_cancel_visit.html.erb
alan: @chrishanes is it the styling consistent with the red bar on the left when there is a validation error?

Is the problem that the form isn't using the form_helpers? If we use them I think we'll be able to remove the global config on `field_error_proc`

StupidCodeFactory: We are using the form helper `#error_container`, which btw could/should be used by field_error_proc as it is the case in the form_helper_gov_uk gem (or any other form libraries...) 

On another side not we are using error_container around pretty much every field, in the process form which we absolutely don't need since we never display validation around a specific field (except the date for the renewal)

config/locales/en/prison_views.yml
StupidCodeFactory: yes you're correct, I hadn't open the image in trello :/

alan: Also, this text is hardcoded, should be dependent on the feature being enabled.

alan:
What was the issue with the 404 in the api? I can't find the commit with the latest change after the latest conversation

app/views/prison/visits/_cancel_visit.html.erb
alan: 👍 we still should have the card for using the gem somewhere

StupidCodeFactory:
@kyliehavelock here is a url to test the feature: 
https://pvb-staff-pr-1142.herokuapp.com/en/prison/visits/997debd5-17d7-45e5-a463-574835bf4582

PR #1148, opened 2017-10-16
Implement use of email address validation gem

KPobeeNorris:
It was established that we were duplicating the email address checking section of our codebase across PVB2 & PVB-public, and so this was extracted out into a rubygem. This PR manages the implementation of the EmailAddressValidation gem.


spec/spec_helper.rb
```diff
@@ -1,7 +1,7 @@
 require 'simplecov'
 require 'simplecov-rcov'
 SimpleCov.formatter = SimpleCov::Formatter::RcovFormatter
-SimpleCov.minimum_coverage 100
+SimpleCov.minimum_coverage 99.79
```
StupidCodeFactory: I"m not keen on lowering the minimum coverage 

I'd rather address the lack of coverage which seems to be this file:
https://6329-44737633-gh.circle-artifacts.com/0/tmp/circle-artifacts.7iUgIsI/coverage/rcov/app-services-metrics_rb.html

StupidCodeFactory: @alan thoughtd?

StupidCodeFactory: (thoughtd being a background process that provides the application it's own personality obviously)

alan: Yes, we should add it in this case since the deleted specs tested that service class.

In pvb-public the coverage percentage (which is less than 100%) changed due to code being deleted so in that case it makes sense to me to update the percentage. There the untested code is unrelated to this gem work.

StupidCodeFactory: This drop in coverage is directly related to this change.
In fact after checking `app/services/metrics_rb` needs to be removed since it is not used anymore.
And we can put the coverage back up 100%.
We should investigate in pvb-public which it is not at 100% and make it so that it hits the 100%.

alan: Yep agreed about pvb-public, we can take a look at that after the gem work since it is unrelated.

KPobeeNorris: Thanks @StupidCodeFactory; I'll get update the PR now.

PR #1167, opened 2017-10-27
Avoid a memory leak in Excon

alan:
Current version of Excon has a memory leak when threads using an Excon
connection get destroyed.

Applies same fix as
https://github.com/ministryofjustice/prison-visits-public/pull/346

The memory leak in pvb2 is slower because Puma is configured to use less threads
so they probably get recycled less often.

Also sets a constant number of threads for Puma in production as we do in
pvb-public.

### DIFF
--- a/app/services/nomis/client.rb
+++ b/app/services/nomis/client.rb
@@ -6,11 +6,28 @@ require 'excon/middleware/deadline'
 module Nomis
   APIError = Class.new(StandardError)
 
+  # This client is **NOT** thread safe. To be used with a connection pool. See below.
   class Client
     TIMEOUT = 2 # seconds
     EXCON_INSTRUMENT_NAME = 'nomis_api'.freeze
     JSON_MIME_TYPE = 'application/json'.freeze
 
+    # Sets `thread_safe_sockets` to `false` on the Excon connection. Setting
+    # to `false` disables Excon connection sockets cache for each thread.
+    #
+    # This cache means that there is a socket connection for every Excon
+    # connection object to the Nomis API for each Thread. This is to make an
+    # Excon connection thread safe.
+    #
+    # For example if Puma runs with 4 threads, and those threads reuse a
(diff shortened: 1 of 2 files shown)


app/services/nomis/client.rb
```diff
@@ -6,11 +6,27 @@
...
     JSON_MIME_TYPE = 'application/json'.freeze
 
+    # Sets `thread_safe_sockets` to `false` on the Excon connection. Setting
+    # to `false` disables Excon connection sockets cache for each thread.
+    #
+    # This cache means that there is a socket connection for every Excon
+    # connection object to the Nomis API for each Thread. This is to make an
+    # Excon connection thread safe.
+    #
+    # For example if Puma runs with 4 threads, and those threads reuse a
```
StupidCodeFactory: "and those reusing"

app/services/nomis/client.rb
```diff
@@ -6,11 +6,27 @@
...
+    # connection pool of 4 Excon connections means that effectevely there can
+    # be up to 16 live connections to the Nomis API.
+    #
+    # This cache is unnecessary in our case since:
+    # - we ensure thread safety by using a connection pool
+    # - we end up opening more sockets than necessary (25 vs 5)
```
StupidCodeFactory: Can you provide more details, this isn't really clear why

app/services/nomis/client.rb
alan: I think my version is right. @KPobeeNorris can you clarify the English grammar rules in this case?

app/services/nomis/client.rb
alan: Better?

app/services/nomis/client.rb
KPobeeNorris: I believe that Alan's original comment is the correct case; just need to change the spelling of 'effectively'.

PR #1176, opened 2017-11-06
Email confirmed bookings csv

alan:
Instead of downloading the csv file it will email the file to the user.

### DIFF
--- a/app/controllers/metrics_controller.rb
+++ b/app/controllers/metrics_controller.rb
@@ -1,5 +1,6 @@
 class MetricsController < ApplicationController
   before_action :authorize_prison_request
+  before_action :authenticate_user, only: :send_confirmed_bookings
 
   def index
     @prisons = PrisonsDecorator.decorate(Prison.enabled.includes(:estate))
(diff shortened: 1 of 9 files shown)


app/controllers/metrics_controller.rb
```diff
@@ -1,18 +1,18 @@
...
 
   def index
     @prisons = PrisonsDecorator.decorate(Prison.enabled.includes(:estate))
     @dataset = MetricsPresenter.new(all_time_counts)
     @graphs_presenter = GraphMetricsPresenter.new
   end
 
-  def confirmed_bookings
-    exporter = WeeklyMetricsConfirmedCsvExporter.new(weeks: 12)
+  def send_confirmed_bookings
```
StupidCodeFactory: Personally not a huge fan of none standard actions in controllers as it's not what you would be expecting in a controller. Shouldn't be seeing any other public methods than `index, show, create, update, delete` in a controller.
One of the main benefits is that it prevents a long list of semantically incorrect `get,put,post,member,collection` macros in the routes file which quickly gets out of hand and typically does not get any love from developers.

The usual pattern is create a controller `SendConfirmedBookingsController` with a `#create` action a `resources :send_confirmed_bookings, only: create`.

app/mailers/admin_mailer.rb
```diff
@@ -12,4 +12,19 @@ def slot_availability(availability)
       subject: "Slot availability - #{Time.zone.today}"
     )
   end
+
+  def confirmed_bookings(email_address)
+    exporter = WeeklyMetricsConfirmedCsvExporter.new(weeks: 12)
+
+    attachments['confirmed_bookings.csv'] = {
```
StupidCodeFactory: Should we be checking the size of the attachments? email provider will be rejecting the email if this goes over a certain size, although we should be ok don't think this might go over 25MG.

app/controllers/metrics_controller.rb
alan: If this was a RESTful API it would make sense, I see little benefit in this case

app/mailers/admin_mailer.rb
alan: Yes, thought of that. I think we have a long while until we get a large csv

PR #1178, opened 2017-11-09
Erlestoke-add-Sunday-and-Christmas-visits

kayesoutham:
Erlestoke are adding Sunday visits from 3/12/17. Adding these an anomolous so visitors can book in advance of 3/12. Will raise a separate PR to add recurring Sunday visits from 3/12. Have also added unbookable Christmas visit slots to this PR.


StupidCodeFactory:
@kayesoutham I think there is a problem with the 24th, don't think it should be in the unbookable since it is in the anomalous

The test import  fails with the following error:
`Slot details An unbookable date may not also be anomalous in /home/circleci/prison-visits-2/db/seeds/prisons/EEI-erlestoke.yml`

PR #1202, opened 2017-11-14
Increase Excon Timeout on the create booking request

StupidCodeFactory:
**CHANGES:**

Increase timeout to 3 seconds when making the #book_visit


app/services/nomis/api.rb
```diff
@@ -116,8 +116,14 @@ def book_visit(offender_id:, params:)
       idempotent = params.key?(:client_unique_ref)
 
       response = @pool.with { |client|
-        client.post(
-          "offenders/#{offender_id}/visits/booking", params, idempotent: idempotent)
+        begin
```
alan: I think a nicer approach would be to pass the timeout to `connection#request` so that we don't have to change the timeout default on the client, and instead setting it for this call only.

From this file the call would look like:
```ruby
client.post("offenders/#{offender_id}/visits/booking", params, idempotent:, timeout: 3)
```

and then modify the client so that it sets the timeout options on the `request` method.

Doing so would mean that we don't have to change the default timeout on the connection, instead we would be setting the timeout for that one request which is what we want in this case. Then the `begin/ensure` can go and testing the client doesn't have to test the instance variables.

What do you think?

spec/services/nomis/client_spec.rb
```diff
@@ -24,6 +24,24 @@
     end
   end
 
+  # rubocop:disable RSpec/InstanceVariable
+  describe 'timeoout=' do
```
alan: typo

StupidCodeFactory:
🤦‍♂️  of course one can shove anything in a giant hash...

PR #1219, opened 2017-11-15
Display internal location

StupidCodeFactory:
**CHANGES:**

* Add internal location to our Establishment DAO
* Show internal location with the location on the booking process/show page

**NOTES:**
Update prod and staging environement PR for Feature Flag https://github.com/ministryofjustice/pvb2-deploy/pull/129

### DIFF
--- a/app/services/nomis/api.rb
+++ b/app/services/nomis/api.rb
@@ -48,7 +48,11 @@ module Nomis
         client.get("/offenders/#{noms_id}/location")
       }
 
-      Nomis::Establishment.new(response['establishment']).tap do |establishment|
+      attributes = response['establishment']
+      if response['internal_location'].present?
+        attributes['internal_location'] = response['internal_location']
+      end
+      Nomis::Establishment.new(attributes).tap do |establishment|
         PVB::Instrumentation.append_to_log(lookup_offender_location: establishment.code)
       end
     end
--- a/app/services/nomis/api.rb
+++ b/app/services/nomis/api.rb
@@ -48,11 +48,7 @@ module Nomis
         client.get("/offenders/#{noms_id}/location")
       }
 
-      attributes = response['establishment']
-      if response['internal_location'].present?
-        attributes['internal_location'] = response['internal_location']
-      end
-      Nomis::Establishment.new(attributes).tap do |establishment|
+      Nomis::Establishment.build(response).tap do |establishment|
         PVB::Instrumentation.append_to_log(lookup_offender_location: establishment.code)
       end
     end
--- a/app/services/prisoner_location.rb
+++ b/app/services/prisoner_location.rb
@@ -12,6 +12,10 @@ class PrisonerLocation
     self.prison_code = prison_code
   end
 
+  def internal_location
+    prisoner_location.internal_location if valid?
+  end
+
 private
 
   attr_accessor :offender, :prison_code
@@ -36,6 +40,7 @@ private
   end
 
   def load_prisoner_location
+    return Nomis::Establishment.new unless offender.valid?
     Nomis::Api.instance.lookup_offender_location(noms_id: offender.noms_id)
   rescue Nomis::APIError => e
     Raven.capture_exception(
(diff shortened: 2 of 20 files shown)


app/views/prison/visits/_prisoner_details.html.erb
```diff
@@ -64,13 +64,36 @@
...
+      <div class="border-box border-box--large border-box--orange">
+        <h3 class="bold-medium">
+          <%= t('.prisoner_restrictions_apply') %><div class="push-left--half font-xsmall tag tag--verified"><%= t('.verified') %></div>
+        </h3>
+        <% @visit.prisoner_restrictions.each do |restriction| %>
+          <%= render 'prisoner_restriction', restriction: restriction %>
+        <% end %>
+      </div>
+    <% else %>
+      &nbsp;
```
StupidCodeFactory: @chrishanes is this an exceptable hack for you?

chrishanes: Why are the restrictions in a column 1/4 now?

StupidCodeFactory: @chrishanes

o-no-idea-dog-meme-facebook

StupidCodeFactory: doing

app/services/nomis/api.rb
```diff
@@ -48,7 +48,11 @@ def lookup_offender_location(noms_id:)
         client.get("/offenders/#{noms_id}/location")
       }
 
-      Nomis::Establishment.new(response['establishment']).tap do |establishment|
+      attributes = response['establishment']
```
alan: I think is cleaner and easier to test to push this logic to a class method on `Nomis::Establishment`, like we do on the [Booking class](https://github.com/ministryofjustice/prison-visits-2/blob/e891d6f726e908f7d844015d825c21698fb890a8/app/services/nomis/booking.rb#L8)

app/services/prisoner_location.rb
```diff
@@ -12,6 +12,10 @@ def initialize(offender, prison_code = nil)
     self.prison_code = prison_code
   end
 
+  def internal_location
+    prisoner_location.internal_location if valid?
```
alan: This object now gets `valid?` called multiple times in its lifecycle for no need. When we build the object we could call `valid?` on it and remove the other calls on it.

alan: Also, is the check needed here? When the object is invalid it returns `nil`, which is the same value that an instance of `Nomis::Establishment` would return.

app/services/prisoner_location.rb
```diff
@@ -36,6 +40,7 @@ def located_at_given_prison
   end
 
   def load_prisoner_location
+    return Nomis::Establishment.new unless offender.valid?
```
alan: What do you think of having `NullEstablishment` similar to the `NullOffender`?

alan: The naming of the `*prisoner_location` methods on this class (called `PrisonerLocation`) had me confused for a while as I thought they were returning instances of `PrisonerLocation`.

I think it would be clearer to rename them to something different like `establishment, load_establishment,...`

spec/services/prisoner_location_spec.rb
```diff
@@ -1,43 +1,81 @@
...
+  let(:noms_id)        { 'a1234bc' }
   let(:prison_code)    { 'BMI' }
-  let(:establishment)  { Nomis::Establishment.new(code: code, api_call_successful: api_successful) }
+  let(:code)           { prison_code }
+  let(:api_successful) { true }
+  let(:establishment)  do
+    Nomis::Establishment.new(
+      code: code,
+      api_call_successful: api_successful,
+      internal_location: 'Cell With View Overlooking the Ocean'
```
alan: I want to reserve one of these

app/services/prisoner_location.rb
StupidCodeFactory: I don't see much problem with an object asking itself about its internal state.

So here we do need to ask if it is valid or not because the prisoner might not be located at the prison provided by the visitor.
I'm going to rename this object to `PrisonerLocationValidation` to make the intent of the object a bit clearer.

StupidCodeFactory:
@alan @KPobeeNorris If you can have a quick  look, I've added a feature flag for the internal location

PR #1244, opened 2017-11-16
Refactor VisitDecorator and StaffNomisChecker

StupidCodeFactory:
**CHANGES:**

This changeset aims at extracting some of the logic inside the VisitDecorator to
a PrisonerDetailsPresenter.

the high level API interacting objects are shared between decorators and controller as helper methods, so easier to share/inject between decorators/view layer and service objects.

I think our decorators are doing a little too much now and we could use higher level objects between the views/decorators and service objects,  that would have the downside of having one more indirection level but on the flip side overall simpler Decorators and Service objects, easier to compose.

### DIFF
--- a/app/presenters/prisoner_details_presenter.rb
+++ b/app/presenters/prisoner_details_presenter.rb
@@ -20,7 +20,7 @@ class PrisonerDetailsPresenter
     end
   end
 
-  def prisoner_details_incorrect?
+  def details_incorrect?
     prisoner_existance_status == INVALID
   end
 
--- a/app/presenters/prisoner_details_presenter.rb
+++ b/app/presenters/prisoner_details_presenter.rb
@@ -1,0 +1,42 @@
+class PrisonerDetailsPresenter
+  VALID    = 'valid'.freeze
+  INVALID  = 'invalid'.freeze
+  NOT_LIVE = 'not_live'.freeze
+
+  def initialize(prisoner_validation, prisoner_location)
+    self.prisoner_validation = prisoner_validation
+    self.prisoner_location   = prisoner_location
+  end
+
+  def prisoner_existance_status
+    return NOT_LIVE unless Nomis::Api.enabled?
+    case prisoner_existance_error
+    when nil
+      VALID
+    when PrisonerValidation::UNKNOWN, PrisonerLocation::INVALID, PrisonerLocation::UNKNOWN
+      prisoner_existance_error
+    else
+      INVALID
+    end
+  end
+
+  def prisoner_details_incorrect?
+    prisoner_existance_status == INVALID
+  end
+
+  def prisoner_existance_error
+    prisoner_validation_errors.first || prisoner_location_errors.first
+  end
+
+  def prisoner_validation_errors
+    @prisoner_validation_errors ||= prisoner_validation.tap(&:valid?).errors.full_messages
+  end
+
+  def prisoner_location_errors
+    @prisoner_location_errors ||= prisoner_location.tap(&:valid?).errors.full_messages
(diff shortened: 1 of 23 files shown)


app/presenters/prisoner_details_presenter.rb
```diff
@@ -0,0 +1,42 @@
...
+    when nil
+      VALID
+    when PrisonerValidation::UNKNOWN, PrisonerLocation::INVALID, PrisonerLocation::UNKNOWN
+      prisoner_existance_error
+    else
+      INVALID
+    end
+  end
+
+  def prisoner_details_incorrect?
```
alan: How about `#incorrect_details?` instead? It doesn't read quite right where is used `prisoner_details.prisoner_details_incorrect?` to me. Similar to the rest of the methods.

app/services/book_to_nomis_config.rb
```diff
@@ -1,9 +1,13 @@
...
-  def initialize(staff_nomis_checker, prison_name, opted_in, already_booked_in_nomis)
-    self.staff_nomis_checker = staff_nomis_checker
-    self.prison_name = prison_name
-    self.opted_in = opted_in
-    self.already_booked_in_nomis = already_booked_in_nomis
+  def initialize(staff_nomis_checker,
+    prison_name, opted_in,
+    already_booked_in_nomis,
+    prisoner_details_presenter)
+    self.staff_nomis_checker        = staff_nomis_checker
```
alan: What do you think of using `RequestStore` for these shared objects? It looks like nomis checker, presenter, etc are needed across the request cycle. They are kind of globals already in the view.

An advantage would be that this class could have it's number of constructer arguments trimmed.

This is an idea I've had just now, not really sure if it would actually be a good idea...

app/services/prisoner_location.rb
```diff
@@ -0,0 +1,46 @@
...
+    self.prison_code = prison_code
+  end
+
+private
+
+  attr_accessor :offender, :prison_code
+
+  def validate_has_location
+    unless prisoner_location.api_call_successful?
+      errors.clear
```
alan: Why do you need to clear the errors? Usually errors are clear at the beginning of the validation lifecycle

app/views/prison/visits/_prisoner_details.html.erb
```diff
@@ -1,20 +1,20 @@
 <h2 class="bold-medium"><%= t('.title') %></h2>
 <% case @visit.prisoner_existance_status %>
-<% when StaffNomisChecker::VALID %>
+<% when PrisonerDetailsPresenter::VALID %>
```
alan: I think it would be better to consolidate this constant values across the classes so one doesn't have to know the particular namespace. Something like `Constants::VALID`, `Constants::INVALID`, etc.

What do you think?

app/services/prisoner_location.rb
StupidCodeFactory: That's to ensure that if an API error happens we only get this objects. 
Both validations get run and we'd get two errors in the case of an API error, doesn't really makes sense to have an `invalid location` when the API actually failed.
Another way to avoid `errors.clear` was to run `validate_has_location` first and then but `located_at_given_prison`
and check for `api_call_successful?` but I didn't quite like it because it implicitely requires validations to be run in a specific order.

app/views/prison/visits/_prisoner_details.html.erb
StupidCodeFactory: I thought about this too but I prefered to keep the constant close to were their values are used as a result of computation, makes it easier to navigate the code

PR #1297, opened 2017-11-23
Fix populate rake task

alan:
Set a rejection reason detail if visit is going to be rejected for other reason.

### DIFF
--- a/lib/tasks/populate.rake
+++ b/lib/tasks/populate.rake
@@ -102,7 +102,13 @@ namespace :pvb do
                      }
                    end
 
+      if attributes[:rejection_attributes][:reasons].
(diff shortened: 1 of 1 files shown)


lib/tasks/populate.rake
```diff
@@ -102,7 +102,13 @@ namespace :pvb do
                      }
                    end
 
+      if attributes[:rejection_attributes][:reasons].
+          include?(Rejection::OTHER_REJECTION_REASON)
+        attributes[:rejection_attributes][:rejection_reason_detail] = "It's complicated"
```
StupidCodeFactory: sounds more like a facebook status 🤣

PR #1352, opened 2017-12-21
Hotfix: Fixed issue with failing migration and new containter being up

lukaszraczylo:


run.sh
```diff
@@ -9,7 +9,13 @@ create)
 migrate_and_seed)
     echo "running migrate and seed"
     bundle exec rake db:migrate db:seed
+    RETURN_CODE=$?
+    while [ $RETURN_CODE -eq 1 ]; do
```
StupidCodeFactory: I could see a couple issues:
1. Probably want to retry while the code is different than `0`, rake return code might not always return 1 in case of failure.
2. Probably only want to retry a limited amount of times

I think this PR is resulting from one of our migration timing out. We fail migration that takes longer than a couple seconds to avoid agressive table lock, typically when modifying a column defaults. 
I don't think this would fixed the issue we had since it would keep retrying the migration.

run.sh
```diff
@@ -9,7 +9,13 @@ create)
...
+      sleep 1
+      bundle exec rake db:migrate db:seed
+      RETURN_CODE=$?
+    done
     ;;
 esac
-bundle exec puma -b tcp://0.0.0.0:3000 -d -C config/puma_prod.rb
-tail -f /usr/src/app/log/logstash_production.json
+#bundle exec puma -b tcp://0.0.0.0:3000 -d -C config/puma_prod.rb
```
StupidCodeFactory: Should't this script attempt to start the server after successfully running the migrations?

run.sh
lukaszraczylo: @StupidCodeFactory thank you for input.
I just changed eq to gt than 0 as per suggestion.
We want to retry as long as it fails because we already have new containers on the instances which won't serve any traffic unless you can either perform manual failback or the migration will succeed.

lukaszraczylo:
@StupidCodeFactory problem we have is that migration occurs at the end of the deployment process as it's triggered from within newly deployed container anyway. Only way to fall back with failed migration in scope would be to deploy older version of the container which needs to be done manually anyway.

Scenario to consider:
Project deploys but migration fails - we have new containers running new code but being unable to render the website, therefore the application is unreachable ( that's what the trying to re-run migration over and over protects us from ) unless migration succeeds - no fallback is required but it can take few attempts.

StupidCodeFactory:
@lukaszraczylo that's how i understood the deployment process. 
This PR will just try to re-run any failing migration until somebody fixes the migration it (rarely seen migrations fix themselves appart from network glitches) but the site will remain unreachable.
Not a big change of behaviour. 

👍

PR #1406, opened 2018-01-28
[DoNotMerge] Decrease size of pvb image by using docker multistep builds

lukaszraczylo:


lukaszraczylo:
Current result:
```pvb latest 884MB
pvb-small lates 277MB
```
It's worth noting that we can run test as a part of build process ( separate step before the final release build ), so it's definitely worth to investigate.

lukaszraczylo:
@StupidCodeFactory it actually does - as long as we can spin up the container with all the environmental variables. It's doable and would cut the number of builds required to do everything to 1.

Example workflow:
- code pushed to github
- base image build triggered automatically
- followed by tests running within the image
- ending up with building "final" and deployable image

As another build step ( CI wise ) we can deploy that image straight away to the desired environment as we know it's been tested and all tests were successful.

alan:
What's remaining to merge this?

lukaszraczylo:
@alan it was a POC that we can do something to make our builds more efficient, do more things and as a result produce smaller images. As a CPT we will have a look at bringing this Docker feature into our organisation ( and with cooperation between CPT and product teams ) we are also willing to make running tests within containers possible.

StupidCodeFactory:
@TobyRet you should probably ask @lukaszraczylo.

lukaszraczylo:
@TobyRet please leave this PR as do not merge - i'm using this branch to work on universal image for all the ruby powered projects.

PR #1413, opened 2018-02-02
Update print visits date picker

StupidCodeFactory:
The date picker on the print visits page did not match the GDS standard
as was a single field, rather than three separate fields.


StupidCodeFactory:
@KPobeeNorris I think this is missing all the newly created files :)

spec/features/print_visits_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+  before do
+    OmniAuth.config.add_mock(:mojsso, sso_response)
+    vst.update!(slot_granted: '2017-12-21T14:00/16:00')
+  end
+
+  context 'when logging in and navigating to the print visits page' do
+    it 'displays the print visit page' do
+      visit new_prison_print_visit_path
+      expect(page).to have_css('.navigation')
+      expect(page).to have_content('Select one or more prisons')
```
StupidCodeFactory: avoid `have_content` they're slow. always use have_css

PR #1415, opened 2018-02-05
Remove Virtus

alan:
Removes Virtus and uses the new upcoming feature in Rails 5.2 that supports
typed attributes.

The main difference between Virtus and Rails approach is that Virtus are not
really types, it uses coercers so the resulting type can be different. In Rails
you specify the final type.

Another difference is that Rails doesn't support typed arrays, what in Virtus
can be written as `Array[Date]` for example. The way I ported these is by
creating specific classes for each typed array. This is not ideal because there
is a few very similar classes. This is something that can be improved when the
Rails API evolves to support (ActiveRecord when used with postgres supports
array types but it hasn't been ported to ActiveModel)

The final main difference is that Virtus allowed undeclared fields to be used
while ActiveModel doesn't.


app/models/concerns/memory_model.rb
```diff
@@ -1,14 +1,12 @@
-module NonPersistedModel
+module MemoryModel
```
StupidCodeFactory: What about `ValueObject`?  These obects aren't doing anything, basically just wrappers around API responses.

app/types/date_list_type.rb
```diff
@@ -0,0 +1,7 @@
+class DateListType < ActiveModel::Type::Value
+  def cast(value)
```
StupidCodeFactory: Should we keep thinkgs consistant and call `ContactsEnumerableType`  => `ContactsListType`

`s/enumberable/list`

alan: `ContactEnumerable` (as a typed list) couldn't name it `ContactList` because there is already a class called `ContactList` and the type cast classes lookup the type based on their name by default so had to call them `ContactEnumerable` and `ContactEnumerableType`.

There would be a lot less boiler plate code and easier naming if attributes supported supported arrays like it does for Postgres.

i.e.

```ruby
attribute :prices, :integer, array: true
```

app/models/concerns/memory_model.rb
alan: Well, there aren't used just for API responses.

The issue I have with using that is that for me Value objects should ideally be immutable objects which `ActiveModel::Model` aren't because it doesn't support read only, and in some use cases they behave like Rails like models with validations, etc

I used `MemoryModel` as a way to port one by one all the old Virtus classes, I could now rename it back to `NonPersistedModel`

PR #1430, opened 2018-02-08
remove virtus - Rails 5.2 upgrade - final part

StupidCodeFactory:
This is a WIP. It will need some more work to integrate the offical Rails 5.2

**CHANGES:**

Removes Virtus which was a dead gem.

**NOTES:**

Originally the work was done by @alan but the active model attributes gem  used didn't quite work end we ended up with a broken master branch in development environment (class loading / autoloading problems).
However most of this work will be relevant when rails 5.2 ships with the official version of the actibve model attributes


app/services/nomis/contact.rb
```diff
@@ -1,18 +1,29 @@
...
-    attribute :contact_type, Hash[Symbol => String]
-    attribute :approved_visitor, Boolean
-    attribute :active, Boolean
-    attribute :restrictions, Array[Restriction]
+    attribute :id, :integer
+    attribute :given_name, :string
+    attribute :middle_names, :string
+    attribute :surname, :string
+    attribute :date_of_birth, :date
+    attribute :gender
```
TobyRet: @StupidCodeFactory - is a default used if a type is not declared?

StupidCodeFactory: good catch will add that in

spec/types/api_slot_list_type_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'rails_helper'
+require_relative 'shared_type_examples'
+
+RSpec.describe ApiSlotListType do
```
TobyRet: @StupidCodeFactory - Sorry I'm not sure what these type specs are testing  Could you explain pls?

TobyRet: - my guess is that it this checking that the ApiSlotListType is enumerable else it raises an error ??

app/types/prison_type.rb
```diff
@@ -0,0 +1,5 @@
+class PrisonType < ActiveModel::Type::Value
+  def cast(value)
```
TobyRet: If input == output is casting necessary in this instance?

StupidCodeFactory: Yeah I'm affraid that's how active model attributes work.

💡  **However** nothing prevents us from writing attr_accessor method and get rid of this nonsense!

StupidCodeFactory: So I'll amend that

PR #1447, opened 2018-02-15
Always display reference num option

KPobeeNorris:
Following the trial rollout of 'Book To Nomis' to Featherstone we
received feedback from staff stating that they still need the option of
adding a reference number to each visit.  The feature had been
implemented in such a way that if a visit could be booked directly to 
Nomis then we no longer displayed the reference number, unless the 
user specifically opted out, or if it needed to be a closed visit.

Therefore, we have amended the feature to allow the user the option of
including a reference number regardless of whether the prison has the
'Book To Nomis' feature enabled or not.


StupidCodeFactory:
can we fill in the refenrence_no in the book to nomis feature tests?

PR #1492, opened 2018-03-06
Fix prisoner details

StupidCodeFactory:
**CHANGES:**

Split logic displaying the overall prisoner details NOMIS info/warning from individual checks


app/controllers/concerns/staff_response_context.rb
```diff
@@ -22,10 +22,18 @@ def nomis_checker
...
+      prisoner_validation
     )
   end
 
+  def nomis_info
+    @nomis_info ||= NomisInfo.new(prisoner_validation, prisoner_location_validation)
+  end
+
+  def prisoner_location
+    @prisoner_location ||= PrisonerLocation.new(prisoner_location_validation)
```
TobyRet: Is PrisonerLocation a presenter (like PrisonerDetailsPresenter)

StupidCodeFactory: Yup, just "presentered" up *all* of the presenters

PR #1513, opened 2018-03-13
Use static GA cookie for visit request tracking

chrishanes:
Each visit request is tracked (sent by PVB2) but the _ga cookie is not set as the event is triggered by public/API. This is creating a new user session in GA for every request which is effecting the stats/data.

This will set a single static cookie ID for each request.

### DIFF
--- a/app/services/ga_tracker.rb
+++ b/app/services/ga_tracker.rb
@@ -18,24 +18,25 @@ class GATracker
 
   def send_unexpected_rejection_event
     if visit_rejected_unexpectedly?
-      send_data(build_event_payload('Manual rejection',
+      send_data(build_event_payload(ga_cookie, 'Manual rejection',
         visit.rejection.reasons.sort.join('-')))
     end
   end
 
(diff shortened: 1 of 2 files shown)


app/services/ga_tracker.rb
```diff
@@ -18,24 +18,25 @@ def set_visit_processing_time_cookie
...
     if visit.rejected?
-      send_data(build_event_payload('Rejection',
+      send_data(build_event_payload(ga_cookie, 'Rejection',
         visit.rejection.reasons.sort.join('-')))
     end
   end
 
   def send_request_event
-    send_data(build_event_payload('Request', visit.slots.count))
+    send_data(build_event_payload('GA1.1.123456789.0123456789',
```
TobyRet: @chrishanes @KPobeeNorris - shouldn't the first parameter refer to 'ga-cookie'?

chrishanes: No. The cookie will not exist, so we need to set a static one

PR #1525, opened 2018-03-19
Change feedback link to banner

chrishanes:
Change the Feedback link from Survey Monkey to our Zendesk contact page as no one is tasked to monitoring this feedback or managing the expectation of the staff who take the time out to use this channel.

### DIFF
--- a/config/locales/en/prison_views.yml
+++ b/config/locales/en/prison_views.yml
@@ -3,6 +3,8 @@ en:
     none: None
   layouts:
     application:
+      contact_banner_html: |
+        <a href='%{url}'>Contact us</a> with feedback or to let us know about a problem.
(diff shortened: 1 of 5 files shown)


config/locales/en/prison_views.yml
```diff
@@ -3,6 +3,8 @@ en:
     none: None
   layouts:
     application:
+      contact_banner_html: |
+        <a href='%{url}'>Contact us</a> with feedback or to let us know about a problem.
```
TobyRet: Do we need to add a Welsh translation as well or is this for another ticket?

chrishanes: No, this is for staff where we don't translate into Welsh

PR #1527, opened 2018-03-19
Remove 'iep_level_enabled' feature flag

TobyRet:
Part of a larger piece of work to remove redundant feature flags
https://trello.com/c/DKCxsMS3/1274-remove-unnecessary-flags

### DIFF
--- a/app/services/nomis/offender.rb
+++ b/app/services/nomis/offender.rb
@@ -27,9 +27,15 @@ class Nomis::Offender
 private
 
   def details
-    @details ||= Nomis::Api.instance.
-                   lookup_offender_details(noms_id: noms_id)
-  rescue Nomis::APIError
+    @details ||= Nomis::Api.instance.lookup_offender_details(noms_id: noms_id)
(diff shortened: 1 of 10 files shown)


app/services/nomis/offender.rb
```diff
@@ -27,9 +27,15 @@ def imprisonment_status
 private
 
   def details
-    @details ||= Nomis::Api.instance.
-                   lookup_offender_details(noms_id: noms_id)
-  rescue Nomis::APIError
+    @details ||= Nomis::Api.instance.lookup_offender_details(noms_id: noms_id)
+  rescue Nomis::DisabledError
```
TobyRet: This is an interesting one ...
The 'process_a_request_nomis_disabled_spec.rb' switches off the nomis api by returning false when calling 'enabled?'. The initialiser will throw a Nomis::DisabledError if the api is not enabled. Therefore we need to resuce from this exception in order to not break the test :-( 
Would be worth revisiting this logic when Yann returns

PR #1537, opened 2018-03-21
Remove NOMIS_STAFF_PRISONER_CHECK_ENABLED flag

KPobeeNorris:
This feature flag is no longer required as the prisoner checks take
place at all prisons and therefore can be removed.


spec/services/staff_nomis_checker_spec.rb
```diff
@@ -63,10 +63,6 @@
     let(:slot) { visit.slots.first }
 
     context 'with prisoner availability' do
```
TobyRet: @KPobeeNorris  - is the context 'with prisoner availability' still relevant?

PR #1557, opened 2018-03-26
Feltham 18-21 slot update

kayesoutham:
Changing the end time for the Saturday AM slot to match NOMIS


StupidCodeFactory:
@kayesoutham don't worry about the failure. It broke the build like a fool.

PR #1583, opened 2018-04-03
Format date selection display

chrishanes:
In order to facilitate the design for PVO & VO selection correctly, we will need to move the radio button for No Dates Available to be inline to date selections. This will allow the date selections to be interacted with like a radio button selection.

We're also highlighting the Clear Selection by making the link into a secondary button.

### DIFF
--- a/app/helpers/date_helper.rb
+++ b/app/helpers/date_helper.rb
@@ -1,15 +1,19 @@
 module DateHelper
-  def format_date_of_birth(date)
-    I18n.l(date.to_date, format: :date_of_birth)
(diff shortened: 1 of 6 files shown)


app/helpers/date_helper.rb
```diff
@@ -1,4 +1,8 @@
 module DateHelper
+  def format_date(date)
```
StupidCodeFactory: The argument could probably be `datetime`

chrishanes: Have changed to datetime for this and other methods for clarity

PR #1600, opened 2018-04-12
Remove old feature flags from docs

jennyd:
We've recently [removed some flags](https://trello.com/c/DKCxsMS3/1274-remove-unnecessary-flags) around features which had already been fully rolled out. Remove the docs for those flags as well.

Also sort out some whitespace issues in the same file.


StupidCodeFactory:
@jennyd should we add a pre-commit hook to fail the commit in case of trailing whitespace?

TobyRet:
A few more of these feature flags can be removed I think @jennyd - happy to push edits if you like

jennyd:
@TobyRet sure, go ahead...

StupidCodeFactory:
@jennyd @TobyRet @KPobeeNorris A while ago I created a card to add 
https://github.com/brigade/overcommit#precommit
Wasn't sure if it was a bit overkill, we may want to add more of those pre/post commits

PR #1601, opened 2018-04-12
Remove old routes without locale in the path

jennyd:
We used to serve the visit processing pages under `/prison/visits`, but last year (in 4b5f41b) we started including the locale in the path, so they're now under `/en|cy/prison/visits`. The old routes were marked as deprecated at the time but weren't removed soon afterwards - they can be removed now.

The controller specs need the locale param so that they use the new routes - they were still testing against the old routes.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -42,11 +42,6 @@ Rails.application.routes.draw do
   end
 
   namespace :prison do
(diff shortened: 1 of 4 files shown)


jennyd:
@TobyRet I've rebased ♻️

PR #1607, opened 2018-04-17
Pre-check prisoner details incorrect

StupidCodeFactory:
**CHANGES:**

When the prisoner location is invalid or unknow we'll pre-check the 'Prisoner
details incorrect' checkbox

### DIFF
--- a/spec/features/process_a_request_reject_spec.rb
+++ b/spec/features/process_a_request_reject_spec.rb
@@ -13,6 +13,18 @@ RSpec.feature 'Processing a request', :expect_exception, :js do
     travel_to(Date.new(2018, 4, 5)) { ex.run }
   end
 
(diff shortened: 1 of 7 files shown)


spec/features/process_a_request_reject_spec.rb
```diff
@@ -13,6 +13,19 @@
     travel_to(Date.new(2018, 4, 5)) { ex.run }
   end
 
+  describe 'when the prisoner is not registered at the prison', vcr: { cassette_name: 'prisoner_not_at_given_prison', record: :new_episodes } do
```
TobyRet: Is 'record_new_episodes' necessary here?

StupidCodeFactory: nope 😅

PR #1618, opened 2018-04-18
Types - add missing declarations and fix broken tests

TobyRet:


app/types/date_list_type.rb
```diff
@@ -1,5 +1,8 @@
+require 'typed_list/date_list'
+
 class DateListType < ActiveModel::Type::Value
   def cast(value)
-    value.map { |date| Date.parse(date) }
+    dates = value.map { |date| Date.parse(date) }
+    DateList.new(dates)
```
StupidCodeFactory: It's not clear to me why just returning an array of dates in the DateListType class doesn't work?

TobyRet: @StupidCodeFactory - yes this part worked originally. Just doing for consistency - that ok?

PR #1623, opened 2018-04-19
Remove unnecessary types

TobyRet:
See Trello card for context
https://trello.com/c/rP65LzmT/1341-refactor-remove-unnecessary-types


app/types/api_slot_list_type.rb
```diff
@@ -1,7 +1,5 @@
 class ApiSlotListType < ActiveModel::Type::Value
   def cast(value)
-    api_slots = value.map { |api_slot| Nomis::ApiSlot.new(api_slot) }
-
-    ApiSlotList.new(api_slots)
+    value.map { |api_slot| Nomis::ApiSlot.new(api_slot) }.dup.freeze
```
StupidCodeFactory: What's rational behing the dup? #map already returns a copy of the original "value"

TobyRet: Yep good point @StupidCodeFactory - typo on my part
Will remove

PR #1625, opened 2018-04-20
Remove load test data task

StupidCodeFactory:


app/services/load_test_data_remover.rb
```diff
@@ -0,0 +1,5 @@
+class LoadTestDataRemover
+  def self.run
+    Visit.joins(:visitors).where("visitors.first_name = 'Load'").where("visitors.last_name = 'Test'").destroy_all
```
jennyd: I think we also need to delete the prisoner record for the visit - creating a visit also [involves creating a prisoner](https://github.com/ministryofjustice/prison-visits-2/blob/6dc76d883d30db252f921e88d60ecdeb81422d12/app/services/booking_request_creator.rb#L48), but the direction of the association means that the `destroy_all` call here won't delete the prisoner (it should delete the visitors, though).

TobyRet:
Moving to alternative solution as discussed with @StupidCodeFactory @KPobeeNorris

PR #1628, opened 2018-04-25
Handle missing methods

KPobeeNorris:


app/services/nomis/api_serialiser.rb
```diff
@@ -0,0 +1,41 @@
+module Nomis
+  Error = Class.new(StandardError)
```
StupidCodeFactory: Let's do the following:

app/services/nomis/error.rb => `class Nomis::Error < SantardError; end`
app/services/nomis/error/unhandled_Api_error => `class Nomis::Error::UnhandledApiError < Nomis::Error; end`
app/services/nomis/error/not_found => `class Nomis::Error::Notfound < Nomis::Error; end`
app/services/nomis/error/disabled => `class Nomis::Error::DisabledError < Nomis::Error; end`

jennyd:
It would be great to have some more explanation of why we need to do this in the commit messages 💭 🎉 ✨

PR #1634, opened 2018-04-26
Remove event tracker for visit requests

chrishanes:
We cannot accurately record the number of visit requests without altering the number of staff user sessions.


spec/services/ga_tracker_spec.rb
```diff
@@ -167,34 +167,6 @@
     end
   end
 
-  describe '#send_request_event' do
-    context "when the visit was created" do
-      before do
-        switch_feature_flag_with :ga_id, web_property_id
-      end
-
-      it 'sends an event', vcr: { cassette_name: 'requested_visit_event' } do
```
StupidCodeFactory: You could also delete the VCR cassette `spec/fixtures/vcr_cassettes/requested_visit_event.yml`

chrishanes: Deleted

PR #1646, opened 2018-04-30
Delete a form submission if a ticket is successfully raised with Zendesk

TobyRet:
- Form submission is deleted if ticket successfully raised with Zendesk
- Form submission not deleted if problem occurs with raising a ticket

**Trello ticket**
https://trello.com/c/mBXoVNiQ/1337-gdpr-create-task-to-remove-public-feedback-a-month-after-its-marked-resolved-closed

### DIFF
--- a/app/jobs/zendesk_tickets_job.rb
+++ b/app/jobs/zendesk_tickets_job.rb
@@ -14,9 +14,7 @@ class ZendeskTicketsJob < ActiveJob::Base
       fail 'Cannot create Zendesk ticket since Zendesk not configured'
     end
 
-    ZendeskAPI::Ticket.create!(
-      Rails.configuration.zendesk_client,
-      ticket_attrs(feedback))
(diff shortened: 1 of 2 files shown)


app/jobs/zendesk_tickets_job.rb
```diff
@@ -14,16 +14,21 @@ def perform(feedback)
...
+    feedback.destroy! if ticket_raised?(feedback)
   end
 
 private
 
   # We have 2 Zendesk inboxes configured, one for the public that matches that
   # the service field is 'prison_visits' and another one for staff that matches
   # tickets tagged with 'staff.prison.visits'.
+
+  def ticket_raised?(feedback)
```
StupidCodeFactory: idiomatically in ruby methods that ends with a `?` returns true or false and have no side effects. I suggest you rename this method to `#raise_ticket!`

PR #1656, opened 2018-05-08
Delete zendesk tickets 

TobyRet:
Creation of a rake task that calls the Zendesk API and deletes zendesk tickets older than 12 months. Part of the GDPR work - https://trello.com/c/mBXoVNiQ/1337-gdpr-delete-feedback-submissions


spec/services/zendesk_cleaner_spec.rb
```diff
@@ -0,0 +1,42 @@
+require 'rails_helper'
+
+RSpec.describe ZendeskCleaner do
+  let(:client) { double(ZendeskAPI::Client) }
+  let(:ticket1) { double(ZendeskAPI::Ticket, id: 1) }
+  let(:ticket2) { double(ZendeskAPI::Ticket, id: 2) }
+  let(:ticket3) { double(ZendeskAPI::Ticket, id: 3) }
+  let(:tickets) { [ticket1, ticket2, ticket3] }
```
TobyRet: The Zendesk API search endpoint actually returns a ZendeskAPI::Collection - is it ok to just fake with an array?

app/services/zendesk_cleaner.rb
```diff
@@ -0,0 +1,31 @@
...
+      each_slice(100).
+      to_a
+  end
+
+  def twelve_months_ago
+    Time.zone.today.months_ago(12).strftime('%Y-%m-%d')
+  end
+
+  def client
+    unless Rails.configuration.try(:zendesk_client)
```
StupidCodeFactory: Why would `Rails.configuration` be nil?

app/services/zendesk_cleaner.rb
```diff
@@ -0,0 +1,31 @@
...
+      fail 'Cannot delete Zendesk tickets as Zendesk is not configured'
+    end
+
+    @client ||= Rails.configuration.zendesk_client
+  end
```
StupidCodeFactory: you want to wrap the whole thing in a `begin/end` block like so:
```ruby
def client
  @client ||= begin
      unless Rails.configuration.try(:zendesk_client)
        fail 'Cannot delete Zendesk tickets as Zendesk is not configured'
      end
      Rails.configuration.zendesk_client
    end
  end
```

spec/services/zendesk_cleaner_spec.rb
StupidCodeFactory: maybe you can use this
https://www.rubydoc.info/gems/zendesk_api/1.10.0/ZendeskAPI/Collection#%3C%3C-instance_method

app/services/zendesk_cleaner.rb
TobyRet: No longer required as discussed

spec/services/zendesk_cleaner_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+  let(:ticket2) { double(ZendeskAPI::Ticket, id: 2) }
+  let(:ticket3) { double(ZendeskAPI::Ticket, id: 3) }
+  let(:tickets) { double(ZendeskAPI::Collection) }
+  let(:ticket_ids) { [ticket1.id, ticket2.id, ticket3.id] }
+
+  subject { described_class.new }
+
+  context 'when Zendesk is configured' do
+    it 'successfully bulk deletes tickets older than 12 months old' do
+      Rails.configuration.zendesk_client = client
```
StupidCodeFactory: you want to use `set_configuration_with(:zendesk_client, client)` because here you're mutating the zendesk client for all the tests

app/services/zendesk_cleaner.rb
```diff
@@ -0,0 +1,27 @@
+class ZendeskCleaner
+  STAFF_INBOX = 'staff.prison.visits created'.freeze
+
+  def initialize(zendesk_client)
```
StupidCodeFactory: I think including  the `ZendeskClient`  module should do the trick rather than including it in the rake task

PR #1657, opened 2018-05-08
Validate booking window

StupidCodeFactory:
**CHANGES:**

Booking window must be is less than 28 days

### DIFF
--- a/app/models/prison.rb
+++ b/app/models/prison.rb
@@ -12,6 +12,10 @@ class Prison < ActiveRecord::Base
     presence: true, if: :enabled?
   validate :validate_unbookable_dates
 
+  validates :booking_window, numericality: {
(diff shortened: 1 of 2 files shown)


app/models/prison.rb
```diff
@@ -12,6 +12,10 @@ class Prison < ActiveRecord::Base
     presence: true, if: :enabled?
   validate :validate_unbookable_dates
 
+  validates :booking_window, numericality: {
+    less_than: PrisonSeeder::SeedEntry::DEFAULT_BOOKING_WINDOW + 1
```
jennyd: Can we use `less_than_or_equal_to` to avoid the `+ 1`?

jennyd:
Also it would be great to explain in the commit message why we're doing this!

StupidCodeFactory:
@jennyd just updated the commit message as well!

PR #1664, opened 2018-05-10
Bugfix: Delete old Zendesk tickets

TobyRet:
- Fix active job bug (do not pass in the Zendesk client)
- Write tests for Zendesk client for code coverage


StupidCodeFactory:
@TobyRet can you explain what the bug was?

app/jobs/zendesk_tickets_job.rb
```diff
@@ -19,8 +20,8 @@ def perform(client, feedback)
   # the service field is 'prison_visits' and another one for staff that matches
   # tickets tagged with 'staff.prison.visits'.
 
-  def ticket_raised!(client, feedback)
-    ZendeskAPI::Ticket.create!(client, ticket_attrs(feedback))
+  def ticket_raised!(feedback)
+    ZendeskAPI::Ticket.create!(self.class.client, ticket_attrs(feedback))
```
StupidCodeFactory: I'm really not convinced about the benefit of the client being a class level property.
What happens if two jobs tries to make a request to zendesk at the same time? :trollface: 

We can ameliorate this later though

TobyRet: @StupidCodeFactory - yes this a bit shit but you cannot add instance methods to classes inheriting from ActiveJob. Class methods are possible though. The alternative is to build the client and not use the module ... Happy to do this if you prefer, but then there is no need for a module ...

StupidCodeFactory: Wait whaaaaaaat? one can't add instance method to ActiveJob class?? What's the rational can you point me where this is stated?

spec/services/zendesk_client_spec.rb
```diff
@@ -0,0 +1,37 @@
+require 'rails_helper'
+
+RSpec.describe ZendeskClient do
+  subject { Class.new }
+
+  URL = 'https://zendesk_api.com'
+  USERNAME = 'bob'
+  TOKEN = '123456'
```
StupidCodeFactory: these should `let`

spec/services/zendesk_client_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  expectations = [
```
StupidCodeFactory: same for this statment

PR #1671, opened 2018-05-14
Add tracking for cancelled and withdrawn visits

chrishanes:


StupidCodeFactory:
@chrishanes all good, do you want to merge?

PR #1674, opened 2018-05-15
Remove unused database columns

KPobeeNorris:
As part of GDPR we have identified the following unused columns in our
database:
Visits:
  - override_delivery_error
  - delivery_error_type
Users:
  - encrypted_password

Therefore we need to remove them, and any data held in them.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20171121152149) do
+ActiveRecord::Schema.define(version: 2018_05_15_110205) do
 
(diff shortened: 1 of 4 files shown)


db/schema.rb
```diff
@@ -229,7 +226,7 @@
       SELECT count(*) AS count,
       vsc.visit_state
      FROM (visits v
-       JOIN visit_state_changes vsc ON (((v.id = vsc.visit_id) AND ((vsc.visit_state)::text = ANY ((ARRAY['booked'::character varying, 'rejected'::character varying])::text[])))))
```
StupidCodeFactory: Can you remove anything that's not related to deleting those three columns?

PR #1689, opened 2018-05-22
Fix ZendeskCleaner

TobyRet:
This PR fixes a bug with the ZendeskCleaner - it blatantly did not work in the first place.


app/services/zendesk/pvb_api.rb
```diff
@@ -0,0 +1,40 @@
+module Zendesk
+  class PvbApi
```
TobyRet: I'm not entirely sure about the naming here - what I am trying to do is to wrap the ZendeskAPI and abstract it away so we only expose what we need

app/services/zendesk/client.rb
```diff
@@ -0,0 +1,16 @@
+module Zendesk
+  class Client
+    include Singleton
```
TobyRet: This will deal with the concurrency issue raised a while back

StupidCodeFactory: what was raised a while back?

app/services/zendesk/pvb_api.rb
StupidCodeFactory: what about `Zendesk::PVBClient`?

app/services/zendesk/pvb_api.rb
```diff
@@ -0,0 +1,40 @@
...
+    def ticket_ids
+      @client.search(old_tickets_query).map(&:id)
+    end
+
+    def destroy_tickets(ids)
+      @client.tickets.destroy_many(ids: ids, verb: :delete).fetch
+    end
+
+    def twelve_months_ago
+      Time.zone.today.months_ago(12).strftime('%Y-%m-%d')
```
StupidCodeFactory: What about `12.months.ago.strftime('%Y-%m-%d')`?

StupidCodeFactory: It would be nice to have this be configurable with an ENV var

spec/services/zendesk/pvb_api_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+
+  subject { described_class.new(client) }
+
+  before do
+    set_configuration_with(:zendesk_url, url)
+    set_configuration_with(:zendesk_username, username)
+    set_configuration_with(:zendesk_token, token)
+  end
+
+  describe ':delete_old_tickets' do
```
StupidCodeFactory:
```ruby
describe '#delete_old_tickets' do
end
```

spec/services/zendesk/pvb_api_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+          ).
+          and_return(ticket)
+
+      expect(ticket).to receive(:save!).once
+
+      subject.raise_ticket(ticket_attributes)
+    end
+  end
+
+  def twelve_months_ago
```
StupidCodeFactory: why not a `let`?

app/services/zendesk/pvb_api.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    def raise_ticket(ticket_attrs)
+      request do |client|
+        ZendeskAPI::Ticket.create!(client, ticket_attrs)
+      end
+    end
+
+    def cleanup_tickets
+      destroy_tickets(fetch_ticket_ids) until fetch_ticket_ids.empty?
+    end
```
StupidCodeFactory: you want to:
1. store in a variable the result of the first call
2. check if empty
3. destroy tickets if any tickets left

spec/services/zendesk/pvb_api_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+    let(:query) do
+      {
+        query: "type:ticket tags:staff.prison.visits updated<#{twelve_months_ago}",
+        reload: true
+      }
+    end
+
+    before do
+      expect(zendesk_api_client).to receive(:search).
+        and_return(ticket_ids, ticket_ids, empty_ticket_ids)
```
StupidCodeFactory: You should not be calling fetch_tickets 3 times if there is only 1 page. See my previous comment

spec/services/zendesk/pvb_api_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+    end
+
+    before do
+      expect(zendesk_api_client).to receive(:search).
+        and_return(ticket_ids, empty_ticket_ids)
+      expect(zendesk_api_client).to receive(:tickets).and_return(tickets)
+    end
+
+    it 'deletes tickets that have not been updated in twelve months or less' do
+      allow(tickets).to receive(:destroy_many!).and_return(tickets)
```
StupidCodeFactory: I think you can get rid of that line and append the `and_return(tickets)` to lines 38..40

PR #1713, opened 2018-06-07
Only return results of visits that have been updated in the last six months

TobyRet:
https://trello.com/c/0E58iabh/1428-only-show-data-from-last-6-months-in-search-results

### DIFF
--- a/spec/services/estate_visit_query_spec.rb
+++ b/spec/services/estate_visit_query_spec.rb
@@ -106,6 +106,25 @@ RSpec.describe EstateVisitQuery do
           is_expected.to eq([booked])
         end
       end
+
+      context 'when visits have not been updated within six months' do
+        let!(:old_booked) { FactoryBot.create(:booked_visit, prison: prison, updated_at: 7.months.ago) }
+        let(:prisoner_number) { old_booked.prisoner.number.downcase + ' ' }
+
+        it 'does not return visits in search results' do
+          expect(instance.processed(limit: limit, query: prisoner_number)).to be_empty
+        end
+      end
+    end
+  end
+
+  shared_examples_for 'returns recent only' do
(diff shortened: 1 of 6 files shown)


spec/services/estate_visit_query_spec.rb
```diff
@@ -151,9 +170,14 @@
       FactoryBot.create(:visit, :requested, prison: prison)
     end
 
+    let!(:old_visit) do
+      FactoryBot.create(:visit, :requested, prison: prison, created_at: 7.months.ago, updated_at: 7.months.ago)
```
StupidCodeFactory: no need to use `FactoryBot` just call `create(:visit)`

spec/services/estate_visit_query_spec.rb
```diff
@@ -106,6 +106,25 @@
           is_expected.to eq([booked])
         end
       end
+
+      context 'when visits have not been updated within six months' do
+        let!(:old_booked) { FactoryBot.create(:booked_visit, prison: prison, updated_at: 7.months.ago) }
+        let(:prisoner_number) { old_booked.prisoner.number.downcase + ' ' }
```
StupidCodeFactory: is the `+ ' '` to see if the class trims the input?

spec/services/estate_visit_query_spec.rb
StupidCodeFactory: basically if you see `FactoryBot` in front of a `create` or `build` or `build_stubbed` and so on rip it off!

spec/services/estate_visit_query_spec.rb
TobyRet: @StupidCodeFactory - I don't think the trailing white space is required. I've removed both instances and tests pass.

PR #1714, opened 2018-06-08
Display visitor restrictions

chrishanes:
This shows a list of restrictions against each visitor.

Awaiting the restrictions feature flag to show/hide visitor restrictions as well as prisoner restrictions.


app/assets/javascripts/modules/moj.matchVisitors.js
```diff
@@ -211,6 +218,10 @@
       return isBanned;
     },
 
+    isContactRestricted: function(contact) {
+      return contact && contact.restrictions && contact.restrictions.length > 0? true : false;
```
TobyRet: @chrishanes - if / else statement  not required here? ie just return bool statement

app/services/staff_nomis_checker.rb
```diff
@@ -56,7 +56,7 @@ def offender
   end
 
   def prisoner_restrictions
-    if Nomis::Feature.offender_restrictions_info_enabled?(@visit.prison_name) &&
+    if Nomis::Feature.restrictions_info_enabled?(@visit.prison_name) &&
         offender.valid?
```
TobyRet: Perhaps outside the scope of this PR but wonder if we need consistency around terms 'prisoner' vs 'offender' - it feels like we are randomly using one or the other

docs/configuration.md
```diff
@@ -58,7 +58,7 @@ A comma separated list of prisons for which PVB2 API will return slots along wit
 are available at the requested prison.
 
 
-### `NOMIS_STAFF_OFFENDER_RESTRICTIONS_ENABLED`
+### `NOMIS_STAFF_RESTRICTIONS_ENABLED`
```
TobyRet: Is the use of 'staff' or potentially redundant or misleading?  The restrictions relate to an offender or visitor  ... Perhaps we no longer need to refer to 'staff'

StupidCodeFactory: It's convinient to have STAFF in env var names as some of the env vars are to turn on flags for the public site

TobyRet: Cool

app/assets/javascripts/modules/moj.matchVisitors.js
chrishanes: What do you mean @TobyRet ?

StupidCodeFactory: I think you could drop the `? true : false` as what's before will return true or false

StupidCodeFactory: i think...

chrishanes: Ah, OK

TobyRet:
From the PR description you say you are ' ... Awaiting the restrictions feature flag to show/hide visitor restrictions'
Has this requirement been met?

app/services/staff_nomis_checker.rb
TobyRet: Will add a ticket

PR #1715, opened 2018-06-08
Restrict print visits search to 6 months

KPobeeNorris:
At present staff are able to enter a date into the print visits and it
will return any visits as far back as they want to search.  However, we
want to establish whether or not staff actually need to search for
visits from further back than 6 months.  As such we are going to restrict
their searches to the last 6 months and advise them to contact us if
they need more historical visits data.

### DIFF
--- a/app/controllers/prison/print_visits_controller.rb
+++ b/app/controllers/prison/print_visits_controller.rb
@@ -8,7 +8,7 @@ class Prison::PrintVisitsController < ApplicationController
 
   def create
     @print_visits = ::PrintVisits.new(permitted_visit_date_params)
-    create_visit_date
+    validate_visit_date
 
     @data = EstateVisitQuery.new(current_estates).visits_to_print_by_slot(@submitted_date)
 
@@ -23,10 +23,11 @@ class Prison::PrintVisitsController < ApplicationController
 
 private
 
-  def create_visit_date
+  def validate_visit_date
     @submitted_date = AccessibleDate.new(
       date_to_accessible_date(@print_visits.visit_date)
       ).to_date
+    @print_visits.valid? unless @submitted_date.nil?
   end
 
   def date_to_accessible_date(date)
--- a/app/models/print_visits.rb
+++ b/app/models/print_visits.rb
@@ -2,4 +2,19 @@ class PrintVisits
   include MemoryModel
 
   attribute :visit_date, :accessible_date
+
+  validate :check_date
+
+  def check_date
+    today = Time.zone.today
+    six_months_ago = 6.months.ago.to_date
+    date_to_check = visit_date.to_date
+
+    unless date_to_check.between?(six_months_ago, today)
+      errors.add(
+        :base,
+        I18n.t('search_date_invalid_html',
+          scope: %i[print_visits errors]))
+    end
+  end
 end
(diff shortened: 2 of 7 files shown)


app/controllers/prison/print_visits_controller.rb
```diff
@@ -27,6 +27,7 @@ def create_visit_date
     @submitted_date = AccessibleDate.new(
       date_to_accessible_date(@print_visits.visit_date)
       ).to_date
+    @print_visits.check_date(@submitted_date.strftime('%F')) unless @submitted_date.nil?
```
TobyRet: This method is doing more than just creating a date ... should it's name be updated accordingly (or perhaps move the changes to its own method ...)

TobyRet: Rather than calling 'check_date?' shouldn't the validation be invoked via 'valid?'

spec/models/print_visits_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+      it 'has an error on the field' do
+        subject.check_date('1-1-2017')
+        expect(subject.errors.messages).to include(base: ["Please choose a date within the last six months, or <a href='%{url}'>contact us</a> if you would like to see older visits."])
+      end
+    end
+
+    context 'when the date is within the the given range' do
+      it 'has no error on the field' do
+        subject.check_date(3.months.ago.strftime("%F"))
+        expect(subject.errors.messages).not_to include(base: ["Please choose a date within the last six months, or <a href='%{url}'>contact us</a> if you would like to see older visits."])
```
TobyRet: Duplication on links 8 and 15 ... Error message could be assigned to shared variable

app/models/print_visits.rb
```diff
@@ -1,5 +1,21 @@
 class PrintVisits
   include MemoryModel
+  # include ActiveModel::Validations
```
TobyRet: Comment

app/controllers/prison/print_visits_controller.rb
StupidCodeFactory: here we should just call `@print_visits.valid?` that @check_date can be called internally by the model since it has access to the date. So no need to pass stuff around

app/models/print_visits.rb
```diff
@@ -1,5 +1,21 @@
...
 
   attribute :visit_date, :accessible_date
+
+  validate :check_date, on: :create
+
+  def check_date(date)
```
StupidCodeFactory: no need to pass the date. We have access to `@accessible_date` then you can `#to_date` it

app/views/prison/print_visits/create.html.erb
```diff
@@ -7,7 +7,9 @@
 <%= render 'visit_date' %>
 
 <% if !@submitted_date.nil? %>
-  <% if @data.any? %>
+  <% if @print_visits.errors.any? %>
```
StupidCodeFactory: `@print_visits.valid?` ?

app/views/prison/print_visits/create.html.erb
```diff
@@ -7,7 +7,9 @@
 <%= render 'visit_date' %>
 
 <% if !@submitted_date.nil? %>
-  <% if @data.any? %>
+  <% if !@print_visits.valid? %>
+    <p class="error-message"><%= t('search_date_invalid_html', scope: [:print_visits, :errors], url: new_prison_feedback_path) %></p>
+  <% elsif @data.any? %>
     <div class="js-print-visits">
```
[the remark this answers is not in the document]
TobyRet: Refactoring to be added to new ticket

PR #1717, opened 2018-06-11
Record User stamp

StupidCodeFactory:
Make sure https://github.com/ministryofjustice/prison-visits-2/pull/1701 is merged first

**CHANGES:**

Sends the current staff user's email to NOMIS API when booking a visit.

### DIFF
--- a/app/services/booking_responder/accept.rb
+++ b/app/services/booking_responder/accept.rb
@@ -29,7 +29,9 @@ class BookingResponder
     end
 
     def nomis_visit_creator
-      @nomis_visit_creator ||= CreateNomisVisit.new(visit)
+      @nomis_visit_creator ||= CreateNomisVisit.new(
+        visit, creator: staff_response.creator
+      )
     end
   end
 end
--- a/app/services/create_nomis_visit.rb
+++ b/app/services/create_nomis_visit.rb
@@ -1,9 +1,10 @@
 class CreateNomisVisit
   ALREADY_BOOKED_IN_NOMIS = 'Duplicate post'.freeze
 
-  def initialize(visit)
+  def initialize(visit, creator:)
     self.visit = visit
     self.api_error = false
+    self.creator = creator
   end
 
   def execute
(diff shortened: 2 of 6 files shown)


app/services/create_nomis_visit.rb
```diff
@@ -71,7 +72,7 @@ def booking_params
       override_visitor_restrictions: false,
       override_vo_balance: false,
       override_slot_capacity: false,
-      client_unique_ref: visit.id
+      client_unique_ref: visit.id, comment: creator.email
```
TobyRet: new line

TobyRet: creator.email needs to be passed to NOMIS via pvb_user_id not comment

StupidCodeFactory: @TobyRet where do you see that pvb_user_id field in API docs?
I think it needs to be passed as through the `X-PVB-User-id` header.
https://github.com/ministryofjustice/nomis-api/blob/dev/src/main/java/syscon/nomisapi/visit/OffenderVisits.java#L628

app/services/booking_responder/accept.rb
```diff
@@ -29,7 +29,9 @@ def book_to_nomis
     end
 
     def nomis_visit_creator
-      @nomis_visit_creator ||= CreateNomisVisit.new(visit)
+      @nomis_visit_creator ||= CreateNomisVisit.new(
+        visit, creator: staff_response.creator
```
TobyRet: Is it also worth passing the visit as a hash for consistency / increased safety?

StupidCodeFactory:
@TobyRet @KPobeeNorris I've reopened the PR against master here https://github.com/ministryofjustice/prison-visits-2/pull/1723

so I'm closing this

app/services/booking_responder/accept.rb
StupidCodeFactory: No, using keywords argument everywhere is just insane clutter. Here `visit` is the main subject of the class which also bares 'Visit` in the class name, I'd rather remove the `creator` keyword argument

PR #1719, opened 2018-06-12
[WIP] Book to NOMIS comments

chrishanes:


db/migrate/20180614095419_add_nomis_comments_to_visits.rb
```diff
@@ -0,0 +1,5 @@
+class AddNomisCommentsToVisits < ActiveRecord::Migration[5.2]
+  def change
+    add_column :visits, :nomis_comments, :string
```
StupidCodeFactory: shouldn't this be a `text` instead of `string` i think strings on hold up to 256 chars which may not be enough

TobyRet: Yep good spot @StupidCodeFactory - I'll change

PR #1723, opened 2018-06-14
Send user stamp into nomis

StupidCodeFactory:
**CHANGES:**

Send userstamp through headers on booking a visit into NOMIS


app/services/create_nomis_visit.rb
```diff
@@ -1,9 +1,11 @@
 class CreateNomisVisit
   ALREADY_BOOKED_IN_NOMIS = 'Duplicate post'.freeze
+  PVB_USER_ID_HEADER_FIELD = 'X-PVB-User-id'.freeze
```
TobyRet: @StupidCodeFactory - any particular reason for using the custom header ('X-') convention? I don't think it's required

StupidCodeFactory: good stop will change

PR #1811, opened 2018-08-23
Pull Request template

KPobeeNorris:
As a team we have been discussing how we can be consistent around our
git styles.  One of the tasks is to create a PR template which will help
the team with this consistency, and act as a prompt for ensuring things
like the documentation has been updated, or the deploy repo has been
updated.

This is a first pass and will need to be reviewed by the wider team as
there are not only devs that create PRs, and we want this to be useful/
meaningful for everyone.

Trello card: https://trello.com/c/uBKazmeE/178-use-gds-git-style-guide

### DIFF
--- a/docs/pull_request_template.md
+++ b/docs/pull_request_template.md
@@ -1,0 +1,24 @@
+<!--- Provide a general summary of your changes in the Title above -->
+
+## Description
+<!--- Describe your changes in detail -->
+
+## Motivation and Context
+<!--- Why is this change required? What problem does it solve? -->
+<!--- Is there a Trello card to link to? -->
+
+## Types of changes
+<!--- What types of changes does your code introduce? Put an `x` in all the boxes that apply: -->
+- [ ] Bug fix (non-breaking change which fixes an issue)
+- [ ] New feature (non-breaking change which adds functionality)
+- [ ] Breaking change (fix or feature that would cause existing functionality to change)
+- [ ] Prison details update (e.g. slot updates)
+
+## Checklist:
+<!--- Go over all the following points, and put an `x` in all the boxes that apply. -->
+<!--- If you're unsure about any of these, don't hesitate to ask! -->
+- [ ] [My commit message follows the GDS git style standards we have adopted](https://github.com/alphagov/styleguides/blob/master/git.md).
+- [ ] My change requires a change to the documentation.
+- [ ] I have updated the documentation accordingly.
+- [ ] I have updated the deployment repo accordingly.
+- [ ] I have added tests to cover my changes.


docs/pull_request_template.md
```diff
@@ -0,0 +1,26 @@
...
+<!--- What types of changes does your code introduce? Put an `x` in all the boxes that apply: -->
+- [ ] Bug fix (non-breaking change which fixes an issue)
+- [ ] New feature (non-breaking change which adds functionality)
+- [ ] Breaking change (fix or feature that would cause existing functionality to change)
+- [ ] Prison details update (e.g. slot updates)
+
+## Checklist:
+<!--- Go over all the following points, and put an `x` in all the boxes that apply. -->
+<!--- If you're unsure about any of these, don't hesitate to ask! -->
+- [ ] My code follows the code style of this project.
```
TobyRet: I wonder if the code style bit is necessary as Code Climate will not pass the build unless styles are adhered to?

docs/pull_request_template.md
```diff
@@ -0,0 +1,26 @@
...
+- [ ] [My commit message follows the GDS git style standards we have adopted](https://github.com/alphagov/styleguides/blob/master/git.md).
+- [ ] My change requires a change to the documentation.
+- [ ] I have updated the documentation accordingly.
+- [ ] I have updated the deployment repo accordingly.
+- [ ] I have added tests to cover my changes.
+- [ ] All new and existing tests passed.
```
TobyRet: This will be checked automatically also?

docs/pull_request_template.md
KPobeeNorris: @TobyRet - Agreed!  Thought we would probably get rid of this, think I just left it in as a way of showing what kinds of checks we could use.

docs/pull_request_template.md
KPobeeNorris: As above re: @TobyRet's comment

docs/pull_request_template.md
```diff
@@ -0,0 +1,24 @@
...
+<!--- What types of changes does your code introduce? Put an `x` in all the boxes that apply: -->
+- [ ] Bug fix (non-breaking change which fixes an issue)
+- [ ] New feature (non-breaking change which adds functionality)
+- [ ] Breaking change (fix or feature that would cause existing functionality to change)
+- [ ] Prison details update (e.g. slot updates)
+
+## Checklist:
+<!--- Go over all the following points, and put an `x` in all the boxes that apply. -->
+<!--- If you're unsure about any of these, don't hesitate to ask! -->
+- [ ] [My commit message follows the GDS git style standards we have adopted](https://github.com/alphagov/styleguides/blob/master/git.md).
```
alan: Should we add the exception about not rebasing PRs? Feels like we should link somewhere else, but that does not exist yet...

PR #2100, opened 2019-02-28
Kubernetes deployment scripts

TobyRet:
We need to migrate Prison Visits from Template Deploy to Kubernetes. To begin with it has been agreed that we will just move staging, and this PR sets up the initial deployment scripts.

Note that we will be deploying to both old and new deployment environments for now.


config/database.yml
```diff
@@ -81,5 +81,5 @@ test:
 production:
   <<: *default
   database: pvb2_production
-  username: pvb2
-  password: <%= ENV['PVB2_DATABASE_PASSWORD'] %>
+#  username: pvb2
+  url: <%= ENV["DATABASE_URL"] %>
```
TobyRet: Need to revert back to the earlier config else the template deployment will break

config/initializers/sidekiq.rb
```diff
@@ -0,0 +1,10 @@
+if Rails.env.production? && Rails.configuration.kubernetes_deployment?
```
TobyRet: We have a new redis instance created in the new Cloud Platform environment. We only want to point to it if we are deploying to Kubernetes else template deploy will break

config/initializers/sidekiq.rb
```diff
@@ -0,0 +1,10 @@
...
+  Sidekiq.configure_server do |config|
+    config.redis = { url: "redis://#{Rails.configuration.redis_url}:6379", network_timeout: 5 }
```
TobyRet: We are currently getting `connection lost` errors with Redis. Possibly a Cloud Platform issue

deploy/staging/deployment.yaml
```diff
@@ -0,0 +1,589 @@
...
+              valueFrom:
+                secretKeyRef:
+                  name: prison-visits-booking-staff-secrets
+                  key: rails_web_concurrency
+            - name: REDIS_URL
+              valueFrom:
+                secretKeyRef:
+                  name: prison-visits-booking-staff-secrets
+                  key: redis_url
+        - name: prison-visits-booking-staff-sidekiq
```
TobyRet: Need to refactor deployment script to use config maps in order to remove duplicated env vars

deploy/staging/deployment.yaml
```diff
@@ -0,0 +1,589 @@
...
+                  key: smtp_username
+            - name: SMTP_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: prison-visits-booking-staff-secrets
+                  key: smtp_password
+            - name: REMOVE_LOAD_TEST_DATA
+              value: "true"
+            - name: PUBLIC_SERVICE_URL
+              value: "https://public-staging-pvb.dsd.io"
```
KPobeeNorris: Do we want to add the new k8s public URL here?

config/initializers/sidekiq.rb
rossjones: Is it worth also specifying the reconnection config as in https://github.com/redis/redis-rb#reconnections

deploy/staging/deployment.yaml
```diff
@@ -0,0 +1,601 @@
...
+        app: prison-visits-booking-staff
+    spec:
+      containers:
+        - name: prison-visits-booking-staff
+          image: 926803513772.dkr.ecr.eu-west-1.amazonaws.com/prison-visits-booking/prison-visits-staff:latest
+          imagePullPolicy: Always
+          command: ['sh', '-c', "bundle exec rails db:migrate && bundle exec rails db:seed && bundle exec puma -p 3000 -C ./config/puma_prod.rb --pidfile /tmp/server.pid"]
+          ports:
+            - containerPort: 3000
+#          livenessProbe:
```
rossjones: Reminder this probably needs to go back in at some point..

deploy/staging/deployment.yaml
```diff
@@ -0,0 +1,601 @@
...
+                secretKeyRef:
+                  name: prison-visits-booking-staff-secrets
+                  key: sentry_js_dsn
+            - name: ZENDESK_TOKEN
+              valueFrom:
+                secretKeyRef:
+                  name: prison-visits-booking-staff-secrets
+                  key: zendesk_token
+            - name: ZENDESK_URL
+              value: ""
```
rossjones: Is this intentionally blank?

deploy/staging/deployment.yaml
TobyRet: Done

deploy/staging/deployment.yaml
TobyRet: Filled in now

deploy/staging/deployment.yaml
TobyRet: Yep - will need to fix those 503s!

.circleci/config.yml
```diff
@@ -1,3 +1,92 @@
...
+
+  test_container_config: &test_container_config
+    docker:
+      - image: circleci/ruby:2.5.3-node-browsers
+        environment:
+          RAILS_ENV: test
+          PGHOST: 127.0.0.1
+          PGUSER: ubuntu
+          RACK_ENV: test
+      - image: circleci/postgres:10.5-alpine
```
KPobeeNorris: @TobyRet - does this need to be 9.4?

TobyRet: ah yes

PR #2159, opened 2019-03-26
Fix metrics cron job

KPobeeNorris:
The metrics cron job hasn't been running in K8s on staging, and it
appears that this could be due to a yaml formatting issue.

## Types of changes

## Checklist:

### DIFF
--- a/deploy/staging/cronjob.yaml
+++ b/deploy/staging/cronjob.yaml
@@ -3,7 +3,7 @@ kind: CronJob
 metadata:
   name: refresh-views
 spec:
-  schedule: "0 0 * * *"
+  schedule: "5 1 * * *"
   startingDeadlineSeconds: 10
   concurrencyPolicy: Forbid
   successfulJobsHistoryLimit: 3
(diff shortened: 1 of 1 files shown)


deploy/staging/cronjob.yaml
```diff
@@ -13,10 +13,10 @@ spec:
       template:
         spec:
           containers:
-          - name: prison-visits-booking-staff-pvb-metrics-refresh
-            image: 926803513772.dkr.ecr.eu-west-1.amazonaws.com/prison-visits-booking/prison-visits-staff:latest
-            imagePullPolicy: Always
-            command: ['sh', '-c', "bundle exec rake pvb:metrics:refresh"]
+            - name: prison-visits-booking-staff-pvb-metrics refresh
```
rossjones: Is the missing hyphen in the name deliberate?

deploy/staging/cronjob.yaml
```diff
@@ -3,7 +3,7 @@ kind: CronJob
 metadata:
   name: refresh-views
 spec:
-  schedule: "0 0 * * *"
+  schedule: "5 1 * * *"
```
TobyRet: I'm not sure I understand how changing the time the job fires fixes the issue? 😢

deploy/staging/cronjob.yaml
TobyRet: 👍

deploy/staging/cronjob.yaml
KPobeeNorris: Sorry - this was part of my initial thinking, but will change back now I've realised it's indentation yaml pain

PR #2237, opened 2019-05-13
Migrate PVB Staff from Live-0 -> Live-1

KPobeeNorris:
When we initially moved PVB Staff staging onto Kubernetes we did so
using the original Cloud Platform Live-0 cluster.  However, they are now
using the Live-1 cluster and in preparation for our full PVB migration
from Template Deploy to Kubernetes we need to migrate our staging env to this new
cluster.

## Types of changes

## Checklist:


Dockerfile.kubernetes
```diff
@@ -62,6 +62,13 @@ RUN bundle install --without development test --jobs 2 --retry 3
 
 COPY . /app
 
+RUN mkdir -p /home/appuser && \
+  useradd appuser -u 1001 --user-group --home /home/appuser && \
+  chown -R appuser:appuser /app && \
+  chown -R appuser:appuser /home/appuser
+
+USER 1001
```
TobyRet: Can you remind me about why we are using `1001` as the user?

KPobeeNorris: @TobyRet - It's about the live-1 cluster security and not running as root

rossjones: Without setting the user uid, it fails to work out that it isn't a root user (uid of 0).  There's normally already a uid of 1000 (which isn't appuser) so it's just one more than the current max uid.

PR #2332, opened 2019-07-23
high-down-slot-update

Ed-Bajo:
<High Down slot update>

## Description
<High Down have requested for slot time changes which will be effective from 1st August 2019.>

## Types of changes

## Checklist:


rossjones:
@Ed-Bajo As it is to take effect in the future, we need to add anomalous dates between now and then, so that the slot updates don't take immediate effect.  I've added all the anomalous dates for this one.

PR #2479, opened 2019-10-24
Redirect the old url to new url.

rossjones:
After the catastrophe that was the migration, we ended up with an old
broken URL and two ingresses in our deployment for production, ingress
(old url) and ingress-for-justice (new justice domain url).

This PR redirects all traffic to the old URL to the new URL with a
temporary redirect. We want a temporary until we are happy that this is
working as expected and then we should change to a permanent.


rossjones:
There is no PVB preproduction.

PR #2489, opened 2019-10-28
Hull-slot-update - pending nomis update

rossjones:
Reverts ministryofjustice/prison-visits-2#2488

Do not merge this until we have had confirmation from Hull that they have updated their NOMIS slots.


KPobeeNorris:
This PR has been open for many months and needs to be closed.  We can create a new PR if the prison contact us again about their slots.

PR #2493, opened 2019-10-29
Pom 432 unbookable slots

starswan:
## Description
Add a page to add a new 'un-bookable slot' for a prison

## Motivation and Context
Make the slot updates self-service

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist:


app/services/prison_seeder/seed_entry.rb
```diff
@@ -65,10 +66,13 @@ def postcode
   end
 
   def slot_details
+    today = Time.zone.today.to_s
     {
       'recurring' => hash.fetch('recurring', {}),
       'anomalous' => hash.fetch('anomalous', {}),
-      'unbookable' => hash.fetch('unbookable', [])
+      'unbookable' => (hash_unbookable_slots + unbookable_slots).
```
rossjones: I've a feeling this would be easier if we imported the current unbookable slots for each prison (e.g. do a db:seed) and then disabled the setting of unbookable slots in the yaml?

starswan: The deprecation of the various bits of YAML is certainly a conversation point - you can't delete them from the UI (as they would come bouncing back on next deploy from the YAML)

PR #2500, opened 2019-10-29
Migrate from old NOMIS API to Elite2 API

rossjones:
Currently PVB uses the old NOMIS API, but we must migrate to Elite2.
This commit includes the basics for obtaining a client_credentials token
from HMPPS SSO for use when calling the Elite2 API.
## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] My change requires a change to the documentation.


app/services/nomis/client.rb
```diff
@@ -50,8 +48,8 @@ def get(route, params = {})
     # rubocop:disable Metrics/MethodLength
     def request(method, route, params, idempotent:, options: {})
       # For cleanliness, strip initial / if supplied
-      route = route.sub(%r{^\/}, '')
-      path = "/nomisapi/#{route}"
+      path = "/elite2api/api/v1/#{route}"
```
starswan: the comment above was for the line of code that you have removed

spec/services/nomis/api_spec.rb
```diff
@@ -261,7 +261,7 @@
     subject { super().fetch_contact_list(params) }
 
     it 'returns an array of contacts' do
-      expect(subject).to have_exactly(27).items
+      expect(subject).to have_exactly(26).items
```
starswan: slightly worried about this one? Suggests a change in behaviour. Maybe improve test on master to expect actual data so that we can see which contact has been mysteriously removed?

KPobeeNorris: This was due to a change in the behaviour of the API and has since been fixed.

PR #2764, opened 2020-08-21
refactor of slots code extracted from POM-463

starswan:
#1663  Description
This is a refactoring pulled from the spike on POM-463.

## Motivation and Context
 It isn't directly related to the slots work, but might be useful as it does de-tangle the code around this area.

## Checklist:


spec/support/helpers/service_helpers.rb
```diff
@@ -22,4 +22,27 @@ def mock_nomis_with(api, double_or_spy)
...
+          ],
+          'links' => {
+            'profile' => 'http://example.com/profile',
+            'logout' => 'http://example.com/logout'
+          }
+        }
+      }
+
+    OmniAuth.config.add_mock(:mojsso, sso_response)
+  end
```
ollietreend, line 47: Looks like this probably isn't needed, as it's unused and will need to be removed once HMPPS Auth is implemented by #2772 POM-807

ollietreend (commented):
These changes generally look sensible – moving things into more 'obvious' places. Although some of the detail is lost on me due to not having delved very far into PVB slot code.

spec/support/helpers/service_helpers.rb
starswan: I've removed it, but it's a bit shocking that it isn't used. I think it was created for the new UI spikes (which have now been binned)

ollietreend (approved):

PR #2772, opened 2020-08-28
Pom-807 spike hmpps signon

starswan:
## Description
Spike to convert from Moj Signon to HMPPS Signon.
Does not include a role as we do not yet have a role defined for these users.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:


spec/models/signon_identity_spec.rb
```diff
@@ -159,11 +158,15 @@
       end
 
       context 'when an admin' do
-        let!(:orgs) { [EstateSSOMapper::DIGITAL_ORG] }
+        let(:roles) { ["ROLE_MOIC_ADMIN"] }
```
ollietreend, line 161: I wonder if we should have a new role made specifically for PVB admin, rather than using `ROLE_MOIC_ADMIN`.

That would make the service more portable if it ever moves to a different team.

spec/rails_helper.rb
```diff
@@ -11,6 +11,7 @@
 require 'support/helpers/controller_helper'
 
 WebMock.disable_net_connect!(allow: 'codeclimate.com', allow_localhost: true)
+# WebMock.disable_net_connect!(allow: ->(uri) { ['codeclimate.com', 'www.google-analytics.com'].include?(uri.host) }, allow_localhost: true)
```
ollietreend, line 14: Is this line needed, or can it be removed?

spec/support/helpers/auth_helper.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module AuthHelper
+  ACCESS_TOKEN = Struct.new(:access_token).new('an-access-token')
+
+  def stub_auth_token
+    allow(Nomis::Oauth::TokenService).to receive(:valid_token).and_return(ACCESS_TOKEN)
+
+    stub_request(:post, "https://gateway.t3.nomis-api.hmpps.dsd.io/auth/oauth/token?grant_type=client_credentials").
```
ollietreend, line 9: Is this URL correct? Should it be stubbing the new URL, rather than the old T3 URL?

```suggestion
    stub_request(:post, "https://sign-in-dev.hmpps.service.justice.gov.uk/auth/oauth/token?grant_type=client_credentials").
```

ollietreend (approved):
@starswan one comment (a question), otherwise this looks good

PR #2935, opened 2022-02-01
Separate commands in Dockerfile

georgeschena:
Separate commands in Dockerfile in order to resolve no valid OpenPGP data found error

[Link to stackoverflow question](https://stackoverflow.com/questions/21338721/gpg-no-valid-openpgp-data-found)


psoleckimoj (approved):
LGTM

PR #3051, opened 2022-08-05
changed to api version 1, nginx updated

hutcheonb-moj:
## Description
Changed k8 api version from v1beta to v1
Changed to new nginx ingress controller

## Motivation and Context
All projects need to incorporate these changes by the end of August

## Checklist:
To be merged Monday before 7am to have the least possible effect on the Users.


psoleckimoj (approved):
LGTM!

PR #3059, opened 2022-08-12
VB-1100, remove Huntercombe number from emails

hutcheonb-moj:
## Description
HMP Huntercombe currently has no working phone line for visits booking, so this change removes their phone number from the Email's we send out.
This may be a temporary change, the Comms team are contacting them to see if this will be permanent

## Types of changes
- [x] Removal of Prison's phone number


psoleckimoj (approved):
LGTM!

PR #3061, opened 2022-08-19
VB-1103, disable Pentonville's public access

hutcheonb-moj:
## Description
HMP Pentonville have requested that we disable their prison as the current system no longer supports their timetable.
This is being followed up separately, but they wish to make this change, at this time

## Types of changes
- [x] Disable public access


psoleckimoj (approved):
LGTM!

PR #3063, opened 2022-08-24
VB-1102, Norwich timetable updated

hutcheonb-moj:
## Description
HMP Norwich requested a change to their weekend timetable.
Currently our groupings don't fit perfectly with the way they do their groupings, so this new timetable is made with the knowledge that it doesn't perfectly fit their schedule.

## Types of changes
- [x] Change of timetable


psoleckimoj (approved):
LGTM!

PR #3091, opened 2022-10-28
VB-1367, block 4 dates as fully booked - huntercombe

hutcheonb-moj:
## Description
5th, 6th, 12th & 13th of November are all fully booked for HMP Huntercombe, so added these dates to the un-bookable list so it no longer is an option for the public

## Types of changes
- [x] Visit slot updates


psoleckimoj (approved):
LGTM!

PR #3092, opened 2022-11-01
VB-1344, added service down message so no page will render

hutcheonb-moj:
## Description
Removed the 'yield' tag and some surrounding code which renders the content on the page, this will ensure no matter which page the users have bookmarks for, they will only be displayed the service down page

## Types of changes
- [x] Temporary service down page


psoleckimoj (approved):
LGTM!

PR #3097, opened 2022-11-01
VB-1344, final edit to service down page, remove down banner

hutcheonb-moj:
## Description
Remove the banner for the upcoming downtime, so less to replace when we're finished with downtime.
Final change to service down display


psoleckimoj (approved):
LGTM!

PR #3125, opened 2022-12-01
VB-1483, re-enable Winchester

hutcheonb-moj:
## Description
Re-enable HMP Winchester

## Types of changes
- [x] Removed from shared environment and re-enable
- [x] Slot updates


tpmcgowan (approved):
LGTM

PR #3146, opened 2022-12-29
VB-1562, email and phone number update, all enabled prisons

hutcheonb-moj:
## Description
Changed emails from zendesk and hmps.gsi to their replacements
Reformatted phone numbers to match a 4-3-4 format


psoleckimoj (approved):
LGTM!

PR #3148, opened 2023-01-03
Updated all links to include HTTPS

hutcheonb-moj:
## Description
Updated all links in PVB to be https://www instead of just HTTP or www
These are sent out on emails, so safer to include this


psoleckimoj (approved):
LGTM!

PR #3161, opened 2023-01-20
VB-1718, Block specific dates - Wymott

hutcheonb-moj:
## Description
Block specific dates in HMP Wymott visit schedule
## Types of changes
- [x] Slot updates - block specific dates


tpmcgowan (approved):
LGTM

PR #3204, opened 2023-04-25
Fix based on Matt Whittaker's suggestions

hutcheonb-moj:
## Description
Changed to timeout seconds based on feedback in #ask-prison-visit-booking


tpmcgowan (approved):
LGTM

PR #3222, opened 2023-05-24
VB-2264, block second date

hutcheonb-moj:
## Description
Block specific date - HMP The Mount

## Types of changes
- [x] Temporary timetable change


tpmcgowan (approved):
LGTM

PR #3227, opened 2023-06-06
VB-2252, onboard NSC using capacity feature

hutcheonb-moj:
## Description
Onboarding for HMP North Sea Camp
They have updated their visit slots in NOMIS, so can use the capacity feature in PVB

## Types of changes
- [x] Prison onboarding


psoleckimoj (approved):
LGTM!

PR #3229, opened 2023-06-08
VB-2252, removed NSC from shared environment, use manual slots

hutcheonb-moj:
## Description
Remove North Sea Camp from shared environments get NOMIS slots, slots haven't been pulled in when the cronjob last run so this is currently blocking the onboarding, needs further investigation


AledWEvans (approved):
LGTM

PR #3234, opened 2023-06-22
VB-2378/VB-2379 block specific dates - Warren Hill / Foston Hall

hutcheonb-moj:
## Description
Block specific dates for HMP Foston Hall
Block specific dates for HMP Warren Hill
## Types of changes
- [x] Temporary timetable change


tpmcgowan (approved):
LGTM

PR #3274, opened 2023-09-26
Update BRI-bure.yml

simonnebMOJ:
Added new dates and time plus 1 anomalous date for 30/09 so we can release now.


hutcheonb-moj (approved):
LGTM

PR #3294, opened 2023-10-16
Revert "Revert "Feature/upgrade to rails 5.2.8.1""

Reyko:
Reverts ministryofjustice/prison-visits-2#3292

Set `config.action_mailer.raise_delivery_errors = false`

We don't want to raise errors for smtp on production as we don't use it


hutcheonb-moj (approved):
LGTM

PR #3336, opened 2023-11-03
Update MTI-the-mount.yml

simonnebMOJ:
The social visits for Monday - Thursday are changing the start time will be 14.00 and the end time will 15.45. 

This will start on Monday 4th December 2023.


hutcheonb-moj (approved):
LGTM

PR #3339, opened 2023-11-06
Update HOI-high-down.yml

simonnebMOJ:
This is to inform you that there will be no visits on Tuesday 26th December at High DOwn.


hutcheonb-moj (approved):
LGTM

PR #3399, opened 2023-11-20
Update LII-lincoln.yml

simonnebMOJ:
Updated timetable to match GOV.UK and VSiP - also removed the email address as not used for social visits.


hutcheonb-moj (approved):
LGTM

PR #3432, opened 2023-11-28
Update EHI-standford-hill.yml

simonnebMOJ:
Removed email address as now Family Services led. They dont want emails coming in.


hutcheonb-moj (approved):
LGTM

PR #3467, opened 2023-11-30
Update VEI-the-verne.yml

simonnebMOJ:
Blocked Boxing day and moved to day after at eh request of The Verne


hutcheonb-moj (approved):
LGTM

PR #3491, opened 2023-12-07
[VB-3007]-move-deployment-over-helm-setup

Reyko:
## Types of changes

## Checklist:


.circleci/config.yml
```diff
@@ -42,6 +42,7 @@ references:
 version: 2.1
 
 orbs:
+  hmpps: ministryofjustice/hmpps@8.0.3
```
psoleckimoj, line 45: This can be changed just to `ministryofjustice/hmpps@8` so we don't need to update to minor versions 🙂

.circleci/config.yml
```diff
@@ -116,112 +117,46 @@ jobs:
           region: $ECR_REGION
       - run: |
           aws ecr get-login-password --region $ECR_REGION | docker login --username AWS --password-stdin ${AWS_ECR_REGISTRY_ID}.dkr.ecr.${ECR_REGION}.amazonaws.com
```
psoleckimoj, line 119: My only comment on this is that we don't use ECR in general, we use quay.io. However, if this works using ECR for now, then we can get this done, then move over to quay.io?

helm_deploy/prison-visits-2/Chart.lock
```diff
@@ -0,0 +1,9 @@
+dependencies:
```
psoleckimoj, line 1: This needs to be in the `.helmignore` and removing from the commit.

mattops (dismissed):
LGTM

hutcheonb-moj (approved):
LGTM

PR #3535, opened 2024-01-21
Update/multiple changes

simonnebMOJ:
multiple changes to blocked dates at establishments - also change to Liverpool name to remove reductant 'Social Visits'.


hutcheonb-moj (approved):
LGTM

PR #3552, opened 2024-02-27
Change dev cronjob runtime to reduce sentry errors

hutcheonb-moj:
## Description

Change development cronjob runtime to 7 / 8 am. As the overnight shutdown is causing excessive errors in Sentry


tpmcgowan (approved):
LGTM

PR #3557, opened 2024-02-28
Load nomis slots - cronjob time change (dev)

hutcheonb-moj:
## Description
Change cronjob 'load_nomis_slots' time on dev to also run at 7am


psoleckimoj (approved):
LGTM!

PR #3609, opened 2024-04-30
Remove unused deployment files

hutcheonb-moj:
Remove deployment files from previous deployment method, now unused


simonnebMOJ (approved):
LGTM - however may be a bit above my dev skills!

PR #3628, opened 2024-06-24
Update foston hall report

hutcheonb-moj:
Update Foston Hall report - only run from 28th May until now (only returns visitors who have seen notice)


psoleckimoj (approved):
LGTM!

PR #3637, opened 2024-07-19
Remove requested visits script

hutcheonb-moj:
This script should be used when re-onboarding a prison back onto PVB

The prison name needs changing in the script

Only targets requested visits, changes them to withdrawn


simonnebMOJ (approved):
LGTM

PR #3660, opened 2024-09-10
Temporarily turn off Brixton from 19/09

simonnebMOJ:
Brixton would like to shut down the online booking service for HMP Brixton from 19.09.2024 until 10.10.2024. 

This is due to staffing issues and staff who will be trained to just book visits through the telephone for this period of time.


dhirajshettymoj (approved):
LGTM

PR #3681, opened 2024-10-15
Update HVI-haverigg.yml

simonnebMOJ:
Fixed timetable to DPS - will use to initiate link between VSI and PVB so DO NOT MERGE until VSIP live.

Tickets:

VB-4717


hutcheonb-moj (approved):
lgtm

PR #3703, opened 2024-11-06
Add rake task for staff messages to visitors

hutcheonb-moj:
Add new rake task:

manual_tasks:staff_messages

Prints a CSV suitable list of all messages sent by visits staff to visitors, over X length and in between Y and Z dates


simonnebMOJ (dismissed):
LGTM

simonnebMOJ (approved):
LGTM

PR #3712, opened 2024-11-22
Add prisoner name/number report

hutcheonb-moj:


lib/tasks/manual_tasks.rake
```diff
@@ -31,4 +31,15 @@ namespace :manual_tasks do
       puts ''
     end
   end
+  desc 'Prisoner name / number, as entered by the visitor'
+    task prisoner_name_and_number: :environment do
+      start_date = '2024-09-01'
+      end_date = '2024-11-01'
+      prisoners = Prisoner.where(created_at: start_date..end_date).order(created_at: :desc)
```
simonnebMOJ, line 38: you could add a `uniq `filter here to only get the unique returns:

```
 prisoners = Prisoner
                .where(created_at: start_date..end_date)
                .order(created_at: :desc)
                .uniq { |prisoner| [prisoner.first_name, prisoner.last_name, prisoner.number] }
```

PR #3724, opened 2024-12-20
Turn Brixton back on

simonnebMOJ:
Brixton have requested we turn off the service over Christmas from 23/12 to 03/01 inclusive.

This PR is for turning Brixton BACK ON.

DO NOT MERGE until afternoon of 03/01.


hutcheonb-moj (approved):
LGTM

PR #3725, opened 2024-12-23
Remove family services email

simonnebMOJ:
At request of Family Services, removing the email address from the prisons they support - most had been done but a few remained.


hutcheonb-moj (approved):
LGTM

PR #3726, opened 2025-01-02
Update LPI-liverpool-social-visits.yml

simonnebMOJ:
Updated and added dates that go live on week commencing 13 January 2025


hutcheonb-moj (approved):
LGTM

PR #4049, opened 2026-05-19
Trim whitespace visitor names

hutcheonb-moj:
Manually tested the tests (removing implementation) - to ensure the test wasn't passing due to a equals vs strict equals scenario


simonnebMOJ (approved):
LGTM

PR #4052, opened 2026-05-19
Trim visitor names received from API

hutcheonb-moj:
Manually tested the tests (removing implementation) - to ensure the test wasn't passing due to a equals vs strict equals scenario

Some 'magic' tests removed


simonnebMOJ (approved):
LGTM

PR #4068, opened 2026-06-15
Remove High Down from NOMIS availability lists

simonnebMOJ:
High Down will use VSIP intelligence, so removing from NOMIS list.


hutcheonb-moj (dismissed):

hutcheonb-moj (approved):
lgtm

hutcheonb-moj (approved):

